PR #4, opened 2023-08-24
Remove unused imports format fix

andrecsilva:
## Overview
Fixes for the `RemoveUnusedImports` codemod

## Description

Will now keep the format as much as possible to avoid unnecessary changes.

Some extra tests to cover those cases


clavedeluna (commented):
How did this decision come about? I thought the format update was a nice touch! Could even be added as parametrization.

andrecsilva:
> How did this decision come about? I thought the format update was a nice touch! Could even be added as parametrization.

Pygoat has some files that clearly has never seen any formatter. Some from imports there have commas without spaces, which was prompting this codemod to change it to a standard formatting even if the imports themselves were not changed.

clavedeluna:
ah ok so this is a result of seeing pygoat results, got it!

clavedeluna (approved):

PR #5, opened 2023-08-25
Implement "sugar" API for SecureRandom codemod example

drdavella:
## Overview
*Implement API to provide a cleaner example for our `SecureRandom` codemod*

## Description

* We need to implement a simpler API for creating custom (and internal) codemods
* This is a highly visible part of our product. It's in our docs and it comes up in demos and talks
* I tried to emulate the simplicity of the Java codemodder example for the same rule
* I worked backwards by writing the codemod the way that I would want to implement it
* Next I wrote some code to actually implement the API
  * I was able to do this without making any major changes to the existing codemodder API
  * I'm not entirely proud of the solution. It's not really general at all and there is a lot of metaprogramming glue to hack it all together
  * However, it serves its purpose and gives us a cleaner working example for this rule
* We are going to need to iterate on the API itself and also on the underlying implementation
* I also think there are many opportunities for improvement in the existing codemod API itself. Making some of these changes might help facilitate a cleaner solution all around
  * I think I'm not a big fan of the use of class-level attributes, but I need to think about a better solution
  * I'm also a little wary of the helper/mixin pattern, although that's hard to avoid too
  * Some of this was clearly driven by the design of libCST itself, which dictates the form of our solutions to some extent
  * However we should think outside the box and consider whether there might be other ways to phrase these codemods, even outside the context of classes
* I'm looking for feedback both on the final result itself but also on other ways forward

### DIFF
--- a/codemodder/codemods/api/__init__.py
+++ b/codemodder/codemods/api/__init__.py
@@ -12,7 +12,6 @@ import yaml
 
 from codemodder.codemods.base_codemod import (
     CodemodMetadata,
-    ReviewGuidance,
     SemgrepCodemod as _SemgrepCodemod,
 )
 from codemodder.codemods.base_visitor import BaseTransformer
@@ -52,12 +51,14 @@ class _CodemodSubclassWithMetadata:
             # NAME, DESCRIPTION, and REVIEW_GUIDANCE here.
 
             cls.METADATA = CodemodMetadata(
-                cls.DESCRIPTION, cls.NAME, cls.REVIEW_GUIDANCE
+                cls.DESCRIPTION,  # pylint: disable=no-member
+                cls.NAME,  # pylint: disable=no-member
+                cls.REVIEW_GUIDANCE,  # pylint: disable=no-member
             )
             cls.YAML_FILES = _create_temp_yaml_file(cls, cls.METADATA)
 
             # This is a little bit hacky, but it also feels like the right solution?
-            cls.CHANGE_DESCRIPTION = cls.DESCRIPTION
+            cls.CHANGE_DESCRIPTION = cls.DESCRIPTION  # pylint: disable=no-member
 
             super().__init_subclass__()
 
@@ -102,7 +103,7 @@ class SemgrepCodemod(
                 Change(str(line_number), self.CHANGE_DESCRIPTION).to_json()
             )
             if (attr := getattr(self, "on_result_found", None)) is not None:
-                new_node = attr(updated_node)
+                new_node = attr(updated_node)  # pylint: disable=not-callable
                 # TODO: where does this actually belong?
                 RemoveImportsVisitor.remove_unused_import_by_node(
                     self.context, original_node
--- a/codemodder/codemods/api/__init__.py
+++ b/codemodder/codemods/api/__init__.py
@@ -1,0 +1,112 @@
+import io
+import os
+import tempfile
+from typing import List
+
+import libcst as cst
+from libcst.codemod import (
+    CodemodContext,
+)
+from libcst.codemod.visitors import RemoveImportsVisitor
+import yaml
+
+from codemodder.codemods.base_codemod import (
+    CodemodMetadata,
+    ReviewGuidance,
+    SemgrepCodemod as _SemgrepCodemod,
+)
+from codemodder.codemods.base_visitor import BaseTransformer
+from codemodder.codemods.change import Change
+from codemodder.file_context import FileContext
+from .helpers import Helpers
+
+
+def _populate_yaml(rule: str, metadata: CodemodMetadata) -> str:
+    config = yaml.safe_load(io.StringIO(rule))
+    # TODO: handle more than rule per config?
+    assert len(config["rules"]) == 1
+    config["rules"][0].setdefault("id", metadata.NAME)
+    config["rules"][0].setdefault("message", "Semgrep found a match")
+    config["rules"][0].setdefault("severity", "WARNING")
+    config["rules"][0].setdefault("languages", ["python"])
+    return yaml.safe_dump(config)
+
+
+def _create_temp_yaml_file(orig_cls, metadata: CodemodMetadata):
+    fd, path = tempfile.mkstemp()
+    with os.fdopen(fd, "w") as ff:
+        ff.write(_populate_yaml(orig_cls.rule(), metadata))
+
+    return [path]
+
+
+class _CodemodSubclassWithMetadata:
+    def __init_subclass__(cls):
+        # This is a pretty yucky workaround.
+        # But it is necessary to get around the fact that these fields are
+        # checked by __init_subclass__ of the other parents of SemgrepCodemod
+        # first.
+        if not cls.__name__ == "SemgrepCodemod":
+            # TODO: if we intend to continue to check class-level attributes
+            # using this mechanism, we should add checks (or defaults) for
+            # NAME, DESCRIPTION, and REVIEW_GUIDANCE here.
+
+            cls.METADATA = CodemodMetadata(
+                cls.DESCRIPTION, cls.NAME, cls.REVIEW_GUIDANCE
+            )
+            cls.YAML_FILES = _create_temp_yaml_file(cls, cls.METADATA)
+
+            # This is a little bit hacky, but it also feels like the right solution?
+            cls.CHANGE_DESCRIPTION = cls.DESCRIPTION
+
+            super().__init_subclass__()
+
+        return cls
+
+
+# NOTE: this shadows base_codemod.SemgrepCodemod but I can't think of a better name right now
+# At least it is namespaced but we might want to deconflict these things in the long term
+class SemgrepCodemod(
+    _CodemodSubclassWithMetadata,
+    _SemgrepCodemod,
+    BaseTransformer,
+    Helpers,
+):
+    CHANGES_IN_FILE: List = []
+
+    def __init__(self, codemod_context: CodemodContext, file_context: FileContext):
+        _SemgrepCodemod.__init__(self, file_context)
+        BaseTransformer.__init__(
+            self,
+            codemod_context,
+            self._results,
+            file_context.line_exclude,
+            file_context.line_include,
+        )
+
+    # TODO: there needs to be a way to generalize this so that it applies
+    # more broadly than to just a specific kind of node. There's probably a
+    # decent way to do this with metaprogramming. We could either apply it
+    # broadly to every known method (which would probably have a big
+    # performance impact). Or we could allow users to register the handler
+    # for a specific node or nodes by means of a decorator or something
+    # similar when they define their `on_result_found` method.
+    # Right now this is just to demonstrate a particular use case.
+    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
+        pos_to_match = self.node_position(original_node)
+        if self.filter_by_result(
+            pos_to_match
+        ) and self.filter_by_path_includes_or_excludes(pos_to_match):
+            line_number = pos_to_match.start.line
+            self.CHANGES_IN_FILE.append(
+                Change(str(line_number), self.CHANGE_DESCRIPTION).to_json()
+            )
+            if (attr := getattr(self, "on_result_found", None)) is not None:
+                new_node = attr(updated_node)
+                # TODO: where does this actually belong?
+                RemoveImportsVisitor.remove_unused_import_by_node(
+                    self.context, original_node
+                )
+                return new_node
+
+        return updated_node
--- a/codemodder/codemods/secure_random.py
+++ b/codemodder/codemods/secure_random.py
@@ -1,65 +1,23 @@
-import libcst as cst
-from libcst.codemod import (
-    CodemodContext,
-)
-from typing import List
-
-from libcst.codemod.visitors import AddImportsVisitor, RemoveImportsVisitor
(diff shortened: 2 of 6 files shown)


drdavella:
Linter rules are failing at the moment but I'm not expecting to merge this in the immediate future so I will fix it a little later.

codemodder/codemods/secure_random.py
drdavella, line 1: Your review should start in this file. This is the end result that we are going for, and all of the other changes facilitate this API.

nahsra: This is so much more beautiful 👨‍🍳 💋

clavedeluna:
> I think I'm not a big fan of the use of class-level attributes, but I need to think about a better solution

I can at least explain the **why** here. It's purely to store state for codemodder runs. So when codemodder runs, We want to store all results for the same codemod for that codemod. The most "natural" OOP   way to do that was as a class attr, but as you've seen us discuss this, it's not great. Given the why, choosing some data storage / state keeping mechanism that works for codemodder would be the goal.

codemodder/api/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def _build_semgrep_codemod_class(orig, metadata: CodemodMetadata):
+    # This is a terrible, terrible (terrible) hack. We should fix it eventually
+    # but for now I need it to work with the existing class attribute contract
+    class has_yaml:
+        # TODO: handle multiple yaml files?
+        YAML_FILES = _create_temp_yaml_file(orig, metadata)
+
+    class cls(has_yaml, orig, SemgrepCodemod, BaseTransformer, Helpers):
```
clavedeluna: just puuurely to avoid confusion, could you rename it to `klas`?

codemodder/api/helpers.py
```diff
@@ -0,0 +1,22 @@
+import libcst as cst
+from libcst.codemod.visitors import AddImportsVisitor
+
+from codemodder.codemods.utils import get_call_name
+
+
+class Helpers:
+    def add_needed_import(self, import_name):
+        # TODO: do we need to check if this import already exists?
```
andrecsilva: `AddImportsVisitor` already checks that.

codemodder/api/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+        METADATA = metadata
+        CHANGES_IN_FILE: List = []
+
+        # This is a little bit hacky, but it also feels like the right solution?
+        CHANGE_DESCRIPTION = metadata.DESCRIPTION
```
clavedeluna: yeah we still haven't made a distinction between the overall codemod description versus each little change's description. fine for now

codemodder/api/helpers.py
```diff
@@ -0,0 +1,22 @@
...
+        AddImportsVisitor.add_needed_import(self.context, import_name)
```
andrecsilva: I've recently discovered that `AddImportsVisitor` will not work well if the file has imports outside of the "header".
Try this example:
```python
import random
random.random()

import os
```
It will produce:
```python
secrets.SystemRandom().random()

import os
import secrets
```

codemodder/codemods/semgrep/secure_random.yaml
```diff
@@ -4,9 +4,8 @@ rules:
...
-      - patterns:
-        - pattern: random.$F(...)
-        - pattern-inside: |
-            import random
-            ...
+    patterns:
+    - pattern: random.$F(...)
+    - pattern-inside: |
+        import random
+        ...
```
clavedeluna, line 11: are you able to entirely remove this file now?

clavedeluna:
things I love
- including the yaml directly in the rule instead of creating our own semgrep yaml file. why didn't we do this before!!! Yay Dan!

concerns
- as you probably saw when implementing this, specifically `leave_Call`, each rule is so subtly differet depending on what we're changing. while I love the API `self.update_call_target(target_node, "secrets.SecureRandom()")`, I'm not sure how generalizable it will be. But I'm on board to start this API and let it inform us and "force" us to generalize. The way we've been operating now is "just write rules, generalize later", but with the new direction coming out of the onsite, it seems like we need to flip that MO. I'm fine with that!

codemodder/api/__init__.py
andrecsilva: Locking this to `BaseTransformer` will make us unable to use other transformations before and after it. 

For reference, check the `UrlSandbox` and `RemoveUnusedImports` codemods for example of codemods that are not simple transformers.

codemodder/codemods/secure_random.py
```diff
@@ -1,65 +1,24 @@
...
-from codemodder.codemods.base_visitor import BaseTransformer
-from codemodder.codemods.utils import get_call_name
-from codemodder.file_context import FileContext
-
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.api import codemod
 
-system_random_object_name = "gen"
 
+@codemod(
```
andrecsilva: Probably the hottest take on this review, but this smells like Java code. Using multiple inheritance to enforce standards on a class is not only more explicit but it would be more natural in python.

My understanding is that usual use case of decorators is to "enhance" something without changing its implementation (directly at least). Here we are not only using a decorator to enforce patterns/standards on the class, but at first glance it looks like a decorator that only adds metadata.

codemodder/codemods/secure_random.py
```diff
@@ -1,65 +1,24 @@
...
-            new_call = cst.Call(
-                func=cst.Attribute(
-                    value=cst.parse_expression("secrets.SystemRandom()"),
-                    attr=cst.Name(value=get_call_name(original_node)),
-                ),
-                args=original_node.args,
-            )
-            return new_call
-        return updated_node
+    def on_result_found(self, original_node):
```
andrecsilva: Ever wondered why the  `leave_*` methods have two parameters for a node (`original_node` and `updated_node`)?

If you make changes to a child node of `original_node`, `original_node` won't contain that change, but `updated_node` will. So making changes directly to `original_node` here will erase any previous changes to its children.

The official docs recommends to always make changes to `updated_node`:
https://libcst.readthedocs.io/en/latest/visitors.html#libcst.CSTTransformer

andrecsilva (changes requested):
To put it bluntly: I don't like it. I can see something like this being used for a alternative for writing simple codemods, but I feel it will crumble with anything that is more complex than replacing a single call. Of course, this is only a problem if you're enforcing this particular standard to all codemods.

The only required change is the one pointed by the comment attached to `on_result_found`. The rest are opinions/suggestions/future work and may be promptly ignored.

nahsra:
> Of course, this is only a problem if you're enforcing this particular standard to all codemods.

I think this is the key part. Simple things should be simple, and complex things should be possible. We should name the codemod subtype according to the narrow pattern is it intended to support.

codemodder/api/__init__.py
clavedeluna, line 1: another idea I thought. The long term goal is to separate the codemodder repo (the thing that runs codemods, the engine) from codemods and codemod writing. So I think this new api dir should go under the codemods directory.

codemodder/codemods/semgrep/secure_random.yaml
drdavella: Yes it's not being used anymore but I kept it around for posterity. I'll remove it once we decide to go with this implementation.

codemodder/codemods/secure_random.py
drdavella: I don't have a particular commitment to using the decorator, although I do think it enabled the kind of metaprogramming that I wanted to do here. But I'd be okay with rephrasing this in terms of inheritance.

codemodder/codemods/secure_random.py
drdavella: Thanks, this is a good call-out. I will fix it.

drdavella:
> > I think I'm not a big fan of the use of class-level attributes, but I need to think about a better solution
> 
> I can at least explain the **why** here. It's purely to store state for codemodder runs. So when codemodder runs, We want to store all results for the same codemod for that codemod. The most "natural" OOP way to do that was as a class attr, but as you've seen us discuss this, it's not great. Given the why, choosing some data storage / state keeping mechanism that works for codemodder would be the goal.

Thanks for the explanation. Could we create some kind of simple data store (could even be a dict as first pass) and inject it into each codemod class using the context? It feels like this data should be centralized somehow instead of stored on each individual codemod.

codemodder/api/__init__.py
drdavella: That's fair. Let's just consider this a first-pass right now and we can revisit.

codemodder/api/helpers.py
drdavella: @andrecsilva this feels like something that needs to be fixed separately, correct?

drdavella:
> To put it bluntly: I don't like it. I can see something like this being used for a alternative for writing simple codemods, but I feel it will crumble with anything that is more complex than replacing a single call. Of course, this is only a problem if you're enforcing this particular standard to all codemods.

@andrecsilva I appreciate your honesty. You are right that it is very brittle as implemented. However, I think we should begin by looking at the very simplest codemods and figuring out the cleanest/clearest way to express those. It's possible that we already have another codemod or two that could be expressed simply in this way (maybe with a few additional helper methods) and I think we should push on those a bit while we iterate on the design.

Also, just to be clear, I also think there should always be a "hard" way to write these that is more expressive and more powerful. Many of our own codemods will need to be written using a more powerful API. But we want to make the easy cases easy.

drdavella:
The fact that we have these kinds of linter errors (and we have already suppressed some of them elsewhere) suggests to me that we could improve the design somehow. I think this is part of what makes me a little suspicious of mixins.
```
Using config file /home/runner/work/codemodder-python/codemodder-python/pylintrc
************* Module codemodder.codemods.secure_random
codemodder/codemods/secure_random.py:23:8: E1101: Instance of 'SecureRandom' has no 'add_needed_import' member (no-member)
codemodder/codemods/secure_random.py:24:15: E1101: Instance of 'SecureRandom' has no 'update_call_target' member (no-member)
************* Module codemodder.api.helpers
codemodder/api/helpers.py:10:44: E1101: Instance of 'Helpers' has no 'context' member (no-member)
************* Module codemodder.api
codemodder/api/__init__.py:86:31: E1102: attr is not callable (not-callable)
************* Module tests.codemods.test_secure_random
tests/codemods/test_secure_random.py:10:[15](https://github.com/pixee/codemodder-python/actions/runs/5970870880/job/16199112807?pr=5#step:6:16): E1101: Class 'SecureRandom' has no 'RULE_IDS' member (no-member)
************* Module integration_tests.test_secure_random
integration_tests/test_secure_random.py:[22](https://github.com/pixee/codemodder-python/actions/runs/5970870880/job/16199112807?pr=5#step:6:23):25: E1101: Class 'SecureRandom' has no 'CHANGE_DESCRIPTION' member (no-member)
```

codemodder/api/helpers.py
andrecsilva: Yes, I sorta know how to do it, but this is something that will probably take a few days to implement. I think we are fine for now, it's mostly an edge case.

drdavella:
I updated this PR to rephrase it in terms of inheritance rather than using a decorator. It required a fairly yucky workaround for the `__init_subclass__` implementations in `SemgrepCodemod` and `BaseTransformer`. 

While I understand the purpose and intent of those `__init_subclass__` methods and the behavior makes good sense to me, I also think that we have uncovered an unintended consequence of doing things that way, and we might want to revisit that implementation. 

The main problem is that `__init_subclass__` makes it practically impossible for anyone else to define a subclass in such a way that the required attributes are defined dynamically. This seems like a reasonable use case to me, and as you can see I had to jump through some hoops to make it work for this (interim) implementation. I think i really want to understand why these things are implemented in terms of attributes that are checked at class definition time rather than attributes and/or properties that are checked later.

Also, just for the record, the API introduced in this PR feels very provisional to me and will require some iteration and redesign over time. But I think it gets us close enough to a simplified example that we can feel comfortable publicizing (cc @nahsra).

codemodder/codemods/api/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+def _create_temp_yaml_file(orig_cls, metadata: CodemodMetadata):
+    fd, path = tempfile.mkstemp()
+    with os.fdopen(fd, "w") as ff:
+        ff.write(_populate_yaml(orig_cls.rule(), metadata))
+
+    return [path]
+
+
+class _CodemodSubclassWithMetadata:
+    def __init_subclass__(cls):
```
clavedeluna, line 43: I"m glad you can now see why we did this... I think if we go ahead and stop using class level attr and instead use instance vars only, which then leads us to assume that one codemod run will instantiate a codemod only once (ahem, singleton?), then we can remove all of this. Of course right now a codemod gets instantiated multiple times, one time per file in `run_codemods_for_file`.

It's crazy that we can't come up with a clean pattern for "I want a class pattern that will store data across all instances of it". It definitely sounds like a singleton, though I know it isn't our favorite pattern either.

andrecsilva (approved):
As long as we understand that this is a provisional solution, I'm fine with this.

While it feels hacky, the operating classes are there and it does give a general feeling of how those will be organized.

clavedeluna:
rebased for you but will let you merge. I can use this branch as a starting point

PR #8, opened 2023-08-28
Api limit readline

clavedeluna:
## Overview
*Transition limit readline codemod to new API*

## Description

Not very controversial changes here. I picked the smallest one so we can slowly iterate because I know some of these will be a challenge.

### DIFF
--- a/codemodder/codemods/limit_readline.py
+++ b/codemodder/codemods/limit_readline.py
@@ -1,50 +1,38 @@
 import libcst as cst
-from libcst.codemod import CodemodContext
-
-from codemodder.codemods.base_codemod import (
-    SemgrepCodemod,
-    CodemodMetadata,
-    ReviewGuidance,
-)
-from codemodder.codemods.base_visitor import BaseTransformer
-from codemodder.codemods.change import Change
-from codemodder.file_context import FileContext
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import SemgrepCodemod
 
 
 default_limit = "5_000_000"
 
 
-class LimitReadline(SemgrepCodemod, BaseTransformer):
-    METADATA = CodemodMetadata(
-        DESCRIPTION=("Adds a size limit argument to readline() calls."),
-        NAME="limit-readline",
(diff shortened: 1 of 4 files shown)


codemodder/codemods/limit_readline.py
```diff
@@ -1,50 +1,38 @@
...
+class LimitReadline(SemgrepCodemod):
+    NAME = "limit-readline"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW
+    DESCRIPTION = "Adds a size limit argument to readline() calls."
+
+    @classmethod
+    def rule(cls):
+        return """
+        rules:
+          - id: limit-readline
```
clavedeluna, line 18: had to make the str be this way for the api to work. There are prob other ways but  there is a todo around yaml generation so I Figured we will come back to this

codemodder/codemods/secure_random.py
```diff
@@ -18,6 +18,7 @@ def rule(cls):
                 ...
         """
 
-    def on_result_found(self, node):
+    def on_result_found(self, original_node, updated_node):
+        self.remove_unused_import(original_node)
```
drdavella, line 22: I'm a little unclear on the proper order of operations here because previously this was happening **after** the other transformations were applied. 

Also unclear on whether this should be `original_node` or `updated_node`.

drdavella (commented):
Looks good to me overall and I think it's a good step towards generalization. However I have some questions/comments and I'm not comfortable hitting approve until they are addressed.

codemodder/codemods/secure_random.py
andrecsilva: The original, since that's the one guaranteed to contain the reference to the import we detect with semgrep. 
About the order, it should not matter.

andrecsilva (approved):

drdavella (approved):

codemodder/codemods/secure_random.py
clavedeluna: 100% what Andre said.

PR #9, opened 2023-08-29
process creation uses new api

clavedeluna:
## Overview
*One more codemod, process creation, to rely on new API*

## Description

* minor changes that push us to generalization

## Documentation

## Performance

## Additional Details

### DIFF
--- a/codemodder/codemods/process_creation_sandbox.py
+++ b/codemodder/codemods/process_creation_sandbox.py
@@ -1,62 +1,39 @@
 import libcst as cst
-from libcst.codemod import CodemodContext
-from codemodder.file_context import FileContext
 from codemodder.dependency_manager import DependencyManager
-from libcst.codemod.visitors import AddImportsVisitor
-from codemodder.codemods.change import Change
-from codemodder.codemods.base_codemod import (
-    SemgrepCodemod,
-    CodemodMetadata,
-    ReviewGuidance,
-)
-from codemodder.codemods.base_visitor import BaseTransformer
-from codemodder.codemods.utils import get_call_name
-
-replacement_import = "safe_command"
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import SemgrepCodemod
 
 
-class ProcessSandbox(SemgrepCodemod, BaseTransformer):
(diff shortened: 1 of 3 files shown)


codemodder/codemods/process_creation_sandbox.py
```diff
@@ -1,62 +1,39 @@
...
-                    value=cst.Name(value=replacement_import),
-                    attr=cst.Name(value=get_call_name(original_node)),
-                ),
-                args=[cst.Arg(original_node.func), *original_node.args],
-            )
-            return new_call
-        return updated_node
+    def on_result_found(self, original_node, updated_node):
+        self.add_needed_import("security", "safe_command")
+        DependencyManager().add(["security==1.0.1"])
```
clavedeluna, line 33: I decided to keep the DM API as is, instead of creating a helper wrapper method, since this is already an API. We can decide later on

andrecsilva (approved):

codemodder/codemods/process_creation_sandbox.py
```diff
@@ -1,62 +1,39 @@
...
-            codemod_context,
-            self._results,
-            file_context.line_exclude,
-            file_context.line_include,
-        )
+    @classmethod
+    def rule(cls):
+        return """
+        rules:
+          - id: sandbox-process-creation
```
drdavella: I think you mentioned this in a previous PR but I'm not sure why we need this field. Also the name is inconsistent with `NAME` above.

I think if you remove the `id` here and change `NAME` to `sandbox-process-creation` it should work.

drdavella (changes requested):
Looks really nice overall. Just a minor change requested related to the rule ID/NAME field.

codemodder/codemods/process_creation_sandbox.py
clavedeluna: woops good catch on the incorrect name here! I was able to remove "id" here  by making the only thing under "rules" be `- pattern-either:`. limit-readline still needs the rule_id bc the yaml uses both `pattern-sources:` and `pattern-sinks:` . We will need to make the yaml generation more flexible.

drdavella (approved):

PR #10, opened 2023-08-29
create api.base_codemod & transtion fstr to api

clavedeluna:
## Overview
*This PR creates `BaseCodemod` under api, generalizes some of that code, and transitions unnecessary f-str codemod to the new api*

## Description

* I made some changes to the api to generalize it. Feel free to critique if you think it's going in the wrong direction. Basically I'm moving toward the original classes will be made redundant at some point and removed. I also am continuing the idea that SemgrepCodemod is a type of BaseCodemod. Maybe one day that won't be the case.
* hoping this new BaseClass will make it easy to add non-semgrep codemods with the new API

## Documentation

* tests passing as is means we're ✅ 

## Performance

## Additional Details
*removing test_multiple.py because it has given us more grief than benefit AND we're likely going toward only running one codemod at a time.


andrecsilva (approved):

codemodder/codemods/api/__init__.py
```diff
@@ -44,7 +46,7 @@ def __init_subclass__(cls):
         # But it is necessary to get around the fact that these fields are
         # checked by __init_subclass__ of the other parents of SemgrepCodemod
         # first.
-        if not cls.__name__ == "SemgrepCodemod":
+        if cls.__name__ not in ("BaseCodemod", "SemgrepCodemod"):
```
drdavella, line 49: This works for now but I do view this whole thing as a workaround for the time being.

drdavella (approved):
I think this looks good. I'm happy with incremental steps for now. I think as long as we keep moving in a sensible direction at some point we'll arrive at an API that we're all happy with.

PR #11, opened 2023-08-30
Packaging updates

drdavella:
## Overview
*Make some incremental updates to the packaging and metadata*

⚠️ This PR will probably cause conflicts with lots of in-progress work. I'm happy to hold off on this PR for a bit and rebase when we're ready, but I also don't want to wait too long. ⚠️ 

## Description
* This is an incremental step to make some updates to the structure and metadata of the Python package itself
* There are [good motivations](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure) for using src-based layouts, so I just went ahead and made that change
* One consequence of doing this is the package needs to be installed prior to testing and linting. This is a Good Thing™ since it means we are testing code in the way it is actually installed/used
  * I also needed to update the way we look for coverage
* Also it seems like the community is standardizing on the use of `pyproject.toml` for packaging metadata
  * This could enable us to explore alternative build tools like [poetry](https://python-poetry.org/) if we were so inclined
* I made an update so that our semgrep files are packaged using `package-data` and can be accessed using `importlib.resources`
* Made some other small cosmetic updates to pipeline steps

## Future Work
* I'm creating an epic to track this, but there are some more things we want to do:
  * Add more package metadata
  * Use script entry point for `codemodder` command instead of `python -m codemodder`
  * Build the package using `python -m build` and test the result using `twine`
  * Use tag-based versioning and releases
  * Release to PyPi


pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "codemodder-python"
+version = "0.57.0"
+requires-python = ">=3.9.0"
```
clavedeluna, line 8: should be 3.8 no?

pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+readme = "README.md"
+license = {file = "LICENSE"}
+dependencies = [
```
clavedeluna, line 11: I assume these are the dependencies that are currently in requirements/codemodder.txt? Or what is the relationship between this and that file? Can we delete the file?

tests/transformations/test_clean_imports.py
```diff
@@ -80,9 +80,8 @@ def test_organize_by_sections(self):
 
         import os
 
-        from third_party import Object
-
         from codemodder.codemods.transformations.clean_imports import CleanImports
+        from third_party import Object
```
clavedeluna: how did this change come about?

src/codemodder/semgrep.py
```diff
@@ -1,14 +1,15 @@
...
+from importlib.resources import files
 from tempfile import NamedTemporaryFile
 from typing import List
 from pathlib import Path
 from codemodder import global_state
 from codemodder.sarifs import results_by_path_and_rule_id
 from codemodder.logging import logger
 
-YAML_FILES_DIR = Path("codemodder") / "codemods" / "semgrep"
+YAML_FILES_DIR = files("codemodder.codemods.semgrep")
```
clavedeluna, line 12: this is magic ✨

clavedeluna (commented):
Once I can go ahead and merge the other 2 pending PRs, I think we can hold off and let you merge this one.

pyproject.toml
drdavella: Sorry I meant to leave a comment here but personally I'd like to just go with 3.9 unless there are strong objections. It narrows our support window, especially with 3.12 on the horizon, and I don't see a compelling reason to go back that far.

pyproject.toml
drdavella: Yes we can delete it 👍

tests/transformations/test_clean_imports.py
drdavella: Sorry, another place I meant to leave a comment. Short answer is: I have no idea. I don't think I touched any code that would have any affect on this. The new version looks reasonable/correct to me, but I'm just not sure.

clavedeluna: sure! I Don't think we have committed to 3.8 in any official capacity or in documentation so that's good with me.

andrecsilva (approved):
Sounds good. Just remember to update CONTRIBUTING.md to reflect these changes.

PR #12, opened 2023-08-30
transition yaml codemods to new api

clavedeluna:
## Overview
*This PR moves both pyyaml and ruamel codemods to use the new api*

## Description

* API change: `update_arg_target` takes a list of new args

## Performance

## Additional Details


codemodder/codemods/harden_pyyaml.py
```diff
@@ -1,53 +1,45 @@
...
+class HardenPyyaml(SemgrepCodemod):
+    NAME = "harden-pyyaml"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = "Ensures all calls to yaml.load use `SafeLoader`."
+
+    @classmethod
+    def rule(cls):
+        return """
+            rules:
+              - id: harden-pyyaml
```
drdavella: Can we omit the `id` field here?

codemodder/codemods/harden_pyyaml.py
```diff
@@ -1,53 +1,45 @@
...
+                      - metavariable-pattern:
+                          metavariable: $ARG
+                          patterns:
+                            - pattern-not:
+                                pattern: yaml.SafeLoader
+
+        """
+
+    def on_result_found(self, _, updated_node):
+        second_arg = cst.parse_expression("yaml.SafeLoader")
```
drdavella: Just a small thing but I think we should hide this behind a helper so that the user doesn't have to make reference to `libCST` if it's not necessary. It could maybe just become `self.parse_expression`.

codemodder/codemods/harden_ruamel.py
```diff
@@ -1,52 +1,38 @@
...
-        return updated_node
+        """
+
+    def on_result_found(self, original_node, updated_node):
+        new_args = make_new_args(original_node.args, target_arg="typ")
+        return self.update_arg_target(updated_node, new_args)
 
 
-def update_arg_target(original_args, target_arg):
+def make_new_args(original_args, target_arg):
```
drdavella: Is this more generally useful now? Could it move to the `Helpers` class?

codemodder/codemods/limit_readline.py
```diff
@@ -35,4 +35,4 @@ def rule(cls):
         """
 
     def on_result_found(self, _, updated_node):
-        return self.update_arg_target(updated_node, cst.Integer(default_limit))
+        return self.update_arg_target(updated_node, [cst.Integer(default_limit)])
```
drdavella, line 38: It probably wouldn't be too hard to keep the old API for single arguments but I don't feel too strongly about it right now.

codemodder/codemods/harden_pyyaml.py
drdavella: Also I think I was guilty of this myself in my original PR, but maybe we should remove the YAML files now.

drdavella (changes requested):
Looks good overall but just a handful of small comments.

codemodder/codemods/limit_readline.py
clavedeluna: I thought about that too but it felt more general to say we can support a list of args instead of one arg only.

codemodder/codemods/harden_pyyaml.py
clavedeluna: I like this thinking. Basically if a codemod has to `import cst` then we can likely move it to a helper.

codemodder/codemods/harden_ruamel.py
clavedeluna: good suggestion, did that

andrecsilva (approved):

drdavella (approved):

PR #13, opened 2023-08-31
new tempfile.mktemp codemod

clavedeluna:
## Overview
*Adding a new codemod to replace `mktemp` with `mkstemp`*

## Description

* see [doc](https://www.notion.so/pixee/Make-secure-tempfiles-5f461c5e3ae14d32bb02e5d2b31117e9)

* I'm following the new api as much as possible. The only important highlight here is the way I decided to handle the from import case. I made the decision to convert `from tempfile import mktemp` to `import tempfile` and the call to be used accordingly. I recognize ideally we would keep the from import, but doing so required work akin to what's happening in url_sandbox, and we're not there yet. Instead, I decided that this is sufficient and works. I'm starting to think this is a good approach to keep a simple api. It's better to have a codemod that makes a good change with a simple API than to have a super complex API that will make a "perfect" codemod.

## Documentation

## Performance

## Additional Details

### DIFF
--- a/tests/codemods/test_tempfile_mktemp.py
+++ b/tests/codemods/test_tempfile_mktemp.py
@@ -1,0 +1,82 @@
+import pytest
+from codemodder.codemods.tempfile_mktemp import TempfileMktemp
+from tests.codemods.base_codemod_test import BaseSemgrepCodemodTest
+
+
+class TestTempfileMktemp(BaseSemgrepCodemodTest):
+    codemod = TempfileMktemp
+
+    def test_rule_ids(self):
+        assert self.codemod.RULE_IDS == ["secure-tempfile"]
+
+    def test_import(self, tmpdir):
+        input_code = """import tempfile
+
+tempfile.mktemp()
+var = "hello"
+"""
+        expexted_output = """import tempfile
+
+tempfile.mkstemp()
+var = "hello"
+"""
+
+        self.run_and_assert(tmpdir, input_code, expexted_output)
+
+    def test_import_with_arg(self, tmpdir):
+        input_code = """import tempfile
+
+tempfile.mktemp('something')
+var = "hello"
+"""
+        expexted_output = """import tempfile
+
+tempfile.mkstemp('something')
+var = "hello"
+"""
+
+        self.run_and_assert(tmpdir, input_code, expexted_output)
+
+    def test_from_import(self, tmpdir):
+        input_code = """from tempfile import mktemp
+
+mktemp()
+var = "hello"
+"""
+        expexted_output = """import tempfile
+
+tempfile.mkstemp()
+var = "hello"
+"""
+        self.run_and_assert(tmpdir, input_code, expexted_output)
+
+    @pytest.mark.skip()
+    def test_import_alias(self, tmpdir):
+        input_code = """import tempfile as _tempfile
+
+_tempfile.mktemp()
+var = "hello"
+"""
+        expexted_output = """import tempfile as _tempfile
+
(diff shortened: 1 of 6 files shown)


drdavella (approved):
> Instead, I decided that this is sufficient and works. I'm starting to think this is a good approach to keep a simple api. It's better to have a codemod that makes a good change with a simple API than to have a super complex API that will make a "perfect" codemod.

I'm good with this for now. I do think we should continue to iterate on an approach that consolidates imports since it's a little cleaner overall, but it's not a big ticket item right now.

Also I think this is a pretty cool codemod. It would be good to find some other deprecations like this.

tests/codemods/test_tempfile_mktemp.py
```diff
@@ -0,0 +1,82 @@
...
+var = "hello"
+"""
+        expexted_output = """import tempfile
+
+tempfile.mkstemp()
+var = "hello"
+"""
+        self.run_and_assert(tmpdir, input_code, expexted_output)
+
+    @pytest.mark.skip()
```
andrecsilva, line 53: I'm gonna guess this breaks the code in this case. If so, you should change the rule/code to ignore calls like this.

andrecsilva (changes requested):

tests/codemods/test_tempfile_mktemp.py
clavedeluna: That's the point of adding skip, so it doesn't run. If I do run this test, nothing breaks, it just gives us the wrong output without the alias (so it's actually valid code, it just removes the aliasing)

clavedeluna: That's the point of adding skip, so it doesn't run. If I do run this test, nothing breaks, it just gives us the wrong output without the alias (so it's actually valid code, it just removes the aliasing)

drdavella: I'd have a slight preference for `xfail` in these cases.

clavedeluna: I prefer skip because it doesn't run the test. Xfail will run it and fail. Tiny time saver. Benefit to xfail if you think we would magically fix this without knowing, which I don't

andrecsilva: Sorry if I wasn't clear, but I meant if the code produced by this case is correct, as in, it compiles and preserves behavior.

I just tested and that is the case.

andrecsilva (changes requested):
Missing integration test.

tests/codemods/test_tempfile_mktemp.py
clavedeluna: I guess it does compile and is "correct" but it isn't what we would want with aliasing:

```
        expexted_output = """import tempfile as _tempfile

_tempfile.mkstemp()
var = "hello"
"""
```
instead the result is 
```
import tempfile

tempfile.mkstemp()
var = "hello"
```

which is fine  but the point of the unit test is to show we would correctly handle aliasing which we don't. 
I think it's interesting though that it does return legitimate code. So maybe we should become less strict with aliasing and accept that the user's aliasing will be overriden?

andrecsilva: Ideally we want to respect aliasing. We are inviting possible errors otherwise, (e.g. shadowing names). I believe those would be a bit rare though.

andrecsilva (approved):

clavedeluna:
we can revisit aliasing testing later on as a whole

PR #16, opened 2023-09-01
Order imports codemod

andrecsilva:
## Overview
Adds a codemod that order and formats imports like isort's default behavior.

Standard unit and integration tests.

## Additional Details

The overall transformation is quite complex and brought to light several idiosyncrasies of libcst. Some design decisions may seem odd.


tests/codemods/test_order_imports.py
```diff
@@ -0,0 +1,96 @@
+from tests.codemods.base_codemod_test import BaseCodemodTest
+from codemodder.codemods.order_imports import OrderImports
+
+
+class TestOrderImports(BaseCodemodTest):
```
drdavella, line 5: These are good tests but I think we need some additional ones for making sure the groupings are handled properly. So there should probably be some tests that include various combinations of `__future__`, standard library modules, etc.

codemodder/codemods/transformations/clean_imports.py
```diff
@@ -57,148 +57,272 @@ def __init__(self, context: CodemodContext, src_path):
...
+                    body=[cst.Import(names=[ia])], leading_lines=comments
+                )
+            )
+        return new_import_stmts
+
+    def _handle_from_imports(self, from_import_stmts):
+        imports_dict = defaultdict(lambda: (set(), []))
+        star_imports_dict = defaultdict()
+        # binsort the aliases into module name, merge comments
+        for stmt in from_import_stmts:
```
drdavella, line 170: Just a small suggestion but this function might read a little cleaner if each of these loops was broken into a separate method. Each comment above these loops suggests a new method name.

codemodder/codemods/order_imports.py
```diff
@@ -0,0 +1,80 @@
...
+            anchor_pos = self.node_position(anchor)
+            if self.filter_by_path_includes_or_excludes(anchor_pos):
+                filtered_blocks.append(block)
+        if filtered_blocks:
+            order_transformer = OrderImportsBlocksTransform(
+                codemodder.global_state.DIRECTORY, filtered_blocks
+            )
+            result_tree = tree.visit(order_transformer)
+
+            if any(order_transformer.changes):
```
drdavella: Just a small suggestion but this check feels redundant since it's also handled within the loop.

codemodder/codemods/transformations/clean_imports.py
```diff
@@ -247,22 +371,22 @@ def _triage_imports(all_imports, src_path):
...
     config = Config(src_paths=(src_path,))
 
-    for imp in all_imports:
+    for stmt in all_imports:
+        imp = stmt.body[0]
         section = isort.place.module(_get_name(imp), config)
 
         if section == isort.sections.FUTURE:
-            future.append(imp)
+            future.append(stmt)
```
drdavella: This isn't from your change specifically, but the multiple `elif`s here make this code feel like it would be better handled by a different data type.

codemodder/codemods/transformations/clean_imports.py
```diff
@@ -247,22 +371,22 @@ def _triage_imports(all_imports, src_path):
...
         elif section == isort.sections.FIRSTPARTY:
-            first_party.append(imp)
+            first_party.append(stmt)
         else:
-            third_party.append(imp)
+            third_party.append(stmt)
 
-    imports_by_section = [future, stdlib, third_party, first_party, local]
-    return imports_by_section
+    return [future, stdlib, third_party, first_party, local]
```
drdavella: Not a huge deal but it feels like we should really be returning a `dict` or custom class of some type here.

codemodder/codemods/transformations/clean_imports.py
```diff
@@ -57,148 +57,272 @@ def __init__(self, context: CodemodContext, src_path):
...
-    """
-    Remove given nodes from the tree.
-    """
+        imports_by_section = list(filter(lambda x: x, imports_by_section))
 
-    def __init__(self, nodes_to_remove: Set[cst.CSTNode]):
-        self.nodes_to_remove = nodes_to_remove
+        # add empty lines after each section
+        # for this, we add an empty line at the leading_lines of the first not at each section
+        for section in imports_by_section[1:]:
```
drdavella, line 281: The `[1:]` here is a bit magical and it's going to be hard to remember the rationale for this in the future.

drdavella (changes requested):
This looks good overall. My only substantive comment is that I think we need some more test cases. I also have some comments about places where refactoring might make the code a little easier to read.

One question: do you think we should respect `isort` directives in the code? Not necessarily for this PR, but eventually?

drdavella:
@andrecsilva just one additional comment but I noticed that you merged the `main` branch into this PR branch. I would like to avoid merges from upstream into topic branches as much as possible since it makes the git history quite difficult to understand, especially when bisecting for the purposes of fixing bugs.

In this particular case, it should be sufficient to just run `git rebase main` on this PR branch and force push. I think in general we should prefer `rebase` to `merge` for these purposes.

codemodder/codemods/transformations/clean_imports.py
andrecsilva: The comment could be a bit more clear, but essentially instead of adding an `EmptyLine` node right after the last node of each section, we're adding the node at the `leading_lines` at the first node of each section (thus the first one is skipped).

Reason being is that, in libcst, nodes with formatting information are usually owned by the node right after it. This way we avoid reporting possible changes when no visual changes happens.

codemodder/codemods/transformations/clean_imports.py
andrecsilva: Good call, it was done this way because the order matters (it's isort's default one). But a dict could make it more modular.

tests/codemods/test_order_imports.py
andrecsilva: Some of the tests included in the clean imports test cover some of these. I'll refactor them to be unit tests of this codemod.

tests/samples/unordered_imports.py
```diff
@@ -0,0 +1,22 @@
+#!/bin/env python
```
clavedeluna, line 1:
```suggestion
```

andrecsilva: This comment is there to catch the case where comments from the "header" of a file should not be attached to an import and thus should not be moved.

andrecsilva:
> This looks good overall. My only substantive comment is that I think we need some more test cases. I also have some comments about places where refactoring might make the code a little easier to read.
> 
> One question: do you think we should respect `isort` directives in the code? Not necessarily for this PR, but eventually?

We could, but from what I've seen, some of them would be pretty easy while others quite complex. I'd say we are good for now.

codemodder/codemods/base_visitor.py
```diff
@@ -18,7 +18,7 @@ def filter_by_result(self, pos_to_match):
 
     def filter_by_path_includes_or_excludes(self, pos_to_match):
         """
-        Returns True if the node, whose position in the file is pos_to_match, matches any of the lines specified in the path-includes or path-excludes flags.
+        Returns False if the node, whose position in the file is pos_to_match, matches any of the lines specified in the path-includes or path-excludes flags.
```
clavedeluna: I think True was correct?

codemodder/codemods/order_imports.py
```diff
@@ -0,0 +1,80 @@
...
+        if self.line_include:
+            return any(match_line(pos_to_match, line) for line in self.line_include)
+        return True
+
+    def node_position(self, node):
+        # See https://github.com/Instagram/LibCST/blob/main/libcst/_metadata_dependent.py#L112
+        return self.get_metadata(PositionProvider, node)
+
+
+def match_line(pos, line):
```
clavedeluna, line 79: I see this is also in remove unused imports, as well as many other methods in this file. I know you mentioned the duping is intentional, but maybe this one thing can be in utils?

codemodder/codemods/utils.py
```diff
@@ -1,9 +1,30 @@
...
+        replacements: dict[
+            cst.CSTNode,
+            Union[cst.CSTNode, cst.FlattenSentinel[cst.CSTNode], cst.RemovalSentinel],
+        ],
+    ):
+        self.replacements = replacements
+
+    def on_leave(self, original_node, updated_node):
+        if original_node in self.replacements.keys():
+            return self.replacements[original_node]
```
clavedeluna, line 24: may not be doing exactly what you are doing here but figured I'd share. I discovered `deep_replace` last week https://libcst.readthedocs.io/en/latest/nodes.html#libcst.CSTNode.deep_replace

andrecsilva: The problem with using `deep_replace` is that everytime you use it to replace a single node, you change the tree. This invalidates node comparisons (e.g. `node == other_node`) and is quite computationally expensive if you have to replace a bunch of nodes (this is the case here).

codemodder/codemods/order_imports.py
andrecsilva: Needs to be refactored. For now inheriting from `UtilsMixin` will not work because of how it calls `super().__init__()`. Will do this on another PR.

codemodder/codemods/base_visitor.py
andrecsilva: I don't remember why I did this. Probably when trying to refactor this somehow. Will correct.

codemodder/codemods/order_imports.py
drdavella: While it has its uses, I'm not in love with the mixin pattern: I wonder whether it makes more sense just to have a utility module with functions rather than implementing these in terms of classes. Obviously this function otherwise needs no reference to `self`, so it's not clear to me that it needs to be a method anyway.

tests/transformations/test_clean_imports.py
```diff
@@ -1,141 +0,0 @@
-from libcst.codemod import CodemodTest
-from codemodder.codemods.transformations.clean_imports import CleanImports
-from pathlib import Path
-
-
-class TestCleanImports(CodemodTest):
```
drdavella, line 6: It feels weird to delete this entirely since it seems like `CleanImports` is logically distinct from `OrderImports`. I don't want to get bogged down right now but maybe this suggests that there is a slightly better design that better separates the concerns?

drdavella (approved):
This looks good for now. I think there are probably some more opportunities to clean up and simplify the implementation a bit, but I don't want to get bogged down right now since we need to keep moving forward on other codemods.

codemodder/codemods/order_imports.py
andrecsilva: Mixin makes a bit more sense when you realize that most of these functionalities hails from libcst metadata. The metadata needs to be calculated at instantiation and they are a bit expensive.
Using the Mixin can make it more modular with regards on how metadata are calculated (i.e. we only use metadata whose functionality are used in some mixin).

tests/transformations/test_clean_imports.py
andrecsilva: I thought about that for a bit and, ultimately, the `CleanImports` transform is essentially a combination of `OrderImports` and `RemoveUnusedImports` so the testing is mostly covered by those. 
Honestly it may be the case that we could do away with `CleanImports`. It's not being used anymore.

PR #17, opened 2023-09-04
Codemod to check verify flag in requests

clavedeluna:
## Overview
*This PR adds a codemod to check (and flip, if needed) the `verify` flag passed to `requests.func_that_uses_this_flag` and some important changes to codemodder, too*

## Description

* docs for codemod are [here](https://www.notion.so/pixee/Set-request-s-verify-flag-to-True-2470dd4e52de47a695f0cf3706817052)

* while this codemod could've instead been done within the url-sandbox codemod instead of on its own, we decided that we're uncertain how users will react to url-sandbox in the first place. So it made sense to have it separate for now.
* Since this new codemod changes `requests.FUNC`, it was the first time I saw one codemod changing multiple files. This showed that we were mistaken in storing changes to each file as a global for the codemod, as this made the results be shared for all files. Instead, I moved these results to file context and things work as expected now.

## Documentation

* necessary changes to integration tests to allow flexibility for codemodder to change more than one file. But we just assert the data for the first one.
* for this new codemod, I tested cases when verify is the default value, when it's True or a string, and when it's False.

## Performance

## Additional Details

### DIFF
--- a/tests/codemods/test_request_verify.py
+++ b/tests/codemods/test_request_verify.py
@@ -1,0 +1,103 @@
+import pytest
+from codemodder.codemods.requests_verify import RequestsVerify
+from tests.codemods.base_codemod_test import BaseSemgrepCodemodTest
+
+REQUEST_FUNCS = ["get", "post", "request"]
+
+
+class TestRequestsVerify(BaseSemgrepCodemodTest):
+    codemod = RequestsVerify
+
+    def test_rule_ids(self):
+        assert self.codemod.RULE_IDS == ["requests-verify"]
+
+    @pytest.mark.parametrize("func", REQUEST_FUNCS)
+    def test_default_verify(self, tmpdir, func):
+        input_code = f"""import requests
+
+requests.{func}("www.google.com")
+var = "hello"
+"""
+        self.run_and_assert(tmpdir, input_code, input_code)
+
+    @pytest.mark.parametrize("func", REQUEST_FUNCS)
+    @pytest.mark.parametrize("verify_val", ["True", "'/some/path'"])
+    def test_verify(self, tmpdir, verify_val, func):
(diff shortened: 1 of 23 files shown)


codemodder/codemods/__init__.py
```diff
@@ -12,6 +12,7 @@
 from codemodder.codemods.process_creation_sandbox import ProcessSandbox
 from codemodder.codemods.remove_unnecessary_f_str import RemoveUnnecessaryFStr
 from codemodder.codemods.tempfile_mktemp import TempfileMktemp
+from codemodder.codemods.requests_verify import RequestsVerify
```
drdavella: Not for this PR but I think we can/should do something to simplify this pattern. Maybe codemods should automatically add themselves to a registry (probably using a metaclass and/or overriding `__new__`).

tests/codemods/test_request_verify.py
```diff
@@ -0,0 +1,89 @@
+import pytest
+from codemodder.codemods.requests_verify import RequestsVerify
+from tests.codemods.base_codemod_test import BaseSemgrepCodemodTest
+
+REQUEST_FUNCS = ["get", "post", "request"]
+
+
+class TestRequestsVerify(BaseSemgrepCodemodTest):
```
drdavella, line 8: These are good tests, but just for my own sanity could we add the following:
1. A test with multiple kwargs
2. A test where `verify` is not the first kwarg
(I see this is in the integration test but it would be good for unit tests as well).

drdavella (changes requested):
Looks very nice overall and thanks for the fix to `CHANGES_IN_FILE`. I have just a small request for more test cases.

drdavella (approved):

andrecsilva (approved):

PR #19, opened 2023-09-05
Implement codemod for sslcontext-minimum-version

drdavella:
## Overview
*Implement new codemod for `sslcontext-minimum-version`*

## Description

* This fixes the case where `SSLContext.minimum_version` is set to an insecure version of SSL/TLS
* In theory the `maximum_version` could also be set to a bad version. This is a bit of a philosophical question but it feels like that should be handled by a separate codemod

### DIFF
--- a/integration_tests/test_upgrade_sslcontext_minimum_version.py
+++ b/integration_tests/test_upgrade_sslcontext_minimum_version.py
@@ -1,0 +1,23 @@
+from codemodder.codemods.upgrade_sslcontext_minimum_version import (
+    UpgradeSSLContextMinimumVersion,
+)
+from integration_tests.base_test import (
+    BaseIntegrationTest,
+    original_and_expected_from_code_path,
+)
+
+
+class TestUpgradeSSLContextMininumVersion(BaseIntegrationTest):
+    codemod = UpgradeSSLContextMinimumVersion
+    code_path = "tests/samples/upgrade_sslcontext_minimum_version.py"
+    original_code, expected_new_code = original_and_expected_from_code_path(
+        code_path,
+        [
+            (1, "import ssl\n\n"),
+            (7, "my_ctx.minimum_version = ssl.TLSVersion.TLSv1_2\n"),
+        ],
+    )
+
+    expected_diff = '--- \n+++ \n@@ -1,8 +1,9 @@\n from ssl import PROTOCOL_TLS_CLIENT, SSLContext, TLSVersion\n+import ssl\n \n my_ctx = SSLContext(protocol=PROTOCOL_TLS_CLIENT)\n \n print("FOO")\n \n my_ctx.maximum_version = TLSVersion.MAXIMUM_SUPPORTED\n-my_ctx.minimum_version = TLSVersion.TLSv1_1\n+my_ctx.minimum_version = ssl.TLSVersion.TLSv1_2\n'
+    expected_line_change = "8"
+    change_description = UpgradeSSLContextMinimumVersion.CHANGE_DESCRIPTION
--- a/src/codemodder/codemods/api/__init__.py
+++ b/src/codemodder/codemods/api/__init__.py
@@ -121,3 +121,16 @@ class SemgrepCodemod(
                 return new_node
 
         return updated_node
+
+    def leave_Assign(self, original_node, updated_node):
+        pos_to_match = self.node_position(original_node)
+        if self.filter_by_result(
+            pos_to_match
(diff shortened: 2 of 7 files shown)


src/codemodder/codemods/api/__init__.py
```diff
@@ -122,3 +122,16 @@ def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
                 return new_node
 
         return updated_node
+
+    def leave_Assign(self, original_node, updated_node):
```
drdavella, line 125: Adding individual handlers like this does not feel like a viable long-term solution.

clavedeluna: The goal is that we're going to get to a point that we have all of these cases. Right now most (all?) our api based codemods related to a func/method call so  leave_Call is appropriate. Your case is new so we need this. I think the django debug / session cookies codemods are similar and maybe we need to move those to the new api too

tests/codemods/test_upgrade_sslcontext_minimum_version.py
```diff
@@ -0,0 +1,58 @@
...
+        This codemod should not update maximum_version
+
+        (Maybe that should be a job for a separate codemod)
+        """
+        input_code = """from ssl import SSLContext, TLSVersion
+
+context = SSLContext()
+context.maximum_version = TLSVersion.TLSv1
+"""
+        self.run_and_assert(tmpdir, input_code, input_code)
```
clavedeluna, line 58: would be good to add a `test_random_alias`, passing or skipped if not

integration_tests/test_upgrade_sslcontext_minimum_version.py
```diff
@@ -0,0 +1,14 @@
+from codemodder.codemods.upgrade_sslcontext_minimum_version import (
+    UpgradeSSLContextMinimumVersion,
+)
+from integration_tests.base_test import BaseIntegrationTest
+
+
+class TestUpgradeSSLContextMininumVersion(BaseIntegrationTest):
+    codemod = UpgradeSSLContextMinimumVersion
+    code_path = "tests/samples/upgrade_sslcontext_minimum_version.py"
+    original_code = 'from ssl import PROTOCOL_TLS_CLIENT, SSLContext, TLSVersion\n\nmy_ctx = SSLContext(protocol=PROTOCOL_TLS_CLIENT)\n\nprint("FOO")\n\nmy_ctx.maximum_version = TLSVersion.MAXIMUM_SUPPORTED\nmy_ctx.minimum_version = TLSVersion.TLSv1_1\n'
```
clavedeluna: simplify using `original_and_expected_from_code_path` as other int tests do

clavedeluna (changes requested):

src/codemodder/codemods/api/__init__.py
drdavella: I'm hoping that we can handle this via clever metaprogramming somehow (maybe with `__getattr__`?) rather than having to override each possible method explicitly.

andrecsilva: Yes, can be done with metaprogramming. Either use the `__dict__` directly (and recursively over all the `mro()` classes), or use the `getmembers` from the `inspect` module.

clavedeluna (approved):

src/codemodder/codemods/upgrade_sslcontext_minimum_version.py
```diff
@@ -0,0 +1,36 @@
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import SemgrepCodemod
+
+
+class UpgradeSSLContextMinimumVersion(SemgrepCodemod):
+    NAME = "upgrade-sslcontext-minimum-version"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = "Replaces minimum SSL/TLS version for SSLContext"
+
+    @classmethod
```
andrecsilva, line 10: If you want to track an object, use taint mode:
https://semgrep.dev/playground/s/eAoX

See also: limit-readline codemod

andrecsilva (changes requested):

src/codemodder/codemods/api/__init__.py
drdavella: Yes that is another possibility. There are a variety of ways to approach this problem.

andrecsilva (approved):

PR #22, opened 2023-09-06
Run tests separately in CI

clavedeluna:
## Overview
*Separate unit test from integration test in CI run to speed things up*

## Description

* Integration tests cannot be run with xdist in parallel because they touch files in the disk. That's ok!
* Unti tests have no problem using xdist. By running them separately with xdist, the total run for testing goes down by almost half! 

🚀

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,10 +33,10 @@ jobs:
         run: pip install .
       - name: Install Dependencies
         run: pip install -r requirements/test.txt
-      - name: Run pytest
-        run: pytest -v --cov=codemodder
-      - name: Report coverage
-        run: coverage report -m --fail-under=95
+      - name: Run unit tests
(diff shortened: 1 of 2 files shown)


andrecsilva (approved):

.github/workflows/test.yml
```diff
@@ -33,10 +33,10 @@ jobs:
         run: pip install .
       - name: Install Dependencies
         run: pip install -r requirements/test.txt
-      - name: Run pytest
-        run: pytest -v --cov=codemodder
-      - name: Report coverage
-        run: coverage report -m --fail-under=95
+      - name: Run unit tests
+        run: pytest -v --cov-fail-under=95 --cov=codemodder tests --numprocesses auto
+      - name: Run integration tests
```
drdavella, line 38: Eventually I think this could just happen in a different action but for now this looks good.

drdavella (approved):

PR #24, opened 2023-09-06
Add codemodder entry point script

drdavella:
## Overview
*Add `codemodder` entry point script*

## Description

* This is mainly motivated by the need to integrate with the codemodder orchestrator
* Eventually I want to deprecate the `python -m codemodder` CLI, which I think will allow us to simplify a lot of the code around argument parsing, etc.

### DIFF
--- a/src/codemodder/__main__.py
+++ b/src/codemodder/__main__.py
@@ -192,6 +192,18 @@ def run(argv, original_args) -> int:
     return 0
 
 
-if __name__ == "__main__":
+def main():
+    # TODO: I'm not sure why this needs to be parsed out separately
+    # Maybe it has something to do with the invocation as python -m codemodder.
+    # But I think we should deprecate that interface which should simplify this.
     sys_argv = sys.argv[1:]
     sys.exit(run(parse_args(sys_argv), sys_argv))
+
+
+if __name__ == "__main__":
+    import warnings
+
+    warnings.warn(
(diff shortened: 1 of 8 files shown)


src/codemodder/__main__.py
```diff
@@ -192,6 +192,14 @@ def run(argv, original_args) -> int:
     return 0
 
 
-if __name__ == "__main__":
+def run_script():
+    # TODO: I'm not sure why this needs to be parsed out separately
+    # Maybe it has something to do with the invocation as python -m codemodder.
```
clavedeluna, line 197: IIRC it's to make testing easier, which isn't the best reason in the world. I do think we should go ahead and deprecate this and update the tests, but if you want to do that later that's cool with me.

src/codemodder/__main__.py
```diff
@@ -192,6 +192,14 @@ def run(argv, original_args) -> int:
     return 0
 
 
-if __name__ == "__main__":
+def run_script():
```
clavedeluna: what about calling it `main`? don't love the wording with script

clavedeluna (approved):

src/codemodder/__main__.py
drdavella: Yeah that's a better idea.

src/codemodder/__main__.py
drdavella: I will update the tests 👍

PR #26, opened 2023-09-07
Run codemodder by codemod name or id

clavedeluna:
## Overview
*Not only should CLI accept codemod name or id, but codemodder should actually RUN with either, too*

## Description

* Previous PR made the CLI work, but I forgot to match codemods by their id, too.
* Added unit tests to demonstrate this does work.

## Additional Details

### DIFF
--- a/src/codemodder/codemods/__init__.py
+++ b/src/codemodder/codemods/__init__.py
@@ -53,10 +53,12 @@ def match_codemods(codemod_include: list, codemod_exclude: list) -> dict:
             name: codemod
             for codemod in DEFAULT_CODEMODS
             if (name := codemod.name()) not in codemod_exclude
+            and (_ := codemod.id()) not in codemod_exclude
         }
 
     return {
         name: codemod
         for codemod in DEFAULT_CODEMODS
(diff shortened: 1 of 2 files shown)


src/codemodder/codemods/__init__.py
```diff
@@ -53,10 +53,12 @@ def match_codemods(codemod_include: list, codemod_exclude: list) -> dict:
             name: codemod
             for codemod in DEFAULT_CODEMODS
             if (name := codemod.name()) not in codemod_exclude
+            and (_ := codemod.id()) not in codemod_exclude
```
drdavella, line 56: Just curious but is there a reason to write it this way rather than just `codemod.id() not in codemod_exclude`?

drdavella (approved):

drdavella:
I'm going to merge this since my work today probably depends on this.

src/codemodder/codemods/__init__.py
clavedeluna: 🤦 of course it should just be `codemod.id() not in codemod_exclude`

If you get a PR in shortly you can go ahead and fix this here and line below

PR #27, opened 2023-09-07
new jwt decode verify codemod

clavedeluna:
## Overview
*Add a codemod that will check either the `verify` (soon to be deprecated) flag to jwt.decode and/or the multiple `verify_*` kwargs in the `options` param*

## Description

* see [docs](https://www.notion.so/pixee/Verified-JWT-decode-2a2a10fda3b8427582bcdc46dfd6fa4d)
* We already had an api solution to flip to `verify=True`, but for `options={...verify...}` I had to write some codemod-specific code. Maybe it can be generalized later on.
* I also think I have covered many edge cases, such as other keys to `options`, a mixture of verify flag and options, etc.

## Additional Details

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Install Dependencies
         run: pip install -r requirements/lint.txt
       - name: Black Format Check
-        run: black --check .
+        run: black --check . --exclude samples/
       - name: Run pylint
         run: make lint
 
--- a/src/codemodder/codemods/jwt_decode_verify.py
+++ b/src/codemodder/codemods/jwt_decode_verify.py
@@ -1,0 +1,86 @@
+import libcst as cst
+from libcst import matchers
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import SemgrepCodemod
+
+
+class JwtDecodeVerify(SemgrepCodemod):
+    NAME = "jwt-decode-verify"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = "Makes any of the multiple `verify` parameters to a `jwt.decode` call be `True`."
+
+    @classmethod
+    def rule(cls):
+        return r"""
+            rules:
+                - pattern-either:
+                  - patterns:
+                      - pattern: jwt.decode(..., verify=False, ...)
+                      - pattern-inside: |
+                          import jwt
+                          ...
+                  - patterns:
+                      - pattern-regex:  |-
+                            jwt.decode\(.*options={.*"verify_(signature|exp|nbf|iat|aud|iss)": False.*}.*\)
+                      - pattern-inside: |
+                          import jwt
+                          ...
+        """
+
+    def _replace_opts_dict(self, opts_dict):
+        new_dict_elements = []
+
+        for element in opts_dict.elements:
+            if is_verify_keyword(element):
+                new_el = cst.DictElement(
+                    key=cst.parse_expression(element.key.value),
+                    value=cst.parse_expression("True"),
+                )
+            else:
+                new_el = element
+            new_dict_elements.append(new_el)
+        return new_dict_elements
+
+    def replace_options_arg(self, node_args):
+        new_args = []
+        for arg in node_args:
+            if matchers.matches(arg.keyword, matchers.Name("options")) and isinstance(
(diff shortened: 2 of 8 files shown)


clavedeluna:
@drdavella This is as ready as I can get it. I'm hoping the pipeline will pass. You can take it over while I'm out for 2 weeks (which would be good because then we can add this codemod to docs.pixee) or wait for me to come back and cleanup whatever minor thing is there.

.github/workflows/lint.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Install Dependencies
         run: pip install -r requirements/lint.txt
       - name: Black Format Check
-        run: black --check .
+        run: black --check . --exclude samples/
```
clavedeluna, line 34: I think this and the corresponding pre-commit-hook update is the right call. There will be other instances when we want to test multiline and similar things and we don't want black to format it if our integration tests are expecting non-formatted code.

src/codemodder/codemods/jwt_decode_verify.py
```diff
@@ -0,0 +1,86 @@
...
+            rules:
+                - pattern-either:
+                  - patterns:
+                      - pattern: jwt.decode(..., verify=False, ...)
+                      - pattern-inside: |
+                          import jwt
+                          ...
+                  - patterns:
+                      - pattern-regex:  |-
+                            jwt.decode\(.*options={.*"verify_(signature|exp|nbf|iat|aud|iss)": False.*}.*\)
```
andrecsilva: There are a lot of problem by just using a plain regex here (whitespaces and quotes, to name a couple). 

I'd suggest following semgrep doc's example on how to match dict keys:
https://semgrep.dev/docs/writing-rules/pattern-examples/

andrecsilva (changes requested):

src/codemodder/codemods/jwt_decode_verify.py
clavedeluna: I don't think this is half bad as it looks, given that we rely on matching that exact pattern or not matching at all. But good find! Going to see if I can adapt it.

.github/workflows/lint.yml
clavedeluna: This change ended up not being needed for my PR but it still seems useful.

src/codemodder/codemods/jwt_decode_verify.py
```diff
@@ -0,0 +1,90 @@
...
+                      - pattern: jwt.decode(..., verify=False, ...)
+                      - pattern-inside: |
+                          import jwt
+                          ...
+                  - patterns:
+                      - pattern: |
+                          jwt.decode(..., options={..., $KEY: False, ...}, ...)
+                      - metavariable-regex:
+                          metavariable: $KEY
+                          regex: \"verify_
```
andrecsilva: This still fails for single quotes and similar alternatives. You can capture the string itself by using `"$KEY"`. Semgrep will take care of the string delimiters.

andrecsilva (changes requested):

src/codemodder/codemods/jwt_decode_verify.py
clavedeluna: You're right! Glad to have you as our semgrep expert, this is the hardest part for me! I've updated and added a unit test to confirm double/single quotes work.

andrecsilva (approved):

src/codemodder/codemods/jwt_decode_verify.py
```diff
@@ -0,0 +1,90 @@
...
+                value='"verify_signature"',
+                lpar=[],
+                rpar=[],
+            )
+            ...
+    where value should be anything with the word `verify`
+    """
+    return (
+        matchers.matches(element.key, matchers.SimpleString())
+        and "verify" in element.key.value
```
drdavella, line 89: Any chance this should be `element.key.value.startswith("verify")`?

drdavella (approved):
Looks very nice. I think this is a good codemod. Just one small question/comment.

src/codemodder/codemods/jwt_decode_verify.py
clavedeluna: don't see why not, good idea!

clavedeluna: turns out doing `startswith` isn't ideal, because sometimes the value is  `'"verify_signature"'` sometimes it's `"'verify_signature'"` so to avoid dealing with those variations I will just keep the original

PR #30, opened 2023-09-11
Enforce SUMMARY field for codemod; add to codetf2 output

drdavella:
## Overview
*Enforce `SUMMARY` field and add to codetf2 output*

## Description
* CodeTF2 expects every result to have a `description` _and_ `summary` field
* We need to enforce this on both the simple API and internally
* I'm not really happy with the current state of this. We need to revisit the design of these classes and the metadata fields. But I need this change in order to make progress with orchestrator work


ryandens (approved):
🎉

PR #31, opened 2023-09-11
Use relative paths for codetf results

drdavella:
## Overview
*Use relative paths when reporting results in CodeTF2*

## Description
* The [spec](https://github.com/pixee/codemodder-specs/blob/b98a60689a0ff159bafb1a250aa4834a0a5c81ca/codetf.json#L34) requires that paths in results must be relative to the top of the repository

### DIFF
--- a/src/codemodder/__main__.py
+++ b/src/codemodder/__main__.py
@@ -74,7 +74,8 @@ def run_codemods_for_file(
 
             codemod_kls.CHANGESET_ALL_FILES.append(
                 ChangeSet(
-                    str(file_context.file_path),
+                    # TODO: we should not be using global state here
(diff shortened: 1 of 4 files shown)


src/codemodder/__main__.py
```diff
@@ -74,7 +74,8 @@ def run_codemods_for_file(
 
             codemod_kls.CHANGESET_ALL_FILES.append(
                 ChangeSet(
-                    str(file_context.file_path),
+                    # TODO: we should not be using global state here
+                    str(file_context.file_path.relative_to(global_state.DIRECTORY)),
```
drdavella, line 78: This is the real change; everything else is just for making tests pass.

dunningdan (approved):

PR #33, opened 2023-09-13
Prevent codemod bleedover

drdavella:
## Overview
*Fix problem where all supported codemod IDs are (incorrectly) added to CodeTF file*

## Description

* We expect only applicable codemod IDs to be present in CodeTF output when running with a list of codemod IDs
  * Prior to this, all supported codemod IDs would be added to CodeTF, which breaks the contract with callers
* The problem appears to be related to the fact that results were stored as class-level attributes on each codemod class
* To fix this, I added a new `CodemodExecutionContext` that is intended to store process-level state
  * Eventually this will be used to store other state that is currently represented using globals/singletons
* I wasn't able to remove all global/shared state in this PR but I intend to open another PR in the near future

### DIFF
--- a/src/codemodder/__main__.py
+++ b/src/codemodder/__main__.py
@@ -11,29 +11,14 @@ from codemodder.logging import logger
 from codemodder.cli import parse_args
 from codemodder.code_directory import file_line_patterns, match_files
 from codemodder.codemods import match_codemods
-from codemodder.codemods.change import Change
+from codemodder.context import CodemodExecutionContext, ChangeSet
 from codemodder.dependency_manager import DependencyManager
 from codemodder.report.codetf_reporter import report_default
 from codemodder.semgrep import run as semgrep_run
 from codemodder.sarifs import parse_sarif_files
 
 # Must use from import here to point to latest state
-from codemodder import global_state
-
-RESULTS_BY_CODEMOD = []
-from dataclasses import dataclass
-
-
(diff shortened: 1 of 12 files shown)


src/codemodder/__main__.py
```diff
@@ -54,13 +40,16 @@ def run_codemods_for_file(
...
+            # TODO: eventually pass execution context here
+            # It will be used for things like dependency management
             file_context,
         )
         if not codemod.should_transform:
             continue
 
         logger.info("Running codemod %s for %s", name, file_context.file_path)
         output_tree = codemod.transform_module(source_tree)
+        # TODO: there has got to be a more efficient way to check this?
```
andrecsilva, line 52: We could compare/diff the generated code (i.e. source_tree.code == output_tree.code). However, it is possible that the code is the same while having distinct trees. Not that is matters to our use case.

andrecsilva (approved):

src/codemodder/__main__.py
drdavella: Couldn't we set some kind of flag on the `FileContext` from within the codemod itself?

andrecsilva: If we want to be reductionist, we could simply check if any changes were added to the changeset. However that would fail to catch codemods that make changes but don't report. Of course, those can be considered bugs, but those happen somewhat often.

PR #34, opened 2023-09-14
Https connection codemod

andrecsilva:
## Overview
Adds a codemod that changes `HTTPConnectionPool` to `HTTPSConnnectionPool` for the `urllib3` module.
Also adds a small API for resolving names of calls.

### DIFF
--- a/src/codemodder/codemods/https_connection.py
+++ b/src/codemodder/codemods/https_connection.py
@@ -1,12 +1,25 @@
-from libcst.metadata import PositionProvider, ScopeProvider
+from libcst import matchers
+from libcst.codemod.visitors import AddImportsVisitor, RemoveImportsVisitor
+from libcst.helpers import get_full_name_for_node
+from libcst.metadata import (
+    Assignment,
+    ImportAssignment,
+    PositionProvider,
+    ScopeProvider,
+)
 from codemodder.codemods.base_codemod import (
     BaseCodemod,
     CodemodMetadata,
     ReviewGuidance,
 )
+from codemodder.codemods.change import Change
 from codemodder.file_context import FileContext
 import libcst as cst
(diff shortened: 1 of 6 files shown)


src/codemodder/codemods/https_connection.py
```diff
@@ -0,0 +1,125 @@
...
+        self.line_include = file_context.line_include
+        self.changes_in_file: list[Change] = []
+
+    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
+        pos_to_match = self.node_position(original_node)
+        line_number = pos_to_match.start.line
+        if self.filter_by_path_includes_or_excludes(pos_to_match):
+            true_name = self.find_base_name(original_node.func)
+            if (
+                self._is_direct_call_from_imported_module(original_node)
```
drdavella, line 67: It would be cool to generalize this with `filter_by_result` somehow.

drdavella (approved):
This looks very nice overall. My only comment is that if we want to write more codemods like this (without using semgrep) it would be nice to generalize the detection part of it a little more so that it could be reused without much repetition.

src/codemodder/codemods/https_connection.py
andrecsilva: That is on my (personal) backlog. But it is not happening before we start cleaning up the overall API (specifically our `__init__` calls).

andrecsilva:
> This looks very nice overall. My only comment is that if we want to write more codemods like this (without using semgrep) it would be nice to generalize the detection part of it a little more so that it could be reused without much repetition.

Yes, it is a bit "raw" at the moment, but I plan on integrating it better to our API later. The issue with `AddImportsVisitor` will probably have some impact on it, so I want to finish that first before having a shot at this.

PR #35, opened 2023-09-14
Remove global state

drdavella:
## Overview
*Remove the use of global state; other miscellaneous code cleanup*

## Description

* The main change here is the removal of global state for storing the working directory
* The `DependencyManager` also effectively stored global state since it's implemented in terms of a `Singleton`
* All codemods now receive a `CodemodExecutionContext` which is used to store centralized state
* This PR also fixes a circular dependency that was exposed when running tests
* I also introduced a Makefile to consolidate some of our testing and linting commands


pyproject.toml
```diff
@@ -23,3 +23,7 @@ codemodder = "codemodder.__main__:main"
 
 [tool.setuptools.package-data]
 "codemodder.codemods.semgrep" = ["src/codemodder/codemods/semgrep/*.yaml"]
+
+[tool.pytest.ini_options]
+# Ignore integration tests and ci tests by default
+testpaths = ["tests"]
```
drdavella, line 29: We could include integration tests here if we want but mainly I want to avoid running the webgoat tests by accident.

andrecsilva (approved):

PR #40, opened 2023-09-19
Implement codemod for walrus-if

drdavella:
## Overview
*Implement codemod to use walrus (`:=`) operator for simple cases*

## Description
* This codemod makes changes of the following form:
```diff
- x = foo()
- if x is not None:
+ if (x := foo()) is not None:
    print(x)
```
* It looks for simple single-variable assignments that are immediately followed by an `if` statement that checks the value of the assignment. Right now we are prioritizing accuracy and safety over breadth
* Semgrep does most of the heavy lifting for detection but we also need some additional validation using libCST to avoid some false negatives
* The codemod attempts to preserve whitespace and comments where possible
* Future codemods will try to find other cases where the walrus can be used

### DIFF
--- a/integration_tests/test_use_walrus_if.py
+++ b/integration_tests/test_use_walrus_if.py
@@ -1,0 +1,32 @@
+from codemodder.codemods.use_walrus_if import UseWalrusIf
+from integration_tests.base_test import (
+    BaseIntegrationTest,
+    original_and_expected_from_code_path,
+)
+
+
+class TestUseWalrusIf(BaseIntegrationTest):
+    codemod = UseWalrusIf
+    code_path = "tests/samples/use_walrus_if.py"
+    original_code, _ = original_and_expected_from_code_path(code_path, [])
+    expected_new_code = """
+if (x := foo()) is not None:
+    print(x)
+
+if y := bar():
+    print(y)
+
+z = baz()
+print(z)
+
+
+def whatever():
+    if (b := biz()) == 10:
+        print(b)
+""".lstrip()
+
+    expected_diff = "--- \n+++ \n@@ -1,9 +1,7 @@\n-x = foo()\n-if x is not None:\n+if (x := foo()) is not None:\n     print(x)\n \n-y = bar()\n-if y:\n+if y := bar():\n     print(y)\n \n z = baz()\n@@ -11,6 +9,5 @@\n \n \n def whatever():\n-    b = biz()\n-    if b == 10:\n+    if (b := biz()) == 10:\n         print(b)\n"
+
+    num_changes = 3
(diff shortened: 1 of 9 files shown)


src/codemodder/codemods/use_walrus_if.py
```diff
@@ -0,0 +1,130 @@
...
+    def leave_Assign(self, original_node, updated_node):
+        if self.node_is_selected(original_node):
+            if self._is_valid_modification(original_node):
+                position = self.node_position(original_node)
+                self._modify_next_if.append((position, updated_node))
+                return cst.RemoveFromParent()
+
+        return original_node
+
+    def leave_SimpleStatementLine(self, original_node, updated_node):
```
andrecsilva, line 110: This breaks the expected types from `cst.Module.body`. If you look at the documentation:
https://libcst.readthedocs.io/en/latest/nodes.html#libcst.Module
you can see we should expect that `body` has type `Sequence[Union[SimpleStatementLine, BaseCompoundStatement]]`. But this method may sneak `EmptyLine` and other types into it.
This can break subsequent transformations (not much of a problem in our current use case, but a problem nonetheless).
If you want to preseve those, the ideal outcome is that those should be inherited by the next statement.

andrecsilva (changes requested):
Lacking integration tests.

integration_tests/test_use_walrus_if.py
```diff
@@ -0,0 +1,32 @@
+from codemodder.codemods.use_walrus_if import UseWalrusIf
+from integration_tests.base_test import (
+    BaseIntegrationTest,
+    original_and_expected_from_code_path,
+)
+
+
+class TestUseWalrusIf(BaseIntegrationTest):
```
drdavella, line 8: @andrecsilva the integration test is here. Is there something specific that is missing?

src/codemodder/codemods/use_walrus_if.py
drdavella: This is interesting and I take your point, but I think that this would be very difficult to implement in terms of the visitor pattern since it would require quite a bit of state to be passed around. I can experiment with it a little longer but since this is already working with the behavior we want I might just have to revisit it later.

integration_tests/test_use_walrus_if.py
andrecsilva: Nah, I've somehow missed it. Sorry

andrecsilva (approved):

src/codemodder/codemods/use_walrus_if.py
drdavella: I added a comment to the code to preserve this info.

PR #43, opened 2023-09-22
Added support for noqa comments in RemoveUnusedImports

andrecsilva:
## Overview
`RemoveUnusedImports` will now respect `#noqa` comments

### DIFF
--- a/src/codemodder/codemods/remove_unused_imports.py
+++ b/src/codemodder/codemods/remove_unused_imports.py
@@ -1,5 +1,11 @@
+from libcst import ensure_type, matchers
 from libcst.codemod.visitors import GatherUnusedImportsVisitor
-from libcst.metadata import PositionProvider, QualifiedNameProvider, ScopeProvider
+from libcst.metadata import (
+    PositionProvider,
+    QualifiedNameProvider,
+    ScopeProvider,
+    ParentNodeProvider,
+)
 from codemodder.codemods.base_codemod import (
     BaseCodemod,
     CodemodMetadata,
@@ -11,6 +17,7 @@ from codemodder.codemods.transformations.remove_unused_imports import (
 )
 import libcst as cst
 from libcst.codemod import Codemod, CodemodContext
+import re
 
 
 class RemoveUnusedImports(BaseCodemod, Codemod):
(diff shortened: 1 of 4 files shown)


src/codemodder/codemods/remove_unused_imports.py
```diff
@@ -53,6 +66,38 @@ def filter_by_path_includes_or_excludes(self, pos_to_match):
...
         return True
 
+    def _has_noqa_comment(self, node):
+        """
+        Check if the import has a #noqa comment attached to it
+        """
+        parent = self.get_metadata(ParentNodeProvider, node)
+        if parent and matchers.matches(parent, matchers.SimpleStatementLine()):
+            stmt = ensure_type(parent, cst.SimpleStatementLine)
+            pattern = re.compile(r"^#\s*noqa")
```
drdavella: Just a very small thing but this can be compiled once at the top of the module. It doesn't have a big performance impact in this case though.

src/codemodder/codemods/remove_unused_imports.py
```diff
@@ -53,6 +66,38 @@ def filter_by_path_includes_or_excludes(self, pos_to_match):
...
+            # has a trailing comment string
+            trailing_comment_string = (
+                stmt.trailing_whitespace.comment.value
+                if stmt.trailing_whitespace.comment
+                else None
+            )
+            if trailing_comment_string and pattern.match(trailing_comment_string):
+                return True
+
+            # has a comment right above it
```
drdavella, line 93: I'm a little unclear on whether this case is typically supported by linters, etc. Did you find some documentation on this?

drdavella (changes requested):
This looks great. The only thing is that I think we should also honor `# pylint: disable=unused-import`, which shouldn't be hard to add.

src/codemodder/codemods/remove_unused_imports.py
andrecsilva: Not really, I've just assume it would be the case. I know pylint will search for comment before it, not sure about other tools.

clavedeluna: can confirm pylint does handle "if pylint disable comment is right above code" case

drdavella (approved):
Looks nice 👍

PR #47, opened 2023-09-25
Implement codemod registry

drdavella:
## Overview
*Move codemods to `core_codemods` package; implement codemod registry*

## Description

* This PR moves the codemods to the `core_codemods` package and implements a codemod registry
* There are a few major goals with this change:
  * Separate individual codemod implementation from the codemodder package itself
  * Move in the direction of "external" validation for codemods rather than using `__init_subclass__` hooks
  * Enable future third-party codemod implementation/registration
* The codemod registry is responsible for detecting and collecting all registered codemods
  * It also performs codemod validation and filtering based on ID
* Codemod docs and semgrep configurations are now more tightly coupled with the codemods themselves
* Codemods are registered with a namespace prefix that no longer needs to be hardcoded in codemodder
* I was able to remove the `__init_subclass__` validation for the "base" API but some additional work will be required before we can get rid of this in the "sugar" API
  * I still think that maybe these shouldn't be class-level attributes; hopefully this moves us a bit closer to that goal
* This is a relatively big change but I also view it as just one step of incremental refactoring
  * I don't think the result is perfect yet but hopefully it moves in a good direction
  * I hope that eventually this will better enable us to clean up the codemod class hierarchy a bit


src/codemodder/registry.py
```diff
@@ -0,0 +1,175 @@
...
+class CodemodCollection:
+    """A collection of codemods that all share the same origin and documentation."""
+
+    origin: str
+    docs_module: str
+    semgrep_config_module: str
+    codemods: list
+
+
+class _CodemodContainer:
```
drdavella: This is a slightly awkward construct but I might suggest that in the longer time this actually outlines the API we want for the `BaseCodemod` class. Eventually maybe this becomes an ABC that all codemods inherit.

src/codemodder/__main__.py
clavedeluna, line 1: I know this PR is in the right direction bc of all the red I see in this file <3

src/codemodder/registry.py
clavedeluna: Both CodemodCollection and CodemodContainer sound to me like containers of some sort, such as a list. What about making one of these inherit from list? Maybe that's a later thing, just a thought.

Also, how about renaming `_CodemodContainer` to `CodemodMetadata` or `CodemodWrapper`? Also I think this class in particular could be a dataclass. What clues me in to that is all the property methods, but some will have to be manual, of course.

clavedeluna: More thoughts, since you have the idea of this class `_CodemodContainer` being the "real" BaseCodemod, what do you think is the distinction between the data a BaseCodemod class should store versus this new `_CodemodContainer`? I still feel they should be one class but I'm listening!

clavedeluna (commented):
I'm liking this a ton., Just left some thoughts around naming to clarify concepts.

andrecsilva (approved):

src/codemodder/registry.py
drdavella: * I did think about making `CodemodCollection` a `list` but I'm not sure whether we really need that API. It might be worth a quick experiment though
* I do think maybe `_CodemodWrapper` is a better name and it would be a good dataclass
* The main difference between `_CodemodContainer/Wrapper` and `BaseCodemod` is that the container knows about the registry and associated metadata and the base codemod does not. But maybe there's a good way to consolidate those  two things somehow. I think that's a future effort though.

PR #48, opened 2023-09-26
Codemod: fix mutable params

drdavella:
## Overview
*Add codemod to fix mutable default params in function/method arguments*

## Description

* This fixes cases where a mutable default argument was used in a function definition
* This is generally considered bad practice and potentially unsafe, so it's useful to have a codemod that fixes it
* It seemed easiest to implement this without using semgrep since it can be done efficiently with LibCST matchers

### DIFF
--- a/src/codemodder/codemods/fix_mutable_params.py
+++ b/src/codemodder/codemods/fix_mutable_params.py
@@ -1,0 +1,112 @@
+import libcst as cst
+from libcst import matchers as m
+
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import BaseCodemod
+
+
+class FixMutableParams(BaseCodemod):
+    NAME = "fix-mutable-params"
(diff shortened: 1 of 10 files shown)


andrecsilva (approved):

src/codemodder/codemods/fix_mutable_params.py
```diff
@@ -0,0 +1,142 @@
...
+
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import BaseCodemod
+
+
+class FixMutableParams(BaseCodemod):
+    NAME = "fix-mutable-params"
+    SUMMARY = "Replace mutable parameters with None"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW
+    DESCRIPTION = "Replace mutable parameters with None"
```
clavedeluna, line 12: could've made this DESCRIPTION = SUMMARY

PR #50, opened 2023-09-27
Jinja autoescape

clavedeluna:
## Overview
*Add a codemod that checks if jinja2.Environment enables autoescape*

## Description

* `autoescape`, a jinja2 setting to guard against XSS attacks, is disabled by default (and can explicitly be disabled)
* This PR ensures users enable autoescape
* Due to confusion around jinja2's `select_autoescape` function (see bug report [here](https://github.com/pallets/jinja/issues/1890)), this codemod does not detect if autoescape is assigned to a callable, such as `Environment(autoescape=select_autoescape())`. this is the recommended way in the docs. We will cover that later.

### DIFF
--- a/src/codemodder/codemods/api/helpers.py
+++ b/src/codemodder/codemods/api/helpers.py
@@ -64,10 +64,8 @@ class Helpers:
 
         for arg in original_node.args:
             if matchers.matches(arg.keyword, matchers.Name(target_arg_name)):
-                new = cst.Arg(
-                    keyword=cst.parse_expression(target_arg_name),
-                    value=cst.parse_expression(target_arg_replacement_val),
-                    equal=arg.equal,
+                new = self.make_new_arg(
+                    target_arg_name, target_arg_replacement_val, arg
                 )
                 arg_added = True
             else:
@@ -75,13 +73,21 @@ class Helpers:
             new_args.append(new)
 
         if add_if_missing and not arg_added:
-            new = cst.Arg(
-                keyword=cst.parse_expression(target_arg_name),
-                value=cst.parse_expression(target_arg_replacement_val),
-                equal=cst.AssignEqual(
-                    whitespace_before=cst.SimpleWhitespace(""),
-                    whitespace_after=cst.SimpleWhitespace(""),
-                ),
-            )
+            new = self.make_new_arg(target_arg_name, target_arg_replacement_val)
             new_args.append(new)
         return new_args
+
+    def make_new_arg(self, name, value, existing_arg=None):
+        equal = (
+            existing_arg.equal
+            if existing_arg
+            else cst.AssignEqual(
+                whitespace_before=cst.SimpleWhitespace(""),
+                whitespace_after=cst.SimpleWhitespace(""),
+            )
+        )
+        return cst.Arg(
+            keyword=cst.parse_expression(name),
+            value=cst.parse_expression(value),
+            equal=equal,
+        )
--- a/src/codemodder/codemods/api/helpers.py
+++ b/src/codemodder/codemods/api/helpers.py
@@ -47,10 +47,21 @@ class Helpers:
     def parse_expression(self, expression: str):
         return cst.parse_expression(expression)
 
-    def replace_arg(self, original_node, target_arg_name, target_arg_replacement_val):
-        """Given a node, return its args with one arg's value changed"""
+    def replace_arg(
+        self,
+        original_node,
+        target_arg_name,
+        target_arg_replacement_val,
+        add_if_missing=False,
+    ):
+        """Given a node, return its args with one arg's value changed.
+
+        If add_if_missing is True, then if target arg is not present, add it.
+        """
         assert hasattr(original_node, "args")
         new_args = []
+        arg_added = False
+
         for arg in original_node.args:
             if matchers.matches(arg.keyword, matchers.Name(target_arg_name)):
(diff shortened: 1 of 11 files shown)


src/core_codemods/docs/pixee_python_enable-jinja2-autoescape.md
```diff
@@ -0,0 +1,17 @@
...
+default behavior is to not autoescape when rendering templates, which makes your applications
+vulnerable to Cross-Site Scripting (XSS) attacks.
+
+Our codemod currently checks if you forgot to turn autoescape on or if you explicitly disabled it. The change looks as follows:
+
+```diff
+  from jinja2 import Environment
+
+- env = Environment()
+- env = Environment(autoescape=False, loader=some-loader)
```
drdavella: Minor thing but the `-` here and below is not valid Python syntax.

src/core_codemods/enable_jinja2_autoescape.py
```diff
@@ -0,0 +1,34 @@
...
+            rules:
+                - patterns:
+                  - pattern-either:
+                    - pattern: |
+                        jinja2.Environment()
+                    - pattern: |
+                        jinja2.Environment(...)
+                    - pattern: |
+                        jinja2.Environment(..., autoescape=False, ...)
+                  - pattern-not: |
```
drdavella: Wouldn't this `pattern-not` by itself be sufficient instead of enumerating the other cases above?

src/codemodder/codemods/api/helpers.py
```diff
@@ -47,18 +47,41 @@ def update_assign_rhs(self, updated_node: cst.Assign, rhs: str):
...
                     value=cst.parse_expression(target_arg_replacement_val),
                     equal=arg.equal,
                 )
+                arg_added = True
             else:
                 new = arg
             new_args.append(new)
+
+        if add_if_missing and not arg_added:
+            new = cst.Arg(
```
drdavella: Just a small thing but it seems like this `Arg` creation could be factored out into another method.

src/core_codemods/enable_jinja2_autoescape.py
```diff
@@ -0,0 +1,34 @@
...
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW
+    SUMMARY = "Enable jinja2 autoescape"
+    DESCRIPTION = "Makes the `autoescape` parameter to jinja2.Environment be `True`."
+
+    @classmethod
+    def rule(cls):
+        return """
+            rules:
+                - patterns:
+                  - pattern-either:
```
drdavella: See comment below but I think this might be redundant with the `pattern-not`.

src/core_codemods/docs/pixee_python_enable-jinja2-autoescape.md
```diff
@@ -0,0 +1,17 @@
+This codemod ensures you configure jinja2 to turn on autoescaping of HTML content. Unfortunately, the jinja2
```
drdavella:
```suggestion
This codemod enables autoescaping of HTML content in `jinja2`. Unfortunately, the jinja2
```

src/core_codemods/docs/pixee_python_enable-jinja2-autoescape.md
```diff
@@ -0,0 +1,17 @@
...
+default behavior is to not autoescape when rendering templates, which makes your applications
+vulnerable to Cross-Site Scripting (XSS) attacks.
+
+Our codemod currently checks if you forgot to turn autoescape on or if you explicitly disabled it. The change looks as follows:
```
drdavella:
```suggestion
Our codemod checks if you forgot to enable autoescape or if you explicitly disabled it. The change looks as follows:
```

src/core_codemods/docs/pixee_python_enable-jinja2-autoescape.md
```diff
@@ -0,0 +1,17 @@
...
+  from jinja2 import Environment
+
+- env = Environment()
+- env = Environment(autoescape=False, loader=some-loader)
++ env = Environment(autoescape=True)
++ env = Environment(autoescape=True, loader=some-loader)
+  ...
+```
+
+At this time, this codemod will not detect if `autoescape` is assigned to a callable.
```
drdavella: I understand the caveat but I think we can omit it for now.
```suggestion
```

drdavella (changes requested):
Looks good overall. The main substantive comment is about the semgrep pattern, plus a few minor comments on the docs.

src/codemodder/codemods/api/helpers.py
clavedeluna: ah yes, thought about that and forgot, will do!

src/core_codemods/docs/pixee_python_enable-jinja2-autoescape.md
clavedeluna: I'll update it but I thought (idk why) we were moving toward more active and not passive voice.

src/core_codemods/enable_jinja2_autoescape.py
clavedeluna: oh hmm good point let me check if tests pass with that

drdavella (approved):

PR #51, opened 2023-09-28
Update tests to use `.name()` and rely on wrapper id

clavedeluna:
## Overview
I noticed we forgot to update tests when we changed the codemod API. I think this is the way to go to encourage us to use `.name()` instead of accessing the metadata attr directly.

Secondly, now that we have a codemod wrapper, and in fact the wrapper is what we use to report id in codetf results, then I was able to remove the base_codemod id class attr. It just needed an update in integration tests.

### DIFF
--- a/integration_tests/base_test.py
+++ b/integration_tests/base_test.py
@@ -1,4 +1,4 @@
-# pylint: disable=no-member
+# pylint: disable=no-member, protected-access, attribute-defined-outside-init
 import git
 import json
 import os
@@ -6,6 +6,7 @@ import pathlib
 import subprocess
 
 from codemodder import __VERSION__
+from codemodder import registry
 
 
 SAMPLES_DIR = "tests/samples"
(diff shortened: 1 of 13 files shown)


integration_tests/base_test.py
```diff
@@ -1,11 +1,12 @@
-# pylint: disable=no-member
+# pylint: disable=no-member, protected-access, attribute-defined-outside-init
```
drdavella, line 1: ❤️ but also maybe still a code smell that indicates we can clean up the class hierarchy a bit.

drdavella (approved):

integration_tests/base_test.py
clavedeluna: this is actually because this integration test code is:
1. calling on `_codemods`. I didn't want to write a class method just for tests calling it
2. bc pylint isn't good at understanding that defining something inside `setup_method` in pytest is kinda the equivalent of defining it in an init method.

PR #52, opened 2023-09-28
Fix typo in tests

drdavella:
## Overview
*Fixes one tiny typo that has propagated itself via copy/paste throughout our tests 😆*


clavedeluna (approved):
this is 10000% a Dani problem!

PR #54, opened 2023-09-29
First steps towards threading.Lock() codemod

drdavella:
## Overview
*Unblock progress on `threading.Lock()` codemod*

## Description

* I picked up where @clavedeluna left off to see if we could make some progress on this codemod
  * The semgrep problem has not been fixed but it appears that we can make progress anyway
* It's not entirely complete, we still need integration tests and a detailed description
* We also might want some more thorough unit tests

### DIFF
--- a/src/core_codemods/with_threading_lock.py
+++ b/src/core_codemods/with_threading_lock.py
@@ -1,0 +1,57 @@
+import libcst as cst
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import SemgrepCodemod
+
+
+class WithThreadingLock(SemgrepCodemod):
+    NAME = "bad-lock-with-statement"
+    SUMMARY = "Replace deprecated usage of threading.Lock context manager"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW
+    DESCRIPTION = "Separates threading lock instantiation and call with `with` statement into two steps."
+
+    @classmethod
+    def rule(cls):
+        return """
+        rules:
+          - patterns:
+            - pattern: |
+                with $BODY:
+                    ...
+            - metavariable-pattern:
+                metavariable: $BODY
+                patterns:
+                - pattern: threading.Lock()
+            - pattern-inside: |
+                import threading
+                ...
+            - focus-metavariable: $BODY
+        """
+
+    def leave_With(self, original_node: cst.With, updated_node: cst.With):
+        # We deliberately restrict ourselves to simple cases where there's only one with clause for now.
+        # Semgrep appears to be insufficiently expressive to match multiple clauses correctly.
+        # We should probably just rewrite this codemod using libcst without semgrep.
+        if len(original_node.items) == 1 and self.node_is_selected(
+            original_node.items[0]
+        ):
+            # TODO: how to avoid name conflicts here?
+            name = cst.Name(value="lock")
+            assign = cst.SimpleStatementLine(
+                body=[
+                    cst.Assign(
+                        targets=[cst.AssignTarget(target=name)],
+                        value=updated_node.items[0].item,
+                    )
+                ]
+            )
+            # TODO: add result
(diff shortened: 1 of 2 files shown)


tests/codemods/test_with_threading_lock.py
```diff
@@ -0,0 +1,63 @@
...
+from tests.codemods.base_codemod_test import BaseSemgrepCodemodTest
+
+
+class TestWithThreadingLock(BaseSemgrepCodemodTest):
+    codemod = WithThreadingLock
+
+    def test_rule_ids(self):
+        assert self.codemod.name() == "bad-lock-with-statement"
+
+    def test_simple_replacement(self, tmpdir):
```
clavedeluna: nitpick but if you could rename these tests similar to the ones we already name, `test_import`, `test_from_import`, etc, just for consistency. that's also a way to see if we miss some test case that another codemod tested

clavedeluna (changes requested):
You got it working! the semgrep rule seems to be the right fix here as it doesn't look like you're hitting the issues I did.

With integration tests and docs this should be good to go.

src/core_codemods/with_threading_lock.py
```diff
@@ -0,0 +1,51 @@
...
+                - pattern: threading.Lock()
+            - pattern-inside: |
+                import threading
+                ...
+            - focus-metavariable: $BODY
+        """
+
+    def leave_With(self, original_node: cst.With, updated_node: cst.With):
+        if original_node.items and self.node_is_selected(original_node.items[0]):
+            # TODO: how to avoid name conflicts here?
```
andrecsilva, line 37: You can use `ScopeProvider` to gather all the name assignments within the current, ancestors and descendants scopes. You want to make sure you don't shadow anything before (ancestors and current scopes) and are not shadowed by anything after (current and descendant scopes). 

It's easier than it seems but if you don't want to deal with it, just open a ticket and I'll look into it later.

src/core_codemods/with_threading_lock.py
```diff
@@ -0,0 +1,51 @@
...
+                patterns:
+                - pattern: threading.Lock()
+            - pattern-inside: |
+                import threading
+                ...
+            - focus-metavariable: $BODY
+        """
+
+    def leave_With(self, original_node: cst.With, updated_node: cst.With):
+        if original_node.items and self.node_is_selected(original_node.items[0]):
```
andrecsilva: You can have several items within a `with` (i.e `with a as b, c as d, ...`, are we purposefully avoiding those cases?

src/core_codemods/with_threading_lock.py
drdavella: That's great, I think I'll make a ticket for it. Even if we can avoid collisions we still need to decide on how to generate a good name.

src/core_codemods/with_threading_lock.py
drdavella: Yes I think we can avoid those for now. It seems like it will be a bit harder to match those using semgrep, but I think the current implementation probably covers 90% of cases anyway.

tests/codemods/test_with_threading_lock.py
```diff
@@ -0,0 +1,63 @@
...
+"""
+        expected = """import threading
+lock = threading.Lock()
+with lock:
+    ...
+
+with threading_lock():
+    ...
+"""
+        self.run_and_assert(tmpdir, input_code, expected)
```
clavedeluna, line 63: would be worth adding a `with a, b` case that Andre mentioned above to show no effect as well

clavedeluna (approved):

clavedeluna:
I didn't realize you intended to merge without an integration test or docs. I thought something prevented merging a new codemod without the corresponding docs file?

drdavella:
@clavedeluna I created a new ticket for integration tests and docs. I mostly wanted to get this over the hump so that someone else could possibly pick up the rest. The test you're referring to only works if the codemod is actually registered, which this one is not (yet).

PR #56, opened 2023-10-02
Updated dependency on libcst

andrecsilva:
## Overview
Updated libcst dependency to include some recent fixes.

## Description
- Updated libcst dependency to include some recent fixes. Right now it points to a specific commit. Should be updated again with a new libcst release.
- Added tests to ensure imports are being added at the right place. These are a bit redundant especially since this is tested within libcst. May be removed in the future.
- Removed a workaround in NameResolutionMixin

## Additional Details
The updated tests were failing locally, not sure why.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,8 @@ license = {file = "LICENSE"}
 dependencies = [
     "dependency-manager @ git+https://github.com/pixee/python-dependency-manager#egg=dependency-manager",
     "isort~=5.12.0",
-    "libcst~=1.0.0",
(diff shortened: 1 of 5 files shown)


drdavella (approved):
This is great, thanks for all this work. I can create a ticket to update the libcst dependency when it's released but do you have any sense of the timeline?

andrecsilva:
> This is great, thanks for all this work. I can create a ticket to update the libcst dependency when it's released but do you have any sense of the timeline?

No clue, but their last minor release was within a couple weeks from the last one. Let's just hope we have something within a month

PR #61, opened 2023-10-03
Codemod Lxml parser defaults

clavedeluna:
## Overview
*New codemod for lxml classes that have security-related parameters*

## Description

* I generalized `replace_arg` to `replace_args` so now we can replace a list of args. The method api accepts `args_info`, which we expect to be a list of the namedtuple `NewArg`.

### DIFF
--- a/src/codemodder/codemods/api/helpers.py
+++ b/src/codemodder/codemods/api/helpers.py
@@ -47,11 +47,18 @@ class Helpers:
     def parse_expression(self, expression: str):
         return cst.parse_expression(expression)
 
-    def replace_args(
-        self,
-        original_node,
-        args_info,
-    ):
+    def replace_args(self, original_node, args_info):
+        """
+        Iterate over the args in original_node and replace each arg
+        with any matching arg in `args_info`.
+
+        :param original_node: libcst node with args attribute.
+        :param list args_info: List of tuples.
+            Ex: [("new_arg", "new_arg_value", True/False)]
+            The True/False value should indicate whether the arg should be added
+            if not present in original_node's args.
+        """
+        assert hasattr(original_node, "args")
         new_args = []
 
         for arg in original_node.args:
@@ -68,36 +75,6 @@ class Helpers:
                 new = self.make_new_arg(arg_name, replacement_val)
                 new_args.append(new)
 
-        return new_args
-
-    def replace_arg(
-        self,
-        original_node,
-        target_arg_name,
-        target_arg_replacement_val,
-        add_if_missing=False,
-    ):
-        """Given a node, return its args with one arg's value changed.
-
-        If add_if_missing is True, then if target arg is not present, add it.
-        """
-        assert hasattr(original_node, "args")
-        new_args = []
-        arg_added = False
-
-        for arg in original_node.args:
-            if matchers.matches(arg.keyword, matchers.Name(target_arg_name)):
-                new = self.make_new_arg(
-                    target_arg_name, target_arg_replacement_val, arg
-                )
-                arg_added = True
-            else:
-                new = arg
-            new_args.append(new)
-
-        if add_if_missing and not arg_added:
-            new = self.make_new_arg(target_arg_name, target_arg_replacement_val)
-            new_args.append(new)
         return new_args
 
     def make_new_arg(self, name, value, existing_arg=None):
--- a/src/codemodder/codemods/api/helpers.py
+++ b/src/codemodder/codemods/api/helpers.py
@@ -47,6 +47,29 @@ class Helpers:
     def parse_expression(self, expression: str):
         return cst.parse_expression(expression)
 
+    def replace_args(
+        self,
+        original_node,
+        args_info,
+    ):
+        new_args = []
+
+        for arg in original_node.args:
+            arg_name, replacement_val, idx = _match_with_existing_arg(arg, args_info)
+            if arg_name is not None:
+                new = self.make_new_arg(arg_name, replacement_val, arg)
+                del args_info[idx]
+            else:
+                new = arg
+            new_args.append(new)
+
+        for arg_name, replacement_val, add_if_missing in args_info:
+            if add_if_missing:
+                new = self.make_new_arg(arg_name, replacement_val)
+                new_args.append(new)
+
+        return new_args
+
     def replace_arg(
         self,
         original_node,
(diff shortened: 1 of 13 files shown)


src/codemodder/registry.py
```diff
@@ -152,7 +152,7 @@ def match_codemods(
 def load_registered_codemods() -> CodemodRegistry:
     registry = CodemodRegistry()
     logger.info("Loading registered codemod collections")
-    for entry_point in entry_points()["codemods"]:
+    for entry_point in entry_points().select(group="codemods"):
```
clavedeluna: Fixed a DeprecationWarning. See https://importlib-metadata.readthedocs.io/en/latest/using.html#entry-points

andrecsilva (approved):

src/codemodder/registry.py
drdavella: Not sure whether you ran into this problem with the change but I think the right way to fix this is probably by adding some conditional logic based on Python version.

src/codemodder/codemods/api/helpers.py
```diff
@@ -47,34 +47,34 @@ def update_assign_rhs(self, updated_node: cst.Assign, rhs: str):
...
-    ):
-        """Given a node, return its args with one arg's value changed.
+    def replace_args(self, original_node, args_info):
+        """
+        Iterate over the args in original_node and replace each arg
+        with any matching arg in `args_info`.
 
-        If add_if_missing is True, then if target arg is not present, add it.
+        :param original_node: libcst node with args attribute.
+        :param list args_info: List of tuples.
```
drdavella: Probably overengineered but this might be a nice place for a namedtuple?

src/core_codemods/lxml_safe_parser_defaults.py
```diff
@@ -0,0 +1,42 @@
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import SemgrepCodemod
+
+
+class LxmlSafeParserDefaults(SemgrepCodemod):
+    NAME = "safe-lxml-parser-defaults"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW
+    SUMMARY = "Enable all security checks in `lxml.etree.XMLParser` call."
+    DESCRIPTION = "...........TODO"
```
drdavella: Right now this actually gets used as the `CHANGE_DESCRIPTION` for the simplified API (we should fix that). But we do need to make sure it's present.

src/core_codemods/lxml_safe_parser_defaults.py
```diff
@@ -0,0 +1,42 @@
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import SemgrepCodemod
+
+
+class LxmlSafeParserDefaults(SemgrepCodemod):
+    NAME = "safe-lxml-parser-defaults"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW
```
drdavella: This one feels like it should be `MERGE_WITHOUT_REVIEW` but I'm not sure.

tests/codemods/test_lxml_safe_parameter_defaults.py
```diff
@@ -0,0 +1,100 @@
...
+        expexted_output = f"""import lxml
+
+parser = lxml.etree.{klass}(resolve_entities=False)
+var = "hello"
+"""
+
+        self.run_and_assert(tmpdir, input_code, expexted_output)
+
+    @each_class
+    def test_from_import(self, tmpdir, klass):
```
drdavella, line 32: Do we need another test case for this:
```python
from lxml import etree
...
parser = etree.XMLParser(...)
```

src/codemodder/registry.py
clavedeluna: I ended up removing this change (sorry forgot to remove my comment) bc yeah py versions. We can leave as is for now.

drdavella (changes requested):
Looks good. Primary request is about `DESCRIPTION` but it also might be nice to consider something like namedtuple for the new parameter arguments.

src/codemodder/codemods/api/helpers.py
clavedeluna: not an oversight (see PR description). My gut feeling with named tuple is that then any codemod that uses it needs to import it. but then again they still need to know the API and to do it right.... so maybe named tuple is the right way

src/core_codemods/lxml_safe_parser_defaults.py
clavedeluna: woops total oversight here that's why the TODO. How verbose can I be, should I be?  Is this used as the PR description when pixeebot makes the PR?

src/core_codemods/lxml_safe_parser_defaults.py
clavedeluna: I think I agree

src/codemodder/codemods/api/helpers.py
drdavella: Oh yeah sorry I missed that comment. 
>My gut feeling with named tuple is that then any codemod that uses it needs to import it

This feels like it might be a feature to me. But I don't feel _super_ strongly about it at the moment.

src/core_codemods/lxml_safe_parser_defaults.py
drdavella: Right now (and again we need to fix this) `DESCRIPTION` -> `CHANGE_DESCRIPTION` for the simple API. `CHANGE_DESCRIPTION` should be a very simple description of the individual change itself. So in this case probably something like "Replace lxml parser parameters with safe defaults". In contrast the `SUMMARY` becomes the title of the PR that gets opened so it should be very brief and descriptive.

src/core_codemods/lxml_safe_parser_defaults.py
```diff
@@ -0,0 +1,42 @@
...
+    SUMMARY = "Enable all security checks in `lxml.etree.XMLParser` call."
```
drdavella: Sorry should have mentioned this before, but this becomes the PR title. I would suggest this could be even briefer, like "Use safe defaults for lxml parsers" (and without a trailing period).

src/codemodder/codemods/api/helpers.py
clavedeluna: Added named tuple

src/core_codemods/lxml_safe_parser_defaults.py
```diff
@@ -0,0 +1,43 @@
...
+                  - pattern-inside: |
+                      import lxml
+                      ...
+        """
+
+    def on_result_found(self, original_node, updated_node):
+        new_args = self.replace_args(
+            original_node,
+            [
+                NewArg(name="resolve_entities", value="False", add_if_missing=True),
```
drdavella, line 38: I like this, I think it makes sense.

drdavella (approved):

PR #63, opened 2023-10-04
Support composable codemods

drdavella:
## Overview
*Correctly support multiple codemod composition*

## Description

* We need to correctly support the application of multiple codemods
* We now invoke semgrep once per codemod and apply the results of each codemod to the source tree
* This means the updated source tree state is now used as input to each subsequent semgrep/codemod invocation
* This comes with a performance penalty but it's the only way to handle multiple codemods correctly
* It is now important to respect the order in which codemods are given using `--codemod-include`
* This solution does not account for the possibility of executing multiple codemods with externally generated results (e.g. from CodeQL or other rule providers). However we don't currently have any such codemods so the impact is nil.
* This PR also involves some refactoring. I still view the `CodemodExecutorWrapper` as an intermediate workaround until we revisit the codemod class hierarchy

## Future Work
* Clean up the class hierarchy to support a cleaner codemod execution API
* The `--dry-run` option should operate on a copy of the entire source tree since right now it doesn't work correctly with multiple codemods (not currently a problem for the platform)


integration_tests/test_url_sandbox.py
```diff
@@ -19,7 +19,7 @@ class TestUrlSandbox(BaseIntegrationTest):
     expected_diff = '--- \n+++ \n@@ -1,4 +1,4 @@\n-import requests\n+from security import safe_requests\n \n-requests.get("www.google.com")\n+safe_requests.get("www.google.com")\n var = "hello"\n'
     expected_line_change = "3"
     change_description = UrlSandbox.CHANGE_DESCRIPTION
-    num_changed_files = 2
+    num_changed_files = 1
```
drdavella, line 22: This was happening because we previously didn't restrict the `--path-include` for these tests.

src/codemodder/cli.py
```diff
@@ -44,7 +44,7 @@ class CsvListAction(argparse.Action):
     """
 
     def __call__(self, parser, namespace, values, option_string=None):
-        items = set(values.split(","))
+        items = values.split(",")
```
drdavella: Order matters now. But we might want to validate that the IDs are unique some other way.

integration_tests/test_url_sandbox.py
clavedeluna: I think the intention may have been to test without using that flag but hey it's fine by me

integration_tests/test_multiple_codemods.py
```diff
@@ -0,0 +1,60 @@
...
+            "--output",
+            str(codetf_path),
+            f"--codemod-include={codemods}",
+            f"--path-include={source_file_name}",
+        ]
+
+        completed_process = subprocess.run(
+            command,
+            check=False,
+        )
```
clavedeluna, line 36: would be nice if you could re-use some of the functionality from base integration test since it's almost the same. Also I think it's worth calling `self.check_code_before()` for sanity checking.

src/codemodder/cli.py
clavedeluna: We could rely on dict since we now know starting in py 3.7+ it preserves order.
```
lst = [1, 2, 2, 3, 4, 4, 5]
unique_lst = list(dict.fromkeys(lst).keys())
```

src/codemodder/codemods/base_codemod.py
```diff
@@ -35,6 +37,16 @@ def __init__(self, execution_context: CodemodExecutionContext, file_context):
         self.execution_context = execution_context
         self.file_context = file_context
 
+    @classmethod
+    def apply_rule(cls, context, *args, **kwargs):
+        """
+        Apply rule associated with this codemod and gather results
+
+        Does nothing by default. Subclasses may override for custom rule logic.
```
clavedeluna, line 45: maybe update the docstring since idk why the `del` line happens

src/codemodder/executor.py
```diff
@@ -0,0 +1,88 @@
...
+        self.docs_module = docs_module
+        self.semgrep_config_module = semgrep_config_module
+
+    def apply(self, context):
+        """
+        Wraps the codemod's apply method to inject additional arguments.
+
+        Not all codemods will need these arguments.
+        """
+        return self.apply_rule(context, yaml_files=self.yaml_files)
```
clavedeluna, line 31: should this not be `codemod.apply_rule`? I don't see an `apply_rule` definition in the executor class

src/codemodder/semgrep.py
```diff
@@ -31,29 +32,12 @@ def run_on_directory(yaml_files: List[Path], directory: Path):
         command.append(str(directory))
         joined_command = " ".join(command)
         logger.debug("Executing semgrep with: `%s`", joined_command)
+        # TODO: don't use the joined command
```
clavedeluna: because? I guess this could be a security risk

src/codemodder/codemodder.py
```diff
@@ -197,15 +172,5 @@ def run(original_args) -> int:
 def main():
     # TODO: I'm not sure why this needs to be parsed out separately
     # Maybe it has something to do with the invocation as python -m codemodder.
```
clavedeluna: I think you can entirely remove this comment. IIRC we parsed it first bc we wanted to send the args to another func, too. Now that's not the case

src/codemodder/codemodder.py
```diff
@@ -177,15 +149,18 @@ def run(original_args) -> int:
...
 
-    analyze_files(
-        context,
-        files_to_analyze,
-        codemods_to_run,
-        sarif_results,
-        argv,
-    )
+    # run codemods in sequence
+    for codemod in codemods_to_run:
```
clavedeluna, line 154: Could you point to where the libcst tree output of codemod 1 is passed to codemod 2? I expected to see that change in `analyze_files` since that's where we gather the code tree but maybe I"m missing a key thing

clavedeluna (changes requested):

andrecsilva (changes requested):
The whole flow of execution for multiple codemods feels wrong.

There are three straightforward strategies to run multiple codemods:
1. The output tree of a codemod is the input of the next;
2. The output code of a codemod is parsed into a new tree which is the input of the next codemod;
3. We rewrite the file with the output code which is then read and parsed into a tree for the next codemod.

These are ordered by speed. We adopt the third one (which is the reason `--dry-run` won't work). I see no reason why we just don't go with the second. The first one is the best, of course, but it relies on each codemod outputting "valid" trees (there is at least one codemod that I know does not).

clavedeluna:
> 1. The output tree of a codemod is the input of the next;

this is what I assumed we wanted to happen which is why I asked for clarification

src/codemodder/codemods/base_codemod.py
drdavella: The `del` is just considered good practice for type checkers since it indicates that the variables are unused.

src/codemodder/executor.py
drdavella: This is because I've implemented it as a proxy so it automatically delegates to the wrapped object's methods.

src/codemodder/semgrep.py
drdavella: I'm not sure about security it just seems cleaner since we already have a list and the API accepts a list.

drdavella:
@andrecsilva @clavedeluna what you are describing is a potential optimization that is only possible if certain conditions hold. Remember, semgrep has no access to the parsed tree: it can only read files on disk. So no matter what, the result of any one codemod needs to be written to disk in order to be visible to the next semgrep invocation.

Now, we could still possibly keep the updated tree in memory and feed that to the next codemod, but only if the tree represents the updated metadata (i.e. line number, column) exactly the way that it is represented on disk. This is because we need to map any _new_ semgrep results from the following codemod onto the updated tree. This is probably the case in libcst but I am not entirely sure (maybe @andrecsilva can confirm).

Even with that, there's another problem: storing parsed tree representations of the entire project can be very memory intensive, especially for very large projects. The Java codemodder has already encountered OOM issues with a similar strategy on very large repositories, so I would hesitate to implement any such optimization without a more detailed investigation of the tradeoffs.

While it may not be the most efficient strategy to parse each file multiple times, I would prefer to revisit any possible optimizations in a future PR.

src/codemodder/codemodder.py
drdavella: The parsed tree is not passed to codemod 2. Instead codemod 2 parses the file from disk again. See the comment below for more details.

integration_tests/test_multiple_codemods.py
drdavella: I did look at doing that but the base integration test encodes a fairly strong assumption about one codemod per test. If I inherit from the base integration test it means this test inherits the `test_*` method which won't work for this particular case.

integration_tests/test_url_sandbox.py
drdavella: Yeah you're probably right but now it's necessary in order for the other tests to work.

src/codemodder/cli.py
drdavella: That's a good idea.

andrecsilva:
> @andrecsilva @clavedeluna what you are describing is a potential optimization that is only possible if certain conditions hold. Remember, semgrep has no access to the parsed tree: it can only read files on disk. So no matter what, the result of any one codemod needs to be written to disk in order to be visible to the next semgrep invocation.
> 
> Now, we could still possibly keep the updated tree in memory and feed that to the next codemod, but only if the tree represents the updated metadata (i.e. line number, column) exactly the way that it is represented on disk. This is because we need to map any _new_ semgrep results from the following codemod onto the updated tree. This is probably the case in libcst but I am not entirely sure (maybe @andrecsilva can confirm).
> 
> Even with that, there's another problem: storing parsed tree representations of the entire project can be very memory intensive, especially for very large projects. The Java codemodder has already encountered OOM issues with a similar strategy on very large repositories, so I would hesitate to implement any such optimization without a more detailed investigation of the tradeoffs.
> 
> While it may not be the most efficient strategy to parse each file multiple times, I would prefer to revisit any possible optimizations in a future PR.

For the semgrep issue, you can still run semgrep on a temp file on a by-need basis. That is, if we are running a semgrep codemod, we dump the tree into a temp file and parse results with semgrep. Otherwise, just pass the tree along.

As long as you're running the codemods one file at a time (that is, pick a file, run all codemods for that file, never touch that file again), I don't see why memory would be an issue.

drdavella:
> As long as you're running the codemods one file at a time (that is, pick a file, run all codemods for that file, never touch that file again), I don't see why memory would be an issue.

This would mean we would need to run semgrep once per file rather than once per codemod.

src/codemodder/semgrep.py
drdavella: Actually in general it's going to be safer to pass a list with `shell=False` but the risk here was pretty low to begin with.

andrecsilva:
> > As long as you're running the codemods one file at a time (that is, pick a file, run all codemods for that file, never touch that file again), I don't see why memory would be an issue.
> 
> This would mean we would need to run semgrep once per file rather than once per codemod.

No, we are running semgrep for each codemod. 

The codemod itself will run semgrep, but not for the original file. As I said, for a given tree `T` passed as an input for a semgrep codemod `c`, transform the `T` into code (use `.code` member in libcst), dump into a temp file, run semgrep on that temp file, then pass `c(T)` to the next codemod.

The point I wanted to make with the quoted paragraph is that you don't really need to keep a tree for __every__ file in memory at once, just the a tree for a single file.

drdavella:
@andrecsilva it's possible that I'm misunderstanding something but you said this:
> pick a file, run all codemods for that file, never touch that file again

In order to run all codemods for that file, I would need to run semgrep for each applicable codemod _on each file_. Our current loop looks something like this:
```python
for codemod in requested_codemods:
    codemod.maybe_apply_semgrep()
    for file in requested_files:
        codemod.make_changes()
```

But it seems like you are suggesting something like this:
```python
for file in requested_files:
    # This means I never need to look at this file again and can potentially cache the tree
    for codemod in requested_codemods:
        codemod.maybe_apply_semgrep()
        codemod.make_changes()
```

There are probably some clever ways to get around this contraint. And you are right that we could potentially use tempfiles only for modified files _but only if we run semgrep once per file_.

There is of course an implicit tradeoff here, but it seems obvious (but also possibly wrong) to me that running semgrep for every file is going to be more expensive than the file I/O we perform currently. But the balance may tip in the other direction if we eventually implement more codemods purely in terms of libcst.

andrecsilva:
> @andrecsilva it's possible that I'm misunderstanding something but you said this:
> 
> > pick a file, run all codemods for that file, never touch that file again
> 
> In order to run all codemods for that file, I would need to run semgrep for each applicable codemod _on each file_. Our current loop looks something like this:
> 
> ```python
> for codemod in requested_codemods:
>     codemod.maybe_apply_semgrep()
>     for file in requested_files:
>         codemod.make_changes()
> ```
> 
> But it seems like you are suggesting something like this:
> 
> ```python
> for file in requested_files:
>     # This means I never need to look at this file again and can potentially cache the tree
>     for codemod in requested_codemods:
>         codemod.maybe_apply_semgrep()
>         codemod.make_changes()
> ```

Below is a more detailed take of what I'm suggesting.
```python
for file in requested_files:
    # This means I never need to look at this file again
    tree = cst.parse_module(open(file).read())
    for codemod in requested_codemods:
        results =  run_semgrep(codemod.YAML_FILES,tree) if codemod.is_semgrep else {}
        tree = codemod.make_changes(tree, results)
    # the final tree will be written to file
    open(file).write(tree.code)
```

where `run_semgrep` would look like:

```python
def run_semgrep(yaml_files, tree):
     # could be a common dir for the whole codemodder run
     temp_dir = ...
     temp_file = NamedTemporaryFile(..., dir = temp_dir)
     temp_file.write(tree.code)
     return run_on_directory(yaml_files, temp_dir)
```

This way we only run semgrep / write to disk as many times as there are semgrep codemods. The way it's done right now means we write to disk after every codemod.

> There are probably some clever ways to get around this contraint. And you are right that we could potentially use tempfiles only for modified files _but only if we run semgrep once per file_.

Why? I don't understand why you think can only do that only if we run semgrep once per file.

drdavella:
@andrecsilva yes but that's exactly my point: the semgrep invocation in your example is now within the inner loop which means it is getting invoked `num_files * num_codemods` times in the limit rather than just `num_codemods` times as we do currently.

andrecsilva:
> @andrecsilva yes but that's exactly my point: the semgrep invocation in your example is now within the inner loop which means it is getting invoked `num_files * num_codemods` times in the limit rather than just `num_codemods` times as we do currently.

In the following snippet:

``` python
for codemod in requested_codemods:
    codemod.maybe_apply_semgrep()
    for file in requested_files:
        codemod.make_changes()
```

Semgrep is called once per codemod, true. But you still have to run each codemod rule for each file in the directory (otherwise, how are you getting the results for that file?).

Surely in my proposed solution there are more calls to semgrep, but the number of times a particular rule is ran is the same (once per codemod x file). Yes, there is an extra overhead because in calling semgrep more times, but we have less file writes. Moreover this overhead and the writes are now proportional to the number of semgrep codemods.

I still think it's a net positive, specially since we expect to decrease the number of codemods that rely on semgrep.

drdavella:
> I still think it's a net positive

You may be right but I'm less certain. In either case, this is something that could be answered empirically. I'd like to suggest that this kind of optimization is probably best undertaken as a separate ticket/effort so that we don't block the availability of this feature in the meantime.

clavedeluna:
> ## [Codecov](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee) Report
> > Merging [#63](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee) ([674961e](https://github.com/pixee/codemodder-python/commit/674961ebb52f8db58364bfa280d6f9cc558ea508)) into [main](https://app.codecov.io/gh/pixee/codemodder-python/commit/0790cc22fb4e896861c4955704267910f8fd1379?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee) ([0790cc2](https://github.com/pixee/codemodder-python/commit/0790cc22fb4e896861c4955704267910f8fd1379)) will **decrease** coverage by `0.51%`.
> > The diff coverage is `93.93%`.
> 
> [Impacted file tree graph](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee)
> 
> ```diff
> @@            Coverage Diff             @@
> ##             main      #63      +/-   ##
> ==========================================
> - Coverage   96.32%   95.81%   -0.51%     
> ==========================================
>   Files          44       45       +1     
>   Lines        1660     1674      +14     
> ==========================================
> + Hits         1599     1604       +5     
> - Misses         61       70       +9     
> ```
> 
> [Files](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee)	Coverage Δ	
> [src/codemodder/cli.py](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee#diff-c3JjL2NvZGVtb2RkZXIvY2xpLnB5)	`100.00% <100.00%> (ø)`	
> [src/codemodder/codemodder.py](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee#diff-c3JjL2NvZGVtb2RkZXIvY29kZW1vZGRlci5weQ==)	`96.34% <100.00%> (ø)`	
> [src/codemodder/codemods/base_codemod.py](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee#diff-c3JjL2NvZGVtb2RkZXIvY29kZW1vZHMvYmFzZV9jb2RlbW9kLnB5)	`100.00% <100.00%> (ø)`	
> [src/codemodder/context.py](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee#diff-c3JjL2NvZGVtb2RkZXIvY29udGV4dC5weQ==)	`100.00% <100.00%> (ø)`	
> [src/codemodder/file_context.py](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee#diff-c3JjL2NvZGVtb2RkZXIvZmlsZV9jb250ZXh0LnB5)	`100.00% <100.00%> (ø)`	
> [src/codemodder/registry.py](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee#diff-c3JjL2NvZGVtb2RkZXIvcmVnaXN0cnkucHk=)	`93.65% <100.00%> (-2.10%)`	⬇️
> [src/codemodder/semgrep.py](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee#diff-c3JjL2NvZGVtb2RkZXIvc2VtZ3JlcC5weQ==)	`95.45% <80.00%> (-0.85%)`	⬇️
> [src/codemodder/executor.py](https://app.codecov.io/gh/pixee/codemodder-python/pull/63?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee#diff-c3JjL2NvZGVtb2RkZXIvZXhlY3V0b3IucHk=)	`89.36% <89.36%> (ø)`	
> ... and [1 file with indirect coverage changes](https://app.codecov.io/gh/pixee/codemodder-python/pull/63/indirect-changes?src=pr&el=tree-more&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pixee)

are you able to see the files in codecov? I just see `No Files covered by tests were changed` which seems off. I'm logged in but idk if it's an account thing

integration_tests/test_multiple_codemods.py
clavedeluna: It could also mean moving some of the methods into a mixin but it's fine.

src/codemodder/executor.py
clavedeluna: ohhhhhhh magic

src/codemodder/codemods/base_codemod.py
clavedeluna: whatttt! I have never seen that! awesome https://stackoverflow.com/a/14836005 I'm going to start using this!

clavedeluna (approved):

drdavella:
@clavedeluna I'm seeing the same thing. I'm hoping we can just ignore the (small) coverage drop for now and address it later.

@andrecsilva you requested changes on this PR but I'm hoping that after our discussion we can defer any potential optimization to a future PR.

andrecsilva (approved):
As long as you understand the highlighted caveats, I'm ok approving this.

PR #64, opened 2023-10-06
New codemod: `safe-lxml-parsing`

clavedeluna:
## Overview
*A new codemod that will check calls to `lxml.etree.parse` and `lxml.etree.fromstring`*

## Description

* [internal docs](https://www.notion.so/pixee/Safe-lxml-parsing-ed786cf78c1347d4856c413249692aac)
* These two functions use a default unsafe parser, so we will set a safe one. If a user does assign a parser, the `safe-lxml-parser-defaults` codemod will kick in (if enabled)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,7 +18,7 @@ jobs:
   linting:
     name: Pylint & Black Checks
     runs-on: ubuntu-20.04
-    timeout-minutes: 3
+    timeout-minutes: 5
     steps:
       - name: Set Up Python
         uses: actions/setup-python@v4
--- a/src/core_codemods/lxml_safe_parsing.py
+++ b/src/core_codemods/lxml_safe_parsing.py
@@ -1,0 +1,52 @@
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import SemgrepCodemod
+from codemodder.codemods.api.helpers import NewArg
+
+
+class LxmlSafeParsing(SemgrepCodemod):
+    NAME = "safe-lxml-parsing"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    SUMMARY = "Use safe parsers in lxml parsing functions"
+    DESCRIPTION = (
+        "Call `lxml.etree.parse` and `lxml.etree.fromstring` with a safe parser"
+    )
+
+    @classmethod
+    def rule(cls):
+        return """
+            rules:
+                - pattern-either:
+                  - patterns:
+                    - pattern: lxml.etree.$FUNC(...)
+                    - pattern-not: lxml.etree.$FUNC(...,parser=..., ...)
+                    - metavariable-pattern:
+                        metavariable: $FUNC
+                        patterns:
+                          - pattern-either:
+                            - pattern: parse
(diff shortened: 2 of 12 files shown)


tests/codemods/test_lxml_safe_parsing.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @each_func
+    def test_from_import(self, tmpdir, func):
+        input_code = f"""from lxml.etree import {func}
+
+{func}("path_to_file")
+var = "hello"
+"""
+        expexted_output = f"""from lxml.etree import {func}
+import lxml
```
clavedeluna: I will point out that this double import is not idea, but it does work and is valid code. Our other codemods (tempfile) also do this at this time.

andrecsilva (approved):

src/core_codemods/lxml_safe_parsing.py
```diff
@@ -0,0 +1,52 @@
...
+                            - pattern: parse
+                            - pattern: fromstring
+                    - pattern-inside: |
+                        import lxml
+                        ...
+        """
+
+    def on_result_found(self, original_node, updated_node):
+        self.remove_unused_import(original_node)
+        self.add_needed_import("lxml")
```
drdavella: I don't think this is sufficient:
```python
>>> import lxml
>>> lxml.etree.XMLParser()
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 lxml.etree.XMLParser

AttributeError: module 'lxml' has no attribute 'etree'
```

I think it needs to be `import lxml.etree`.

Not trying to pick on your PR in particular (because it's very nice overall) but imo this is one fairly strong argument for functional integration tests.

drdavella (changes requested):

src/core_codemods/lxml_safe_parsing.py
clavedeluna: woah, really good catch here. In fact it's so correct there are no examples of `import lxml ... lxml.etree` online! I will update the tests accordingly.

clavedeluna: I actually had to change BOTH lxml codemods, great find! Glad I waited for the docs

drdavella: This really hammers in the need for functional-level tests for me. Any of us could have (and probably have) made similar mistakes, but it's really hard to know whether we're generating correct code unless it actually executes.

I think the right way to think about this is: what kind of test would have alerted you to this problem immediately? In this particular case, I think that just importing the output of the existing integration test would have caught it, so maybe that's a good starting point for functional integration tests.

.github/workflows/lint.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   linting:
     name: Pylint & Black Checks
     runs-on: ubuntu-20.04
-    timeout-minutes: 3
+    timeout-minutes: 5
```
drdavella, line 21: ❤️

src/core_codemods/lxml_safe_parsing.py
clavedeluna: I'm fully convinced! I will follow your lead in prioritizing that work and happy to take it on.

drdavella (approved):

PR #66, opened 2023-10-09
Updated libcst dependency

andrecsilva:
## Overview
- Updated libcst dependency to 1.1.0, removed test_add_imports for redundancy, and small refactor in NameResolutionMixin.


clavedeluna (approved):

drdavella:
@andrecsilva not a big deal but we also need to remove the workaround with the rust toolchain in the dockerfile.

PR #68, opened 2023-10-10
Execute output code in integration tests

clavedeluna:
## Overview
*This PR implements a simple API to run during testing that will execute a codemod's output code*

## Description

We want to have more confidence that the code changes we make actually do work. Executing the code during integration tests will ensure not only imports work but also functions run. Some of the ramifications of that are:
* our testing environment has to have any external lib in any output code
* integration tests are going to take longer. They will also **execute code**: such as making requests, serializing, etc. Whatever the output code is, that will run. We're ok with time cost and risk for now since we consider it minimal risk for the benefit of validating our code
* I had to fix some integration tests to call real imports and to fix any errors that may happen from execution which are expected (like making a request to a bogus url, we don't care about that).
* I did handle at a codemod by codemod basis exceptions that are expected. Doing it this way means we're ok catching a FileNotFound error for a codemod that has `open(...)...` call which we expect if we try to open a file that doesn't exist, but not catch it in other codemods in case that is a legitimate error.
* In the long term we may want to also call `execute_code` in unit tests so I left its API flexible for either path or code str.

### DIFF
--- a/tests/validations.py
+++ b/tests/validations.py
@@ -1,20 +1,30 @@
 import importlib.util
 import tempfile
 
-def validate_code(*, path=None, code=None):
+
+def execute_code(*, path=None, code=None, allowed_exceptions=None):
     """
-    Ensure that code written in `path` or in `code` str is importable.
+    Ensure that code written in `path` or in `code` str is executable.
     """
-    assert (path is None) != (code is None), "Must pass either path to code or code as a str."
+    assert (path is None) != (
+        code is None
+    ), "Must pass either path to code or code as a str."
 
     if path:
-        _try_code_import(path)
+        _run_code(path, allowed_exceptions)
         return
-    with tempfile.NamedTemporaryFile(suffix=".py", mode='w+t') as temp:
+    with tempfile.NamedTemporaryFile(suffix=".py", mode="w+t") as temp:
         temp.write(code)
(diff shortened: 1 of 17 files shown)


tests/validations.py
```diff
@@ -0,0 +1,30 @@
...
+
+def _run_code(path, allowed_exceptions=None):
+    """Execute the code in `path` in its own namespace."""
+    allowed_exceptions = allowed_exceptions or ()
+
+    spec = importlib.util.spec_from_file_location("output_code", path)
+    module = importlib.util.module_from_spec(spec)
+    try:
+        spec.loader.exec_module(module)
+    except allowed_exceptions:
```
drdavella, line 29: This is super cool.

drdavella (approved):
This is awesome. Eventually we can have more sophisticated behavioral assertions if we want, but this is exactly what we need right now.

drdavella:
@clavedeluna it looks like there's maybe some kind of version/compatibility issue with `jwt` causing the tests to fail but once that's fixed we should be good.

clavedeluna:
I had blindly added `jwt` to requirements without realizing what we want is Pyjwt:

Screenshot 2023-10-11 at 7 28 36 AM

andrecsilva (approved):

PR #73, opened 2023-10-13
Script to automate codemod docs

clavedeluna:
## Overview
*Create a script that will automatically create docs for each codemod.*

## Description
- the new script can now be run with `generate-docs ~/docs/docs/codemods/python/` Where the argument is a path to a dir, such as our docs repo. 
- With this output, I created [this PR](https://github.com/pixee/docs/pull/92) which should be reviewed in parallel
- I also added the references field to all codemods and added references for all codemods by copying the urls already listed in each codemod doc. For now the description is just the url but there is a task to come back and actually write descriptions
- With the script running I was able to correct some discrepancies between codemod metadata and docs data, such as review guidance. This should now be prevented!
- I updated codemod summaries (titles) to follow the pattern of "Verb something Noun"
- in general, if you see changes in review guidance or wording it's because the docs repo had the "correct" data

**From now on, any changes to codemod docs should be updated in this repo and the `generate-script` should be run against the local docs repo path and then a PR created**

### DIFF
--- a/src/codemodder/codemods/base_codemod.py
+++ b/src/codemodder/codemods/base_codemod.py
@@ -23,6 +23,21 @@ class CodemodMetadata:
     REVIEW_GUIDANCE: ReviewGuidance
     REFERENCES: list = field(default_factory=list)
 
+    # TODO: remove post_init update_references once we add description for each url.
+    def __post_init__(self):
+        object.__setattr__(self, "REFERENCES", self.update_references(self.REFERENCES))
+
+    @staticmethod
+    def update_references(references):
+        updated_references = []
+        for reference in references:
+            updated_reference = dict(
+                reference
+            )  # Create a copy to avoid modifying the original dict
+            updated_reference["description"] = updated_reference["url"]
+            updated_references.append(updated_reference)
+        return updated_references
+
 
 class BaseCodemod:
     # Implementation borrowed from https://stackoverflow.com/a/45250114
--- a/src/codemodder/codemods/base_codemod.py
+++ b/src/codemodder/codemods/base_codemod.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import Enum
 from typing import List, ClassVar
 
@@ -21,6 +21,7 @@ class CodemodMetadata:
     DESCRIPTION: str  # TODO: this field should be optional
     NAME: str
     REVIEW_GUIDANCE: ReviewGuidance
+    REFERENCES: list = field(default_factory=list)
 
 
 class BaseCodemod:
--- a/src/core_codemods/django_debug_flag_on.py
+++ b/src/core_codemods/django_debug_flag_on.py
@@ -18,6 +18,16 @@ class DjangoDebugFlagOn(SemgrepCodemod, Codemod):
         DESCRIPTION=("Flips django's debug flag if on."),
         NAME="django-debug-flag-on",
         REVIEW_GUIDANCE=ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW,
+        REFERENCES=[
+            {
+                "url": "https://owasp.org/www-project-top-ten/2017/A3_2017-Sensitive_Data_Exposure",
+                "description": "",
+            },
+            {
+                "url": "https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DEBUG",
+                "description": "",
+            },
+        ],
     )
     SUMMARY = CHANGE_DESCRIPTION = "Flip Django debug flag to off"
     YAML_FILES = [
--- a/src/core_codemods/django_debug_flag_on.py
+++ b/src/core_codemods/django_debug_flag_on.py
@@ -15,7 +15,7 @@ from codemodder.codemods.utils import is_django_settings_file
 
 class DjangoDebugFlagOn(SemgrepCodemod, Codemod):
     METADATA = CodemodMetadata(
-        DESCRIPTION=("Flips django's debug flag if on."),
+        DESCRIPTION="Flip Django Debug Flag to Off",
         NAME="django-debug-flag-on",
         REVIEW_GUIDANCE=ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW,
         REFERENCES=[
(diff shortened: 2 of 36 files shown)


src/codemodder/codemods/base_codemod.py
```diff
@@ -22,6 +22,22 @@ class CodemodMetadata:
     DESCRIPTION: str  # TODO: this field should be optional
     NAME: str
     REVIEW_GUIDANCE: ReviewGuidance
+    REFERENCES: list = field(default_factory=list)
+
+    # TODO: remove post_init update_references once we add description for each url.
```
clavedeluna, line 26: This entire green block will be deleted once we add descriptions. I had already added the urls for each codemod so I decided to just automated and copy paste the url into the description field

src/core_codemods/docs/pixee_python_enable-jinja2-autoescape.md
```diff
@@ -1,6 +1,6 @@
 This codemod enables autoescaping of HTML content in `jinja2`. Unfortunately, the jinja2
 default behavior is to not autoescape when rendering templates, which makes your applications
-vulnerable to Cross-Site Scripting (XSS) attacks.
+potentially vulnerable to Cross-Site Scripting (XSS) attacks.
```
clavedeluna, line 3: this was wording in the docs repo

src/scripts/generate_docs.py
```diff
@@ -0,0 +1,181 @@
...
+@dataclass
+class DocMetadata:
+    """Codemod specific metadata only for documentation"""
+
+    importance: str
+    guidance_explained: str
+    need_sarif: str = "No"
+
+
+# codemod-specific metadata that's used only for docs, not for codemod API
```
clavedeluna, line 16: I'm not totally in love with this as a dict but neither do I care for having a separate json file

drdavella: I agree, it's a little awkward either way but this is better than JSON.

src/core_codemods/https_connection.py
```diff
@@ -22,12 +22,24 @@
 
 class HTTPSConnection(BaseCodemod, Codemod):
     METADATA = CodemodMetadata(
-        DESCRIPTION=("Enforce HTTPS connection"),
+        DESCRIPTION="Enforce HTTPS Connection for urllib3",
```
drdavella: We need to fix the API for this, but in general `DESCRIPTION` should not be capitalized because it is getting used as `CHANGE_DESCRIPTION` which is rendered by the platform as a comment on the PR.

src/core_codemods/django_session_cookie_secure_off.py
```diff
@@ -15,11 +15,21 @@
 
 class DjangoSessionCookieSecureOff(SemgrepCodemod, Codemod):
     METADATA = CodemodMetadata(
-        DESCRIPTION=("Sets Django's `SESSION_COOKIE_SECURE` flag if off or missing."),
+        DESCRIPTION=("Sets Django's `SESSION_COOKIE_SECURE` Flag if Off or Missing."),
```
drdavella: I think we need to keep the previous capitalization here (see comment about `DESCRIPTION` -> `CHANGE_DESCRIPTION` -> PR comment).

src/core_codemods/django_debug_flag_on.py
```diff
@@ -15,11 +15,21 @@
 
 class DjangoDebugFlagOn(SemgrepCodemod, Codemod):
     METADATA = CodemodMetadata(
-        DESCRIPTION=("Flips django's debug flag if on."),
+        DESCRIPTION="Flip Django Debug Flag to Off",
```
drdavella: Keep previous capitalization.

src/codemodder/codemods/base_codemod.py
```diff
@@ -22,6 +22,22 @@ class CodemodMetadata:
...
+    def __post_init__(self):
+        object.__setattr__(self, "REFERENCES", self.update_references(self.REFERENCES))
+
+    @staticmethod
+    def update_references(references):
```
drdavella, line 31: I don't see where this method is used at all?

src/core_codemods/process_creation_sandbox.py
```diff
@@ -5,11 +5,21 @@
 
 class ProcessSandbox(SemgrepCodemod):
     NAME = "sandbox-process-creation"
-    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW
-    SUMMARY = "Sandbox process creation"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
```
drdavella: I know you're probably just matching whatever is currently in the docs but I don't think any of our codemods that add dependencies should be `MERGE_WITHOUT_REVIEW` right now.

drdavella (changes requested):
This looks great overall. I think this is going to make our lives easier (and hopefully you agree 😅). My comments are just related to some of the actual content changes. We'll need to iterate on this a bit more going forward but there are a few things here I think we need to address first.

andrecsilva (approved):

src/scripts/generate_docs.py
andrecsilva: I feel like those could go as some sort of optional dataclass field within the codemods themselves, even if those are not used by the API directly.

src/codemodder/codemods/base_codemod.py
clavedeluna: 3 lines above :) ln29

src/core_codemods/django_debug_flag_on.py
clavedeluna: Sorry can you give me the specific change you want? There were 2 changes here

src/core_codemods/https_connection.py
clavedeluna: Aha gotcha, now I understand, will fix.

src/core_codemods/process_creation_sandbox.py
clavedeluna: right I was matching that. I can switch it to merge after cursory review then

src/scripts/generate_docs.py
clavedeluna: I don't disagree but Dan wanted to keep this totally separate from the codemods. I don't mind either way at this time.

drdavella: @clavedeluna is correct: I don't want these to be part of the codemodder API since they don't contribute in any way to user-visible results. This is purely for the purposes of our own documentation.

drdavella (approved):

PR #75, opened 2023-10-16
Report failed parsing

clavedeluna:
## Overview
*Codetf report should include list of fails that filed to be cst parsed ~or codemod-transformed~*

## Description

* previously we have been skipping reporting for a codemod if it had no results, however this doesn't really align with having a list of failed files. All files could fail so there would be no results. We should report anyway
* @drdavella I do need confirmation that this is ok with pixeebot - that a codemod without anything in changeset will still be in the report.
* we now also report fails that were either not correctly cst parsed (already handled) ~and not able to be transformed by a codemod (new)~
* added corresponding test cases to demonstrate.

### DIFF
--- a/.github/workflows/codemod_pygoat.yml
+++ b/.github/workflows/codemod_pygoat.yml
@@ -38,4 +38,4 @@ jobs:
       - name: Run Codemodder
         run: codemodder --output output.codetf pygoat
       - name: Check PyGoat Findings
-        run: make webgoat-test
+        run: make pygoat-test
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,8 @@ test:
 integration-test:
 	${PYTEST} integration_tests
 
-webgoat-test:
-	${PYTEST} -v ci_tests/test_webgoat_findings.py
+pygoat-test:
+	${PYTEST} -v ci_tests/test_pygoat_findings.py
 
 lint:
 	pylint -v codemodder core_codemods tests integration_tests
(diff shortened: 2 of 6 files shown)


clavedeluna:
the pygoat failure may now be expected, need to investigate

andrecsilva (dismissed):

andrecsilva (approved):

src/codemodder/codemodder.py
```diff
@@ -97,13 +97,19 @@ def analyze_files(
...
             sarif_for_file,
         )
-
-        apply_codemod_to_file(
-            execution_context,
-            file_context,
-            codemod,
-            source_tree,
-        )
+        try:
```
drdavella: I don't think there's anything wrong with this change (and I know we discussed it at some other point) but right now it makes me a little concerned that we might cover up any weird edge cases we've missed and so I'd prefer to let this continue to fail loudly for now.

drdavella (changes requested):
Overall looks great but I want to hold off on the one behavior change you introduced for now.

src/codemodder/codemodder.py
clavedeluna: Sounds reasonable and I feel similarly, I just wanted to follow what codetf said which was "record failure if parsing or transformation fails". I can revert

drdavella: Yeah I see now where the spec says that. Still, let's hold off for now and revisit once exceptional behavior is a bit better defined. I think the parser error case makes sense since it's almost certainly out of our control (e.g. a malformed input file) but this case is more likely to be our own mistake.

drdavella (approved):

PR #76, opened 2023-10-16
Implement codemod for replacing unsafe xml methods with defusedxml

drdavella:
## Overview
*Implement codemod for replacing unsafe `xml` methods with `defusedxml`*

## Description

* The Python documentation explicitly calls out the provided `xml` modules as unsafe and recommends the use of `defusedxml`, which is a third-party package
* This codemod replaces the unsafe methods from `xml` with safe ones from `defusedxml`
* In an effort to break this up into smaller units, this PR only contains the transformation logic itself
* I was able to reuse quite a bit of code from the `https-connection` codemod
* **NOTE**: This codemod is **not** enabled in the registry yet. See below for additional work.

## Next Steps
* Add the `defusedxml` dependency. I am leaving this as separate work because I want to make sure we are adding a good CodeTF comment as well, which requires some additional infrastructure.
* Implement an integration test, add documentation, add codemod to the registry

### DIFF
--- a/src/codemodder/cli.py
+++ b/src/codemodder/cli.py
@@ -43,7 +43,7 @@ class CsvListAction(argparse.Action):
     argparse Action to convert "a,b,c" into ["a", "b", "c"]
     """
 
-    def __call__(self, parser, namespace, values: str, option_string=None):
+    def __call__(self, parser, namespace, values, option_string=None):
         # Conversion to dict removes duplicates while preserving order
         items = list(dict.fromkeys(values.split(",")).keys())
         self.validate_items(items)
--- a/src/codemodder/codemods/imported_call_modifier.py
+++ b/src/codemodder/codemods/imported_call_modifier.py
@@ -1,0 +1,109 @@
+import abc
+from typing import Generic, Mapping, Sequence, Set, TypeVar, Union
+
+import libcst as cst
+from libcst import matchers
+from libcst.codemod import CodemodContext, VisitorBasedCodemodCommand
+from libcst.metadata import PositionProvider
+
+from codemodder.change import Change
+from codemodder.codemods.utils_mixin import NameResolutionMixin
+from codemodder.file_context import FileContext
+
+
+# It seems to me like we actually want two separate bounds instead of a Union but this is what mypy wants
+FunctionMatchType = TypeVar("FunctionMatchType", bound=Union[Mapping, Set])
+
+
+class ImportedCallModifier(
+    Generic[FunctionMatchType],
+    VisitorBasedCodemodCommand,
+    NameResolutionMixin,
+    metaclass=abc.ABCMeta,
+):
+    METADATA_DEPENDENCIES = (PositionProvider,)
+
+    def __init__(
+        self,
+        codemod_context: CodemodContext,
+        file_context: FileContext,
+        matching_functions: FunctionMatchType,
+        change_description: str,
+    ):
+        super().__init__(codemod_context)
+        self.line_exclude = file_context.line_exclude
+        self.line_include = file_context.line_include
+        self.matching_functions: FunctionMatchType = matching_functions
+        self.change_description = change_description
+        self.changes_in_file: list[Mapping] = []
+
+    def updated_args(self, original_args: Sequence[cst.Arg]):
+        return original_args
+
+    @abc.abstractmethod
+    def update_attribute(
+        self,
+        true_name: str,
+        original_node: cst.Call,
+        updated_node: cst.Call,
+        new_args: Sequence[cst.Arg],
+    ):
+        """Callback to modify tree when the detected call is of the form a.call()"""
+
+    @abc.abstractmethod
+    def update_simple_name(
+        self,
+        true_name: str,
+        original_node: cst.Call,
+        updated_node: cst.Call,
+        new_args: Sequence[cst.Arg],
+    ):
+        """Callback to modify tree when the detected call is of the form call()"""
+
+    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
+        pos_to_match = self.node_position(original_node)
+        line_number = pos_to_match.start.line
+        if self.filter_by_path_includes_or_excludes(pos_to_match):
+            true_name = self.find_base_name(original_node.func)
+            if (
+                self._is_direct_call_from_imported_module(original_node)
+                and true_name
+                and true_name in self.matching_functions
+            ):
+                self.changes_in_file.append(
+                    Change(str(line_number), self.change_description).to_json()
+                )
+
+                new_args = self.updated_args(updated_node.args)
+
+                # has a prefix, e.g. a.call() -> a.new_call()
+                if matchers.matches(original_node.func, matchers.Attribute()):
+                    return self.update_attribute(
+                        true_name, original_node, updated_node, new_args
+                    )
+
+                # it is a simple name, e.g. call() -> module.new_call()
+                return self.update_simple_name(
+                    true_name, original_node, updated_node, new_args
+                )
+
+        return updated_node
+
+    def filter_by_path_includes_or_excludes(self, pos_to_match):
+        """
+        Returns False if the node, whose position in the file is pos_to_match, matches any of the lines specified in the path-includes or path-excludes flags.
+        """
+        # excludes takes precedence if defined
+        if self.line_exclude:
+            return not any(match_line(pos_to_match, line) for line in self.line_exclude)
+        if self.line_include:
+            return any(match_line(pos_to_match, line) for line in self.line_include)
+        return True
+
+    def node_position(self, node):
+        # See https://github.com/Instagram/LibCST/blob/main/libcst/_metadata_dependent.py#L112
+        return self.get_metadata(PositionProvider, node)
+
+
+def match_line(pos, line):
+    return pos.start.line == line and pos.end.line == line
--- a/src/core_codemods/https_connection.py
+++ b/src/core_codemods/https_connection.py
@@ -1,23 +1,55 @@
-from typing import Sequence
-from libcst import matchers
+from typing import Sequence, Set
+
+import libcst as cst
+from libcst.codemod import Codemod, CodemodContext
 from libcst.codemod.visitors import AddImportsVisitor, RemoveImportsVisitor
-from libcst.metadata import (
-    PositionProvider,
-)
+from libcst.metadata import PositionProvider
+
 from codemodder.codemods.base_codemod import (
     BaseCodemod,
     CodemodMetadata,
     ReviewGuidance,
 )
-from codemodder.change import Change
-from codemodder.codemods.utils_mixin import NameResolutionMixin
-from codemodder.file_context import FileContext
-import libcst as cst
-from libcst.codemod import (
-    Codemod,
-    CodemodContext,
-    VisitorBasedCodemodCommand,
-)
+from codemodder.codemods.imported_call_modifier import ImportedCallModifier
+
+
+class HTTPSConnectionModifier(ImportedCallModifier[Set[str]]):
+    def updated_args(self, original_args):
+        """
+        Last argument _proxy_config does not match new method
+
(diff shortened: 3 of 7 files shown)


src/codemodder/codemods/imported_call_modifier.py
```diff
@@ -0,0 +1,90 @@
...
+                # has a prefix, e.g. a.call() -> a.new_call()
+                if matchers.matches(original_node.func, matchers.Attribute()):
+                    return self.update_attribute(
+                        true_name, original_node, updated_node, new_args
+                    )
+
+                # it is a simple name, e.g. call() -> module.new_call()
+                return self.update_simple_name(
+                    true_name, original_node, updated_node, new_args
+                )
```
andrecsilva, line 88: Just a small clarification om this bit of code here.
Within the context of https codemod this comment and logic was accurate. But this may not be general enough.
Consider:
```python
from lib import call
call()
```
We have the option to simply change the call...
```python
from lib import call
import lib2
lib2.call2()
```
or changing the call and import:
```python
from lib2 import call2
call2()
```
The second has an advantage of not having a possibly useless import (not much of a problem since we have a codemod to amend that). Then again, then reason why I've favored the first one for the https codemod was because the change became more explicit.

The second transformation is a bit harder to do but within the realm of possibility. Just tell me if you want me to support that.

src/core_codemods/https_connection.py
```diff
@@ -1,135 +1,82 @@
...
+        del true_name
+        AddImportsVisitor.add_needed_import(self.context, "urllib3")
+        RemoveImportsVisitor.remove_unused_import_by_node(self.context, original_node)
+        return updated_node.with_changes(
+            args=new_args,
+            func=cst.parse_expression("urllib3.HTTPSConnectionPool"),
+        )
+
+    def count_positional_args(self, arglist: Sequence[cst.Arg]) -> int:
+        for i, arg in enumerate(arglist):
```
clavedeluna: nitpick and personal preference to not use `i` and instead use `idx` (easier to grep)

src/codemodder/codemods/imported_call_modifier.py
```diff
@@ -0,0 +1,90 @@
...
+        """Callback to modify tree when the detected call is of the form a.call()"""
+
+    @abc.abstractmethod
+    def update_simple_name(self, true_name, original_node, updated_node, new_args):
+        """Callback to modify tree when the detected call is of the form call()"""
+
+    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
+        pos_to_match = self.node_position(original_node)
+        line_number = pos_to_match.start.line
+        if self.filter_by_path_includes_or_excludes(pos_to_match):
```
clavedeluna, line 66: just checking in case you're able to make this `leave_Call` be as similar to the one in the api. Maybe not but if so it would be good.

src/core_codemods/use_defused_xml.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+ETREE_METHODS = ["parse", "fromstring", "iterparse", "XMLParser"]
+SAX_METHODS = ["parse", "make_parser", "parseString"]
+DOM_METHODS = ["parse", "parseString"]
+# TODO: add expat methods?
+
+
+class UseDefusedXml(BaseCodemod):
+    NAME = "use-defused-xml"
```
clavedeluna: since the package name is `defusedxml` should this instead be `use-defusedxml`? or `replace-xml-for-defusedxml`? too long prob

src/core_codemods/use_defused_xml.py
```diff
@@ -0,0 +1,97 @@
...
+    SUMMARY = "Use defusedxml for parsing XML"
```
clavedeluna:
```suggestion
    SUMMARY = "Use `defusedxml` for Parsing XML"
```
since it's the title, right?

src/core_codemods/https_connection.py
```diff
@@ -1,135 +1,82 @@
...
+    ]
 
     METADATA_DEPENDENCIES = (PositionProvider,)
 
-    matching_functions = {
-        "urllib3.HTTPConnectionPool",
-        "urllib3.connectionpool.HTTPConnectionPool",
+    matching_functions: dict[str, str] = {
+        "urllib3.HTTPConnectionPool": "urllib3",
+        "urllib3.connectionpool.HTTPConnectionPool": "urllib3",
```
andrecsilva: I don't like this change. Or at least not how it's being used. It may be the case that we want to change it to a different function from another module (i.e. `a.call()` -> `b.func()`). 

I know it's still possible just point the function in the dict itself (e.g. `{a.call() : b.func()}`), as it is being used here, it makes the purpose of `matching_functions` parameter less clear.

In short, I'd prefer if you pointed the function itself here (i.e. `urllib3.HTTPConnectionPool : urllib3.HTTPSConnectionPool`) instead of the lib.

src/core_codemods/use_defused_xml.py
```diff
@@ -0,0 +1,97 @@
...
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_REVIEW
+    DESCRIPTION = "Use method from defusedxml"
```
clavedeluna: I would maybe say "Replace builtin `xml` with `defusedxml`.

src/codemodder/codemods/imported_call_modifier.py
```diff
@@ -0,0 +1,90 @@
...
+    def updated_args(self, original_args: Sequence[cst.Arg]):
+        return original_args
+
+    @abc.abstractmethod
+    def update_attribute(self, true_name, original_node, updated_node, new_args):
+        """Callback to modify tree when the detected call is of the form a.call()"""
+
+    @abc.abstractmethod
+    def update_simple_name(self, true_name, original_node, updated_node, new_args):
+        """Callback to modify tree when the detected call is of the form call()"""
```
andrecsilva: These parameters could use type hints. Even better a small docstring explaining how to use them.

src/core_codemods/use_defused_xml.py
```diff
@@ -0,0 +1,97 @@
...
+    REFERENCES = [
```
clavedeluna, line 47: it seems like references also tend to include OWASP links if available, maybe reuse some of the lxml ones too?

src/core_codemods/use_defused_xml.py
```diff
@@ -0,0 +1,97 @@
...
+        },
+        {
+            "url": "https://pypi.org/project/defusedxml/",
+            "description": "",
+        },
+    ]
+
+    # Mapping from original function to corresponding defusedxml module
+    matching_functions: dict[str, str] = {}
+    matching_functions.update(
```
clavedeluna: it's a bit awkward to have so man updates at the class level, maybe as a separate private method?

src/codemodder/codemods/imported_call_modifier.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
+        pos_to_match = self.node_position(original_node)
+        line_number = pos_to_match.start.line
+        if self.filter_by_path_includes_or_excludes(pos_to_match):
+            true_name = self.find_base_name(original_node.func)
+            if (
+                self._is_direct_call_from_imported_module(original_node)
+                and true_name in self._matching_function_names
```
andrecsilva: Changing it to a list here will degrade the `in` operation from O(1) to O(n) (average case). Either use the original `matching_functions` attribute or at least make `_matching_function_names` a set.

Also, I'm not sure I understood why you've felt the need to create `_matching_function_names` in the first place.

tests/codemods/test_use_defused_xml.py
```diff
@@ -0,0 +1,127 @@
...
+    @pytest.mark.parametrize("module", ["ElementTree", "cElementTree"])
+    def test_etree_simple_call(self, tmpdir, module, method):
+        original_code = f"""
+            from xml.etree.{module} import {method}, ElementPath
+
+            et = {method}('some.xml')
+        """
+
+        new_code = f"""
+            from xml.etree.{module} import ElementPath
```
clavedeluna, line 25: this line is unused so I would've expected the remove import visitor to have removed it

clavedeluna (changes requested):

andrecsilva (changes requested):
My only caveat is the comment about the `_matching_function_names`. Treat the others as suggestions.

src/core_codemods/use_defused_xml.py
drdavella: Yes good idea.

src/codemodder/codemods/imported_call_modifier.py
drdavella: @andrecsilva that makes sense. It might be good to revisit at some point but right now I like that the updated call becomes completely explicit.

tests/codemods/test_use_defused_xml.py
drdavella: That's an interesting point...

src/core_codemods/use_defused_xml.py
drdavella: I think `use-defusedxml` is a good idea.

src/core_codemods/https_connection.py
drdavella: I don't like this change either 😅 but I'm also not sure about the right way to encode this information. In fact, the https codemod doesn't use this at all so maybe there's a cleaner solution.

src/codemodder/codemods/imported_call_modifier.py
drdavella: Yeah it would be good but unfortunately I don't think we have access to those methods since it's in a different class hierarchy.

src/core_codemods/https_connection.py
```diff
@@ -1,135 +1,82 @@
...
+        {
+            "url": "https://urllib3.readthedocs.io/en/stable/reference/urllib3.connectionpool.html#urllib3.HTTPConnectionPool",
+            "description": "",
+        },
+    ]
 
     METADATA_DEPENDENCIES = (PositionProvider,)
 
-    matching_functions = {
+    matching_functions: set[str] = {
```
andrecsilva, line 68: You could have this be a keys-only dict. It would feel more consistent.

andrecsilva (approved):

src/core_codemods/https_connection.py
drdavella: I'm not aware of any such construct; do you just mean a dict where the values are `None`?

drdavella:
Codecov is upset but I don't think I actually changed the coverage here; I just refactored some previously uncovered lines into a new file that it doesn't recognize.

src/codemodder/cli.py
```diff
@@ -43,7 +43,7 @@ class CsvListAction(argparse.Action):
     argparse Action to convert "a,b,c" into ["a", "b", "c"]
     """
 
-    def __call__(self, parser, namespace, values: str, option_string=None):
+    def __call__(self, parser, namespace, values, option_string=None):
```
clavedeluna, line 46: I think this was mypy, right? Isn't it weird that we fix a typing problem by ... removing typing?

tests/codemods/test_use_defused_xml.py
clavedeluna: will this be something you can resolve now or ticket for later?

src/codemodder/cli.py
drdavella: Yeah I (wrongly) introduced this on a previous PR when `mypy` wasn't working for me. The underlying code itself is not typed correctly and I thought adding this would fix it (but it does not).

tests/codemods/test_use_defused_xml.py
drdavella: I'll ticket it for later; I don't think it directly is a result of this PR.

clavedeluna (approved):

PR #77, opened 2023-10-17
Remove py3.9, use py3.12

clavedeluna:
## Overview
*PY3.12 recently came out so we will use it and stop using 3.9*

## Description
* just a couple of changes and fixes to get this to work
* I did look to see if we have some obivous code that would benefit from match/case but surprisingly our number of `else:` statements is pretty small

- [ ] need to update PR necessary checks before merging this

### DIFF
--- a/src/codemodder/registry.py
+++ b/src/codemodder/registry.py
@@ -103,7 +103,7 @@ class CodemodRegistry:
 def load_registered_codemods() -> CodemodRegistry:
     registry = CodemodRegistry()
     logger.debug("loading registered codemod collections")
-    for entry_point in entry_points()["codemods"]:
+    for entry_point in entry_points().select(group="codemods"):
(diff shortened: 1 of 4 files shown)


andrecsilva (dismissed):
I think the dockerfile should also be updated with the new 3.12 image.

src/codemodder/registry.py
```diff
@@ -103,7 +103,7 @@ def match_codemods(
 def load_registered_codemods() -> CodemodRegistry:
     registry = CodemodRegistry()
     logger.debug("loading registered codemod collections")
-    for entry_point in entry_points()["codemods"]:
+    for entry_point in entry_points().select(group="codemods"):
```
drdavella, line 106: ❤️

drdavella (approved):

PR #80, opened 2023-10-17
Pass include/exclude args to semgrep run

clavedeluna:
## Overview
*Like codemodder, semgrep also has include/exclude cli args that we are now going to pass along to improve performance*

## Description

* I'll immediately point out that I was eager to add unit tests, but as soon as I added the flags (initially incorrectly) all tests exploded until I correctly added them. So I think as far as testing we're good since semgrep is used in every single test run.
* semgrep can take multiple of these patterns as `--include=foo.* --include=bar.*'`, just like we do with `--config`

**new scope**
while working on this PR new scope was added and some ticketed:
* we decided that the correct behavior is that incluide/exclude should be **relative to project directory**, not global
1. this means if we call `codemodder tests/samples` then a correct includes path should be `--path-include=unverified_request.py` NOT `--path-include=tests/samples/unverified_request.py`, for example
2. this also meant passing the parent path to semgrep. I discovered some really weird discrepancies which led me to realize we should only pass parent path to **exclude**, not to include. It appears that semgrep's internals have some intersecting behavior between include/excludes, which I could not pin point but could at least determine what we should not be passing to it to get the behavor we want.
3. To correctly handle `fnmatch` behavior, if a pattern  is like `"**.py"` we concat without a "/"

**Follow up work:**
1.  We had unit tests that mimic running `codemodder . ...args`. We don't correctly handle this type of relative path right now so I will ticket it.
2. We are currently passing all include/exclude patterns to semgrep, but in fact we should be handling patterns with line numbers differently.  This behavior was clarified here https://github.com/pixee/codemodder-specs/pull/10 and we will work on it soon.


drdavella (approved):

andrecsilva (approved):
Looks good, but just as a sanity check how does it behave when you pass a line include/exclude with a line number (e.g. `path/to/code.py:42`)

clavedeluna:
good catch, for some reason I thought they were separate flags, need to test that out.

src/codemodder/code_directory.py
```diff
@@ -36,13 +36,26 @@ def file_line_patterns(file_path: str | Path, patterns: Sequence[str]):
...
         else [x for x in (patterns or []) if ":" not in x]
     )
+
+    # TODO: handle case when parent path is "."
+    patterns = [
+        str(Path(parent_path) / Path(pat))
+        if not pat.startswith("*")
+        else parent_path + pat
+        for pat in patterns
+    ]
```
clavedeluna, line 58: I'd like to combine and avoid doing multiple list comp operations but it was honestly easier to iterate and easier to understand separating it. We can refactor later on with a generator or something.

Also, I don't feel super confident with this `if not pat.startswith("*")` logic, but tests show it works right now. happy to revisit

tests/test_code_directory.py
```diff
@@ -102,40 +113,36 @@ def test_match_excluded_precedence_over_included(self, dir_structure):
...
 
     def test_test_directory_not_excluded(self, dir_structure):
-        expected = ["test_insecure_random.py", "test_make_request.py"]
+        expected = ["something.py", "test_insecure_random.py", "test_make_request.py"]
         files = match_files(
             dir_structure, exclude_paths=["**/samples/**", "**/more_samples/**"]
         )
         self._assert_expected(files, expected)
 
-    def test_include_test_overridden_by_default_excludes(self, mocker):
```
clavedeluna, line 111: These tests are using parent_dir as "." which does not work as expected. We will revisit

src/codemodder/code_directory.py
```diff
@@ -36,13 +36,26 @@ def file_line_patterns(file_path: str | Path, patterns: Sequence[str]):
...
     patterns = (
         [x.split(":")[0] for x in (patterns or [])]
         if not exclude
         # An excluded line should not cause the entire file to be excluded
         else [x for x in (patterns or []) if ":" not in x]
     )
+
+    # TODO: handle case when parent path is "."
+    patterns = [
+        str(Path(parent_path) / Path(pat))
```
drdavella, line 54: This is probably faster with `os.path.join`

src/codemodder/code_directory.py
```diff
@@ -36,13 +36,26 @@ def file_line_patterns(file_path: str | Path, patterns: Sequence[str]):
...
+        if not pat.startswith("*")
```
drdavella, line 55: This condition doesn't make sense to me. Do tests fail without it? This means that the behavior is going to be entirely dependent on the presence or absence of a trailing slash in `parent_path`, which I don't think is what we want. If my pattern is `*.py` I want it to mean `/path/to/foo/*.py` and not `/path/to/foo*.py`.

I'm wondering whether removing this also correctly handles the `.` case, although that would also be fixed by normalizing all of the given paths using `os.path.abspath`.

src/codemodder/semgrep.py
```diff
@@ -26,6 +26,24 @@ def run(execution_context: CodemodExecutionContext, yaml_files: List[Path]) -> d
             "-o",
             temp_sarif_file.name,
         ]
+
+        if execution_context.path_exclude:
+            command.extend(
+                itertools.chain.from_iterable(
+                    map(
+                        lambda f: ["--exclude", f"{execution_context.directory}{f}"],
```
drdavella, line 34: It feels like there should be a path separator here.

src/codemodder/semgrep.py
```diff
@@ -26,6 +26,24 @@ def run(execution_context: CodemodExecutionContext, yaml_files: List[Path]) -> d
...
+                        execution_context.path_exclude,
+                    )
+                )
+            )
+        if execution_context.path_include:
+            # Note: parent path is not passed with --include
```
drdavella, line 40: I don't understand why this should be the case.

clavedeluna:
@andrecsilva @drdavella fyi I've request re-review

drdavella (changes requested):
I feel like I'm just becoming more confused by this. I thought that each pattern should just be joined to the parent path and used that way consistently both internally and when passed to semgrep. If that's not working for some reason, I think we need to revisit the design.

src/codemodder/code_directory.py
andrecsilva: Suggestion:
I think you can use a combination of `glob` and `pathlib` to solve issues with globs and relative paths
```
full_path = (Path(parent_path) / Path(pat)).resolve()
files = glob.glob(str(full_path))
```
There are two caveats: (1) resolve will also make it absolute, we may need to handle the current directory carefully, (2) glob will return actual files in the filesystem which is slower than `fnmatch`

drdavella:
@clavedeluna I appreciate your work on this and am sorry that it became a bit of a tar bit. However, I think we're going to pursue a slightly different optimization here so I'm closing this PR (for now).

PR #81, opened 2023-10-17
Ignore tests and other files by default

drdavella:
## Overview
*Ensure test directories and other non-source directories are ignored by default*

## Description

* We now ignore a variety of directories by default
* We also make sure that we respect unix-style glob patterns

### DIFF
--- a/integration_tests/test_multiple_codemods.py
+++ b/integration_tests/test_multiple_codemods.py
@@ -27,12 +27,14 @@ class TestMultipleCodemods:
             "--output",
             str(codetf_path),
             f"--codemod-include={codemods}",
-            f"--path-include={source_file_name}",
+            "--path-include",
+            f"**/{source_file_name}",
         ]
 
         completed_process = subprocess.run(
             command,
             check=False,
+            shell=False,
         )
 
         assert completed_process.returncode == 0
--- a/src/codemodder/cli.py
+++ b/src/codemodder/cli.py
@@ -2,6 +2,7 @@ import argparse
 import sys
 
 from codemodder import __VERSION__
+from codemodder.code_directory import DEFAULT_INCLUDED_PATHS, DEFAULT_EXCLUDED_PATHS
 from codemodder.logging import OutputFormat, logger
 
 
@@ -144,13 +145,13 @@ def parse_args(argv, codemod_registry):
     parser.add_argument(
         "--path-exclude",
         action=CsvListAction,
-        default="",
+        default=DEFAULT_EXCLUDED_PATHS,
         help="Comma-separated set of UNIX glob patterns to exclude",
     )
     parser.add_argument(
         "--path-include",
         action=CsvListAction,
-        default="**/*.py",
+        default=DEFAULT_INCLUDED_PATHS,
         help="Comma-separated set of UNIX glob patterns to include",
     )
 
--- a/src/codemodder/code_directory.py
+++ b/src/codemodder/code_directory.py
@@ -1,35 +1,56 @@
+from typing import Optional, Sequence
+
+import fnmatch
+import itertools
 from pathlib import Path
 
+DEFAULT_INCLUDED_PATHS = ["**.py", "**/*.py"]
+DEFAULT_EXCLUDED_PATHS = [
+    "**/test/**",
+    "**/tests/**",
+    "**/conftest.py",
+    "**/build/**",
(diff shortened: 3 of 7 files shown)


src/codemodder/code_directory.py
```diff
@@ -1,65 +1,79 @@
...
-        file_path.match(pattern.rsplit(":")[0]) for pattern in patterns
-    )
+from typing import Optional, Sequence
 
+import fnmatch
+import itertools
+from pathlib import Path
 
-def file_line_patterns(file_path, patterns):
+DEFAULT_INCLUDED_PATHS = ["**.py", "**/*.py"]
```
andrecsilva, line 7: I believe the second glob pattern includes the files from the first one.

andrecsilva (approved):

integration_tests/test_multiple_codemods.py
```diff
@@ -27,12 +27,14 @@ def test_two_codemods(self, codemods, tmpdir):
...
             f"--codemod-include={codemods}",
-            f"--path-include={source_file_name}",
+            "--path-include",
+            f"**/{source_file_name}",
         ]
 
         completed_process = subprocess.run(
             command,
             check=False,
+            shell=False,
```
clavedeluna, line 37: need to change this in base_test.py too?

tests/test_code_directory.py
```diff
@@ -87,11 +96,18 @@ def test_match_excluded_precedence_over_included(self, dir_structure):
         expected = ["empty_for_testing.py", "insecure_random.py"]
         files = match_files(
             dir_structure,
-            exclude_paths=["*request.py"],
+            exclude_paths=["**/tests/**", "*request.py"],
             include_paths=["*request.py", "*empty*.py", "*random.py"],
         )
         self._assert_expected(files, expected)
 
+    def test_test_directory_not_excluded(self, dir_structure):
```
clavedeluna, line 104: given all the indexing and iteration now in code_directory.py I was hoping to see more unit tests, including times when a user may pass in non-glob compliant patterns. also test a few of the default patterns. Maybe it's already there and I just didn't see the tests, but pointing it out

src/codemodder/cli.py
```diff
@@ -144,13 +145,13 @@ def parse_args(argv, codemod_registry):
     parser.add_argument(
         "--path-exclude",
         action=CsvListAction,
-        default="",
+        default=DEFAULT_EXCLUDED_PATHS,
         help="Comma-separated set of UNIX glob patterns to exclude",
```
clavedeluna, line 149: just noticing from my semgrep PR that we need the defaults exclude/include to always be lists, which you've done in this PR. Pointing that out since the help str indicates command sep which is what the user does, then gets converted to a list and we always assume it's a list.

src/codemodder/code_directory.py
drdavella: @andrecsilva I thought so too but it appears not to be the case:
```python
>>> fnmatch("foo.py", "**.py")
True
>>> fnmatch("foo.py", "**/*.py")
False
>>>
```

integration_tests/test_multiple_codemods.py
drdavella: I think I might have done that already but I'll check.

tests/test_code_directory.py
drdavella: Yeah I relied on the existing tests a lot for this work but I could add some more.

src/codemodder/cli.py
drdavella: Yup I added some type annotations to clarify that for the `code_directory` API.

tests/test_code_directory.py
drdavella: @clavedeluna I added a few more tests showing default behavior and override with respect to test directories. I'm not sure what to do about non-compliant patterns; that's basically user error and it just means things aren't going to work they way they expect.

src/codemodder/code_directory.py
clavedeluna: good test case to add somehow bc I thought the same thing. Honestly the best thing about this PR is adding fnmatch!

tests/test_code_directory.py
clavedeluna: Yeah that's fair.

clavedeluna (approved):

PR #83, opened 2023-10-18
Add optional JSON logging

drdavella:
## Overview
*Implement JSON logging per the [logging spec](https://github.com/pixee/codemodder-specs/blob/main/logging.md#json-output)*

```
{"timestamp": "2023-10-18 14:36:18,122", "level": "INFO", "message": "\n[startup]", "file": "logging.py", "line": 46}
{"timestamp": "2023-10-18 14:36:18,122", "level": "INFO", "message": "codemodder: python/0.57.0", "file": "codemodder.py", "line": 135}
{"timestamp": "2023-10-18 14:36:18,122", "level": "INFO", "message": "\n[setup]", "file": "logging.py", "line": 46}
{"timestamp": "2023-10-18 14:36:18,122", "level": "INFO", "message": "running:", "file": "logging.py", "line": 53}
```

### DIFF
--- a/src/codemodder/codemodder.py
+++ b/src/codemodder/codemodder.py
@@ -129,7 +129,7 @@ def run(original_args) -> int:
         )
         return 1
 
-    configure_logger(argv.verbose, argv.log_format)
+    configure_logger(argv.verbose, argv.log_format, argv.project_name)
 
     log_section("startup")
     logger.info("codemodder: python/%s", __VERSION__)
--- a/src/codemodder/codemodder.py
+++ b/src/codemodder/codemodder.py
@@ -129,7 +129,7 @@ def run(original_args) -> int:
         )
         return 1
 
-    configure_logger(argv.verbose)
+    configure_logger(argv.verbose, argv.log_format)
 
     log_section("startup")
     logger.info("codemodder: python/%s", __VERSION__)
--- a/src/codemodder/logging.py
+++ b/src/codemodder/logging.py
@@ -13,6 +13,10 @@ class OutputFormat(Enum):
     HUMAN = "human"
     JSON = "json"
 
+    def __str__(self):
+        """For rendering properly in argparse help."""
+        return self.value.lower()
+
 
 def log_section(section_name: str):
     """
--- a/src/codemodder/logging.py
+++ b/src/codemodder/logging.py
@@ -1,6 +1,7 @@
 from enum import Enum
 import logging
 import sys
+from typing import Optional
 
 from pythonjsonlogger import jsonlogger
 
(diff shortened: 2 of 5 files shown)


drdavella:
I don't really understand what's happening here, when I run this locally it's telling me logging.py is even more covered than it was before:
```
src/codemodder/logging.py                                             46      8    83%
```

src/codemodder/logging.py
```diff
@@ -13,6 +16,28 @@ class OutputFormat(Enum):
...
+
+
+class CodemodderJsonFormatter(jsonlogger.JsonFormatter):
+    project_name: Optional[str]
+
+    def __init__(self, *args, project_name: Optional[str] = None, **kwargs):
+        self.project_name = project_name
+        super().__init__(*args, **kwargs)
+
+    def add_fields(self, log_record, record, message_dict):
```
drdavella, line 31: I'm tempted to ignore this block for the purposes of coverage. I tried to write some tests here but it's proving fairly difficult to get the mocks right.

src/codemodder/codemodder.py
```diff
@@ -129,7 +129,7 @@ def run(original_args) -> int:
         )
         return 1
```
drdavella, line 130: Deos this work?

andrecsilva (approved):

src/codemodder/logging.py
```diff
@@ -30,22 +55,33 @@ def log_list(level: int, header: str, items: list, predicate=None):
...
 
-    # TODO: this should all be conditional on the output format
     stdout_handler = logging.StreamHandler(sys.stdout)
     stdout_handler.setLevel(log_level)
-    stdout_handler.addFilter(lambda record: record.levelno <= logging.WARNING)
+    handlers = [stdout_handler]
 
-    stderr_handler = logging.StreamHandler(sys.stderr)
-    stderr_handler.setLevel(logging.ERROR)
+    match log_format:
```
clavedeluna, line 70: <3

src/codemodder/logging.py
clavedeluna: I'm super surprised codecov says most of this isn't tested. Curious, if you put a debugger in here and run codemodder as pixeebot would, it does it it, right? That would tell me integration tests would test this but right now we only calc coverage based on unit tests. This makes sense bc the way we run integration tests with subprocess.run creates a separate process and pytest / cov can't calculate coverage there.

Is there any help from the python-json-logger docs in regards to testing? Does chatgpt help in any way here? If you've tried these avenues and felt like you've spent sufficient time, I would be ok with adding no-cover here though it does break my heart a little.

drdavella: This codepath is only covered if we use `--log-format=json` which probably doesn't happen in any of the integration tests right now. This basically just boils down to level of effort vs reward right now and I can't afford to spend any more time trying to get this covered.

clavedeluna: Totally fine, ticket and move on. Could be a good starter task for someone who may join in the upcoming months or just an easy "end of day" task.

clavedeluna (approved):

PR #86, opened 2023-10-20
Add ChangeSet to CodeTF when adding dependency to requirements.txt

drdavella:
## Overview
*Dependency updates now result in new `ChangeSet` entry in CodeTF*

## Description

* Whenever we add a dependency the change needs to be reflected as a `ChangeSet` entry in CodeTF
* Each codemod that introduces a dependency will contain its own `ChangeSet` entry
* This required some changes to the `DependencyManager` since it needs to be able to generate a diff and provide some contextual information
  * It is much easier to iterate on `DependencyManager` if it's in the same git source tree. 
  * We can revisit whether this belongs in a separate package at a later date.
  * There is going to be some more iteration on this component as we add support for other dependency locations
* `use-defusedxml` now adds a dependency but there is still additional work before it can be enabled
* We also need to add contextual comments for dependency changes so that they are available to the platform


andrecsilva (approved):
Looks good, still not sure about integrating the DM here to codemodder, but at least it will make things easier. Maybe we can separate them later when the project is a bit more mature.

drdavella:
@andrecsilva I agree that we can/should revisit this eventually. There are two major considerations right now:
* We expect to make many changes to the dependency manager in the near future and need to iterate quickly
* It ends up being more tightly coupled to the concept of codemods than originally expected. The entire purpose of the dependency manager is to enable modifications to dependency files. So at the very least, it needs to return a filename, a diff, and a list of changed lines. This is basically a `ChangeSet` so it feels a little arbitrary to draw a strict package boundary at that point when it would require translation between two similar datatypes. However, I also see how this could be of more general use, so I think it's a good idea to revisit eventually.

PR #90, opened 2023-10-23
Initial Version of SQL Parameterizer Codemod

andrecsilva:
## Overview
Adds a codemod that parameterizes SQL queries.


src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+                                    gparent.value.visit(visitor)
+                                    return visitor.leaves
+        return [node]
+
+    def recurse_Attribute(self, node: cst.Attribute) -> list[cst.CSTNode]:
+        # TODO attributes may have been assigned, should those be modified?
+        # research how to detect attribute assigns in libcst
+        return [node]
+
+    def _find_gparent(self, n: cst.CSTNode) -> Optional[cst.CSTNode]:
```
drdavella, line 262: We have some similar logic in `use-walrus-if`: https://github.com/pixee/codemodder-python/blob/7112886e9ea2da9dc0a9b961c21fe108031cf099/src/core_codemods/use_walrus_if.py#L93-L95
It might be useful to generalize this into a separate utility function.

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+                                ] + value.sequence
+                            case _:
+                                changes_dict[key] = value
+                    return updated_node.with_changes(**changes_dict)
+                case cst.CSTNode():
+                    return replacement
+        return updated_node
+
+
+class SQLQueryParameterization(BaseCodemod, Codemod):
```
drdavella: I think this ends up being cleaner if we use `codemodder.api.BaseCodemod` as the base here instead. It requires a few tweaks to the metadata fields.

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+    )
+    SUMMARY = "Parameterize SQL queries."
+    CHANGE_DESCRIPTION = ""
+
+    METADATA_DEPENDENCIES = (
+        PositionProvider,
+        ScopeProvider,
+        ParentNodeProvider,
+    )
+    METADATA_DEPENDENCIES = (
```
drdavella: This looks like copy/paste error? Unless I'm misunderstanding, only the second of these is actually used.

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+        Codemod.__init__(self, context)
+        BaseCodemod.__init__(self, *codemod_args)
+
+    def _build_param_element(self, middle, index: int) -> cst.BaseExpression:
+        # TODO maybe a parameterized string would be better here
+        # f-strings need python 3.6 though
+        if index == 0:
+            return middle[0]
+        operator = cst.Add(
+            whitespace_after=cst.SimpleWhitespace(" "),
```
drdavella, line 81: Minor comment but this whitespace shows up often enough that I wonder whether it could be factored out as an attribute on the class.

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+                modulo_2 = 0
+                leaves.append(end)
+            else:
+                modulo_2 = 1
+
+    def _is_not_a_single_quote(self, expression: cst.CSTNode) -> bool:
+        match expression:
+            case cst.SimpleString():
+                prefix = expression.prefix.lower()
+                if "b" in prefix:
```
drdavella, line 325: Shouldn't this just be a `match`? At the very least it seems like it should be `==` instead of `in`.

tests/samples/sql_injection.py
```diff
@@ -0,0 +1,14 @@
...
+
+def foo(cursor: sqlite3.Cursor, name: str, phone: str):
+    a = "SELECT * FROM Users"
+    b = " WHERE name ='" + name
+    c = "' AND phone = '" + phone + "'"
+    r = cursor.execute(a + b + c)
+    print(r.fetchone())
+
+
+foo(connection.cursor(), "Jenny", "867-5309")
```
drdavella, line 16: 😆

tests/transformations/test_remove_empty_string_concatenation.py
```diff
@@ -0,0 +1,82 @@
...
+
+class RemoveEmptyStringConcatenationCodemod(Codemod):
+    def transform_module_impl(self, tree: cst.Module) -> cst.Module:
+        return tree.visit(RemoveEmptyStringConcatenation())
+
+
+class TestRemoveEmptyStringConcatenation(CodemodTest):
+    TRANSFORM = RemoveEmptyStringConcatenationCodemod
+
+    def test_left(self):
```
drdavella, line 17: This is a little bit pedantic but could we also have a test case involving multiple operators?
```python
"" + "whatever" + ""
```
and
```python
"hello" + "" + " world"
```

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+        return False
+
+
+class FindQueryCalls(ContextAwareVisitor):
+    """
+    Find all the execute calls with a sql query as an argument.
+    """
+
+    # right now it works by looking into some sql keywords in any pieces of the query
+    # Ideally we should infer what driver we are using
```
drdavella, line 404: I'm just not terribly optimistic that this is going to be possible in general, but we can revisit.

tests/codemods/test_sql_parameterization.py
```diff
@@ -0,0 +1,262 @@
...
+
+            a = "SELECT * from USERS "
+            b = "WHERE name = '" + name
+            c = "' AND phone = ?"
+            return cursor.execute(a + b + c, (phone,))
+        """
+        self.run_and_assert(tmpdir, dedent(input_code), dedent(input_code))
+        assert len(self.file_context.codemod_changes) == 0
+
+    def test_wont_change_class_attribute(self, tmpdir):
```
drdavella, line 237: I'm not sure I fully understand the limitation here and below. Is the idea that we can't be sure whether these kinds of attributes are being modified elsewhere? If that's the case, could you add a comment or docstring to each of these tests?

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+    """
+    Detects injections and gather the expressions that are injectable.
+    """
+
+    METADATA_DEPENDENCIES = (
+        ScopeProvider,
+        ParentNodeProvider,
+    )
+
+    quote_pattern = re.compile(r"(?<!\\)\\'|(?<!\\)'")
```
drdavella, line 282: I might be misreading the regexes here and below but it looks like the `?` in this case is actually part of the pattern and we really should be using the value of `parameter_token` when building the regex.

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+        return False
+
+    def recurse_Name(self, node: cst.Name) -> list[cst.CSTNode]:
+        assignment = self.find_single_assignment(node)
+        if assignment:
+            base_scope = assignment.scope
+            # TODO make this check in detect injection, to be more precise
+
+            # Ensure that this variable is not used anywhere else
+            # variables used in the global scope / class scope may be referenced in other files
```
drdavella, line 237: 👍 I think this explains my question below but it would still be good to have a comment in the test cases.

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+    def __init__(self, sequence: list[cst.CSTNode]) -> None:
+        self.sequence = sequence
+
+
+class Prepend:
+    def __init__(self, sequence: list[cst.CSTNode]) -> None:
+        self.sequence = sequence
+
+
+class ReplaceNodes(cst.CSTTransformer):
```
drdavella: Seems like it could be more generally useful, maybe this belongs in a utility module in the `codemodder` package?

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+                    first_arg.value.visit(query_visitor)
+                    for expr in query_visitor.leaves:
+                        match expr:
+                            case cst.SimpleString() | cst.FormattedStringText() if self._has_keyword(
+                                expr.value
+                            ):
+                                self.calls[original_node] = query_visitor.leaves
+
+
+def _get_function_name_node(call: cst.Call) -> Optional[cst.Name]:
```
drdavella: Just an opinion but I think whenever we're adding module-level functions to codemods like this it's a sign that it probably belongs in a separate utility module.

src/core_codemods/docs/pixee_python_sql-parameterization.md
```diff
@@ -0,0 +1,23 @@
...
++ cursor.execute("SELECT * from USERS WHERE name =?", (name, ))
+```
+
+If you have feedback on this codemod, [please let us know](mailto:feedback@pixee.ai)!
+
+## F.A.Q.
+
+### Why is this codemod marked as Merge With Cursory Review
+
+Python has a wealth of database drivers that all use the same interface. Different drivers may require different string tokens used for parameterization, and Python's dynamic typing makes it quite hard, and sometimes impossible, to detect which driver is being used just by looking at the code.
```
drdavella: This is a nice explanation. Let's explicitly call out the name `dbapi2` here with a link: https://peps.python.org/pep-0249/

drdavella (changes requested):
This is some really impressive work overall. I'm relying heavily on the test cases to help me understand the feature. Most of my comments are minor in themselves but I think they all add up to a requested change.

src/core_codemods/sql_parameterization.py
andrecsilva: You're right, it should not be there.

src/core_codemods/sql_parameterization.py
andrecsilva: You can have multiple prefixes in a string, like `bR` or `RF`". See:
https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

src/core_codemods/sql_parameterization.py
andrecsilva: The `(?<!...)` piece is a negative lookback assertion. See:
https://docs.python.org/3/library/re.html#regular-expression-syntax

src/core_codemods/sql_parameterization.py
andrecsilva: This is a hack-ish version of the simpler `ReplaceNodes` in `codemodder.codemod.utils`. I'm not entirely sure it's a drop-in replacement (seems like it but I need to test it properly). It mimics the `with_changes(...)` interface of libcst with a few additions. I'll replace it and make sure it works.

integration_tests/test_sql_parameterization.py
```diff
@@ -0,0 +1,41 @@
...
+    original_code, expected_new_code = original_and_expected_from_code_path(
+        code_path,
+        [
+            (7, """    b = " WHERE name =?"\n"""),
+            (8, """    c = " AND phone = ?"\n"""),
+            (9, """    r = cursor.execute(a + b + c, (name, phone, ))\n"""),
+        ],
+    )
+
+    # fmt: off
```
clavedeluna, line 20: oh nice, we should add this elsewhere as separate work

src/core_codemods/docs/pixee_python_sql-parameterization.md
```diff
@@ -0,0 +1,23 @@
+This codemod refactors SQL statements to be parameterized, rather than built by hand.
```
clavedeluna, line 1: this docs file should only include everything up to  the code diff, without `If you have feedback....` You will also need to update `generate_docs` there instead.

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+
+
+class Append:
+    def __init__(self, sequence: list[cst.CSTNode]) -> None:
+        self.sequence = sequence
+
+
+class Prepend:
+    def __init__(self, sequence: list[cst.CSTNode]) -> None:
+        self.sequence = sequence
```
clavedeluna: Append and Prepend are identical, if you want to still keep a diff namespace you can just do `Prepend = Append` without a second impl

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+                                changes_dict[key] = value
+                    return updated_node.with_changes(**changes_dict)
+                case cst.CSTNode():
+                    return replacement
+        return updated_node
+
+
+class SQLQueryParameterization(BaseCodemod, Codemod):
+    METADATA = CodemodMetadata(
+        DESCRIPTION=("Parameterize SQL queries."),
```
clavedeluna: this should be either a little longer than SUMMARY or just assign SUMMARY = DESCRIPTION = CHANGE_DESCRIPTION

src/core_codemods/sql_parameterization.py
```diff
@@ -0,0 +1,484 @@
...
+                        str(line_number), SQLQueryParameterization.CHANGE_DESCRIPTION
+                    ).to_json()
+                )
+                result = result.visit(RemoveEmptyStringConcatenation())
+
+        return result
+
+    # Replace node entirelly -> CSTNode
+    # update some attribute -> dict[str,any]
+    # apend or prepend an argument -> Append([]) / Preprend([])
```
clavedeluna: move # into docstring

tests/codemods/test_sql_parameterization.py
```diff
@@ -0,0 +1,262 @@
...
+        import sqlite3
+
+        name = input()
+        phone = input()
+        connection = sqlite3.connect("my_db.db")
+        cursor = connection.cursor()
+        cursor.execute("SELECT * from USERS WHERE name =?" + " AND phone =?",  (name, phone, ))
+        """
+        self.run_and_assert(tmpdir, dedent(input_code), dedent(expected))
+        assert len(self.file_context.codemod_changes) == 1
```
clavedeluna, line 52: how about moving this line to `run_and_assert`

clavedeluna (changes requested):

src/core_codemods/sql_parameterization.py
andrecsilva: These were just some random notes that I was jotting down while thinking on how to solve a problem that originated within the method below. The solution is now the `ReplaceNodes` transform in the same file.

I've deleted those and moved the explanation to `ReplaceNodes` docstring.

tests/codemods/test_sql_parameterization.py
andrecsilva: This change will touch every unit file. Let's do it in another PR.

integration_tests/test_sql_parameterization.py
andrecsilva: I've tried to use a multiline string first, but `textwrap.dedent` will eliminate trailing whitespaces from "empty" lines and the pragma won't work if it's not properly indented.

src/core_codemods/sql_parameterization.py
andrecsilva: I want it to inherit from libcst's `Codemod` and not `BaseTransformer` -> `VisitorBasedCodemodCommand`.

src/core_codemods/sql_parameterization.py
andrecsilva: We could go the way of `NameResolutionMixin` and make something like `ParentProviderMixin`, but this feels like a small enough operation that it does not warrant a whole class for it. Besides I can't think of another operation that would go into that class and be used in other codemods for now.

tests/codemods/test_sql_parameterization.py
clavedeluna: cool as long as it's ticketed I"m happy

src/core_codemods/sql_parameterization.py
clavedeluna: Yeah and as long as we inherit from libcst's `Codemod` we can't use our current API. We'll have to create api base classes for this later on

src/core_codemods/sql_parameterization.py
andrecsilva: If I make `Prepend = Append` then any instance of `Prepend` will match with `Append()`. I've extracted a common base class for both instead.

src/core_codemods/docs/pixee_python_sql-parameterization.md
clavedeluna: file still contains "If you have feedback ... " and everything after it

tests/samples/my_db.db
drdavella, line 1: I don't think this was intended to be committed. Can you add a pattern to `.gitignore` so we avoid this in the future?

andrecsilva:
The 100% coverage here is not happening unless I remove the formatted string placeholders and some sanity checks (e.g. the exception in `_find_gparent()`).

.pre-commit-config.yaml
```diff
@@ -6,10 +6,6 @@ repos:
     -   id: check-json
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
-        exclude: |
```
drdavella: I'm not sure why but this reverts the change I introduced on `main`.

drdavella (changes requested):
Looks good except for two minor comments that need to be addressed.

tests/samples/my_db.db
andrecsilva: We need this to run the integration test without issues.

.pre-commit-config.yaml
andrecsilva: Good catch, I probably messed up when solving the issues while rebasing.

src/core_codemods/sql_parameterization.py
drdavella: I was thinking something more like a simple function/method that returned a grandparent node but it's not that important.

tests/samples/my_db.db
drdavella: In general I'd like to avoid committing binary artifacts like this to the source tree. I don't want to hold this PR up any longer but eventually we should create this database as part of the test setup and then remove it at teardown.

drdavella (approved):

tests/samples/my_db.db
andrecsilva: I just learned that you can create a memory-only database with the  ":memory:" argument in `sqlite3`. I'll adjust the integration test and remove the extra file.

clavedeluna (approved):

drdavella (approved):

PR #97, opened 2023-10-25
Initial Python Repo Manager

clavedeluna:
## Overview
*The purpose here is to kickstart the code for analyzing a project directory, its metadata and dependencies.*

## Description

* a `PythonRepoManager`'s goal is to take the parent directory passed to codemodder and to analyze that repo's files for existence of any pkg management system or dependency store, such as requirements.txt, project.toml, setup.py, setup.cfg, poetry's more special project.toml, Pipfile, conda's .yaml files.
* flexibility is key here as each file type is unique and a repo may contain none or many of these.
* I've added this architecture but codemodder is not running it at this time. I've also copied some code from DependencyManager which will be deduped later on. This will come soon.

There is some key work that needs to happen going forward:
1. Dudupe / connect `PythonRepoManager` with `DependencyManager`. I think ideally DM will be inside PRM, but we shall see as a next step. There's quite a bit of changes going on to DM right now so I'd like to sandbox it
2. Handle edge cases. I've added quite a few todos to each parser to handle all sorts of edge cases in each file.
3. Add remaining file parsers: I added a few but we still need: poetry's more special project.toml, Pipfile, conda's .yaml files.
4. Higher-level testing:  maybe have an integration test that runs PRM on known OSS packages and see the results we get. This would add variety to see if we need edge-case handling.
5. Heuristics for how to handle a parent_directory with 0 or more of these file types. Where will we add dependencies we need to add? How do we handle dependencies declared in multiple files? How do we handle conflicting information? Do we only take the first found file? Lots of discussion we need here.


src/codemodder/project_analysis/file_parsers/setup_py_file_parser.py
```diff
@@ -0,0 +1,77 @@
...
+
+        return PackageStore(
+            type=self.file_name,
+            file=str(file),
+            dependencies=self._parse_dependencies_from_cst(visitor.install_requires),
+            py_versions=self._parse_py_versions(visitor.python_requires),
+        )
+
+
+class SetupCallVisotor(cst.CSTVisitor):
```
andrecsilva: Visotor -> Visitor

clavedeluna:
the codecov failure isn't legitimate to me. It mentions lack of coverage on a file I didn't even touch.

src/codemodder/project_analysis/file_parsers/utils.py
```diff
@@ -0,0 +1,7 @@
+import libcst as cst
+
+
+def clean_simplestring(node: cst.SimpleString | str) -> str:
+    if isinstance(node, str):
+        return node.strip('"')
+    return node.value.strip('"')
```
andrecsilva: Use `node.raw_value`. It will strip the quotes and prefixes.

andrecsilva (dismissed):
Minor changes. See comments.

drdavella:
> poetry's more special project.toml, Pipfile, conda's .yaml files

These are good call-outs. I hope we can defer these until later and just try to handle the more "standard" cases for now.

> Heuristics for how to handle a parent_directory with 0 or more of these file types. Where will we add dependencies we need to add? How do we handle dependencies declared in multiple files? How do we handle conflicting information? Do we only take the first found file? Lots of discussion we need here.

A lot of really good points here too. I think the simplest heuristic will be to have some kind of priority order that we define in advance, and make updates only to the first of those files that we detect. We can build out more functionality from that point. It feels like we're paving a new path here, though, and we're on our own to figure it out.

src/codemodder/context.py
```diff
@@ -39,14 +40,16 @@ def __init__(
...
+        repo_manager: PythonRepoManager,
+    ):  # pylint: disable=too-many-arguments
         self.directory = directory
         self.dry_run = dry_run
         self.verbose = verbose
         self._results_by_codemod = {}
         self._failures_by_codemod = {}
         self.dependencies = {}
         self.registry = registry
+        self.repo_manager = repo_manager
```
drdavella, line 53: Very minor thing but we should add the type declaration at the class definition above.

src/codemodder/project_analysis/file_parsers/base_parser.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class BaseParser(ABC):
+    def __init__(self, parent_directory: Path):
+        self.parent_directory = parent_directory
+
+    @property
+    @abstractmethod
+    def file_name(self):
+        ...  # pragma: no cover
```
drdavella, line 16: Just a tiny thing but I think you can have the `...` on the same line as the `def` which probably makes the `# pragma` unnecessary.

src/codemodder/project_analysis/file_parsers/setup_cfg_file_parser.py
```diff
@@ -0,0 +1,37 @@
...
+
+from .base_parser import BaseParser
+
+
+class SetupCfgParser(BaseParser):
+    @property
+    def file_name(self):
+        return "setup.cfg"
+
+    def _parse_dependencies_from_toml(self, config: configparser.ConfigParser):
```
drdavella: Copy paste probably, I don't think this is `toml`.

drdavella (approved):
This looks great. I'm sure there is going to be a lot of iteration on this and the dependency manager, but this gives us a really good foundation to build upon. Just a few minor comments but otherwise 👍

src/codemodder/project_analysis/file_parsers/base_parser.py
clavedeluna: oh I hope so!

src/codemodder/context.py
clavedeluna: Are you referring to something else that's not already line 43, `repo_manager: PythonRepoManager,`?

drdavella: Yes, it should be added on line 35 above too.

src/codemodder/project_analysis/file_parsers/base_parser.py
clavedeluna: black doesn't like inline ... so I left it as is

PR #100, opened 2023-10-25
Implement file-level parallelization

drdavella:
## Overview
*Implement (optional) per-codemod file-level parallelization*

## Description

* This PR enables files to be processed in parallel for each codemod
* There is no parallelism _between_ codemods since we need to preserve strict ordering
* The default behavior for now is to use a single thread, so there is no difference from the previous behavior
* All per-file codemod state is now managed by `FileContext` and then aggregated at the end of each codemod pass
* This will enable us to perform testing in more resource constrained environments where more threads might be useful. In particular, we suspect that environments that are highly limited by IO bandwidth will benefit from parallelization

### DIFF
--- a/src/codemodder/cli.py
+++ b/src/codemodder/cli.py
@@ -158,6 +158,12 @@ def parse_args(argv, codemod_registry):
         default=DEFAULT_INCLUDED_PATHS,
         help="Comma-separated set of UNIX glob patterns to include",
     )
+    parser.add_argument(
+        "--max-workers",
+        type=int,
+        default=1,
+        help="maximum number of workers (threads) to use for processing files in parallel",
+    )
 
     # At this time we don't do anything with the sarif arg.
     parser.add_argument(
--- a/src/codemodder/codemodder.py
+++ b/src/codemodder/codemodder.py
@@ -1,3 +1,4 @@
+from concurrent.futures import ThreadPoolExecutor
 import datetime
 import difflib
 import logging
@@ -29,18 +30,14 @@ def update_code(file_path, new_code):
 
 
 def apply_codemod_to_file(
-    execution_context: CodemodExecutionContext,
+    base_directory: Path,
     file_context,
     codemod_kls: CodemodExecutorWrapper,
     source_tree,
+    dry_run: bool = False,
 ):
-    name = codemod_kls.id
     wrapper = cst.MetadataWrapper(source_tree)
-    codemod = codemod_kls(
-        CodemodContext(wrapper=wrapper),
-        execution_context,
-        file_context,
-    )
+    codemod = codemod_kls(CodemodContext(wrapper=wrapper), file_context)
     if not codemod.should_transform:
         return False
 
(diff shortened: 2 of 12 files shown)


src/codemodder/cli.py
```diff
@@ -158,6 +158,12 @@ def parse_args(argv, codemod_registry):
         default=DEFAULT_INCLUDED_PATHS,
         help="Comma-separated set of UNIX glob patterns to include",
     )
+    parser.add_argument(
+        "--max-workers",
+        type=int,
+        default=1,
+        help="maximum number of workers (threads) to use for parallelism",
```
clavedeluna: may be worth mentioning this is within each codemod parallelism, not across codemods

src/codemodder/codemodder.py
```diff
@@ -57,59 +54,86 @@ def apply_codemod_to_file(
...
-        # NOTE: file context will become more important if/when we parallelize this loop
-        file_context = FileContext(
-            file_path,
-            line_exclude,
-            line_include,
-            sarif_for_file,
+    with ThreadPoolExecutor(max_workers=cli_args.max_workers) as executor:
+        logger.debug(
+            "using executor with %s threads",
+            executor._max_workers,  # pylint: disable=protected-access
```
clavedeluna: any reason not to use `cli_args.max_workers` here like the lines above?

src/codemodder/codemodder.py
```diff
@@ -29,18 +30,14 @@ def update_code(file_path, new_code):
...
+    dry_run: bool = False,
 ):
-    name = codemod_kls.id
     wrapper = cst.MetadataWrapper(source_tree)
-    codemod = codemod_kls(
-        CodemodContext(wrapper=wrapper),
-        execution_context,
-        file_context,
-    )
+    codemod = codemod_kls(CodemodContext(wrapper=wrapper), file_context)
```
clavedeluna, line 40: maybe the diff is making it hard to read and `execution_context` is still getting passed but don't we need it in each codemod instance? Maybe its attrs aren't getting used right now but soon `repo_manager` will be, right?

clavedeluna (commented):
How much of a PITA would it be to add an integration test with --max-workers used?

src/codemodder/codemodder.py
drdavella: No, the execution context is no longer passed to individual codemods. I made this change because previously each codemod could potentially update `execution_context` which is not thread-safe. Anything we pass is going to need to go through `FileContext` now (which maybe needs a better name), so we can add `RepoManager` or just relevant fields like `python_versions` or `dependencies` or something.

src/codemodder/codemodder.py
drdavella: Nope, good catch.

andrecsilva (approved):

src/codemodder/codemodder.py
clavedeluna: I would love to combine classes at some point so we don't confuse FileContext and ExecutionContext

clavedeluna (approved):

drdavella:
@clavedeluna since the default is to use one worker and since this is purely to enable performance testing I'm going to skip the integration test for now. Once we decide we want to use `--max-workers > 1` in production environments we can revisit with an integration test.

PR #101, opened 2023-10-26
Flask set-cookie codemod

clavedeluna:
## Overview
*New codemod to check if flask's `set_cookie` is called with the most secure params*

## Description

* This codemod is merge after cursory review because it is possible to call `set_cookie` with default (seemingly unsafe) parameters but the safety is configured at the Flask app level configurations. However, whatever is passed to `set_cookie` takes precedence so calling it with unsafe params should be changed.
* Checking flask cookie config settings will be a separate codemod.

### DIFF
--- a/src/core_codemods/docs/pixee_python_secure-flask-cookie.md
+++ b/src/core_codemods/docs/pixee_python_secure-flask-cookie.md
@@ -1,5 +1,5 @@
 This codemod sets the most secure parameters when Flask applications call `set_cookie` on a response object. Without these parameters, your Flask
-application cookies may be vulnerable to being intercepted and used against your system.
+application cookies may be vulnerable to being intercepted and used to gain access to sensitive data.
 
 The changes from this codemod look like this:
 
--- a/src/core_codemods/docs/pixee_python_secure-flask-cookie.md
+++ b/src/core_codemods/docs/pixee_python_secure-flask-cookie.md
@@ -1,0 +1,15 @@
+This codemod sets the most secure parameters when Flask applications call `set_cookie` on a response object. Without these parameters, your Flask
+application cookies may be vulnerable to being intercepted and used against your system.
+
+The changes from this codemod look like this:
+
+```diff
+  from flask import Flask, session, make_response
+  app = Flask(__name__)
+  @app.route('/')
+    def index():
+      resp = make_response('Custom Cookie Set')
+    - resp.set_cookie('custom_cookie', 'value')
+    + resp.set_cookie('custom_cookie', 'value', secure=True, httponly=True, samesite='Lax')
+      return resp
+```
--- a/tests/codemods/test_secure_flask_cookie.py
+++ b/tests/codemods/test_secure_flask_cookie.py
@@ -30,89 +30,81 @@ class TestSecureFlaskCookie(BaseSemgrepCodemodTest):
         """
         self.run_and_assert(tmpdir, dedent(input_code), dedent(expexted_output))
 
+    @each_func
+    def test_from_import(self, tmpdir, func):
+        input_code = f"""\
+        from flask import {func}
 
-#     @each_func
-#     def test_from_import(self, tmpdir, func):
-#         input_code = f"""from lxml.etree import {func}
-#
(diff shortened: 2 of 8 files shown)


tests/codemods/test_secure_flask_cookie.py
```diff
@@ -0,0 +1,110 @@
...
+                "secure=True",
+                "secure=True, httponly=True, samesite='Lax'",
+            ),
+            (
+                "secure=True, httponly=True, samesite='Lax'",
+                "secure=True, httponly=True, samesite='Lax'",
+            ),
+            (
+                "secure=True, httponly=True, samesite='Strict'",
+                "secure=True, httponly=True, samesite='Lax'",
```
clavedeluna: We could consider allowing users to pass samesite=Strict and not changing it to Lax

andrecsilva (approved):
Lacking tests with `open(...)` as a source, otherwise ok.

clavedeluna:
> Lacking tests with `open(...)` as a source, otherwise ok.

Can you clarify with a code example?

andrecsilva:
> > Lacking tests with `open(...)` as a source, otherwise ok.
> 
> Can you clarify with a code example?

Your semgrep rule has 3 possibilities for source (`flask.make_response(...)`, `flask.Response(...)` and `open(...)`). See:
```
            pattern-sources:
              - pattern-either:
                  - patterns:
                    - pattern: flask.make_response(...)
                    - pattern-inside: |
                        import flask
                        ...
                  - patterns:
                    - pattern: flask.Response(...)
                    - pattern-inside: |
                        import flask
                        ...
                  - pattern: open(...)
```

All the unit tests are parameterized with this:
```python
each_func = pytest.mark.parametrize("func", ["make_response", "Response"])
```

Thus, no test covers an example like:
```python
import flask
response = open(something)
var = "hello"
response.set_cookie("name", "value")
```

I'm not sure if this is actually valid code. Just following the semgrep rule.

clavedeluna:
> > > Lacking tests with `open(...)` as a source, otherwise ok.
> > 
> > 
> > Can you clarify with a code example?
> 
> Your semgrep rule has 3 possibilities for source (`flask.make_response(...)`, `flask.Response(...)` and `open(...)`). See:
> 
> ```
>             pattern-sources:
>               - pattern-either:
>                   - patterns:
>                     - pattern: flask.make_response(...)
>                     - pattern-inside: |
>                         import flask
>                         ...
>                   - patterns:
>                     - pattern: flask.Response(...)
>                     - pattern-inside: |
>                         import flask
>                         ...
>                   - pattern: open(...)
> ```
> 
> All the unit tests are parameterized with this:
> 
> ```python
> each_func = pytest.mark.parametrize("func", ["make_response", "Response"])
> ```
> 
> Thus, no test covers an example like:
> 
> ```python
> import flask
> response = open(something)
> var = "hello"
> response.set_cookie("name", "value")
> ```
> 
> I'm not sure if this is actually valid code. Just following the semgrep rule.

OHHHH this is a total typo, from copy paste. open is not a valid source. I will remove it. Good catch!

tests/codemods/test_secure_flask_cookie.py
drdavella: Yes I think we should not modify `samesite` if it is already less permissive than our codemod.

drdavella (changes requested):
This looks really nice to me overall, but I do think we should preserve `samesite="Strict"` if it's already there. We don't want to make the app less permissive, only moreso.

src/core_codemods/docs/pixee_python_secure-flask-cookie.md
```diff
@@ -0,0 +1,15 @@
+This codemod sets the most secure parameters when Flask applications call `set_cookie` on a response object. Without these parameters, your Flask
+application cookies may be vulnerable to being intercepted and used against your system.
```
drdavella: Just a minor suggestion.
```suggestion
application cookies may be vulnerable to being intercepted and used to gain access to sensitive data.
```

drdavella (approved):

PR #102, opened 2023-10-26
Fix pragma error in RemoveUnusedImports

andrecsilva:
## Overview
Fixes errors with pragma parsing in RemoveUnusedImports. Also added a check to ignore `__init__.py` files.

### DIFF
--- a/src/core_codemods/remove_unused_imports.py
+++ b/src/core_codemods/remove_unused_imports.py
@@ -1,4 +1,4 @@
-from libcst import ensure_type, matchers
+from libcst import CSTVisitor, ensure_type, matchers
 from libcst.codemod.visitors import GatherUnusedImportsVisitor
 from libcst.metadata import (
     PositionProvider,
@@ -20,7 +20,7 @@ from libcst.codemod import Codemod, CodemodContext
 import re
 from pylint.utils.pragma_parser import parse_pragma
 
-NOQA_PATTERN = re.compile(r"^#\s*noqa")
+NOQA_PATTERN = re.compile(r"^#\s*noqa", re.IGNORECASE)
 
 
 class RemoveUnusedImports(BaseCodemod, Codemod):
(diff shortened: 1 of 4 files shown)


src/core_codemods/remove_unused_imports.py
```diff
@@ -45,6 +45,9 @@ def __init__(self, codemod_context: CodemodContext, *codemod_args):
         BaseCodemod.__init__(self, *codemod_args)
 
     def transform_module_impl(self, tree: cst.Module) -> cst.Module:
+        # Do nothing in __init__.py files
+        if self.file_context.file_path.name == "__init__.py":
```
clavedeluna, line 49: Is this getting tested by a unit test already?

clavedeluna (approved):

src/core_codemods/remove_unused_imports.py
andrecsilva: Good catch, I'll add one.

src/core_codemods/remove_unused_imports.py
```diff
@@ -111,25 +119,42 @@ def _is_disabled_by_linter(self, node: cst.CSTNode) -> bool:
...
 
 
 def _is_pylint_disable_next_unused_imports(comment: str) -> bool:
-    parsed = parse_pragma(comment)
-    for p in parsed:
-        if p.action == "disable-next" and (
-            "unused-import" in p.messages or "W0611" in p.messages
-        ):
-            return True
+    # If pragma parse fails, ignore
```
drdavella, line 150: Do we have a test for this case? It would be useful to add the case that caused the error in the first place.

src/core_codemods/remove_unused_imports.py
```diff
@@ -111,25 +119,42 @@ def _is_disabled_by_linter(self, node: cst.CSTNode) -> bool:
...
 
 
 def _is_pylint_disable_unused_imports(comment: str) -> bool:
-    parsed = parse_pragma(comment)
-    for p in parsed:
-        if p.action == "disable" and (
-            "unused-import" in p.messages or "W0611" in p.messages
-        ):
-            return True
+    # If pragma parse fails, ignore
```
drdavella, line 136: Do we have a test for this case?

drdavella (changes requested):
This looks good but I would expect to see tests that cover the `except` cases.

src/core_codemods/remove_unused_imports.py
andrecsilva: It was failing because of a `# NOQA` comment in one of the files in `pytest`. The thing is, any comment that is not recognized as a pragma by `pylint` will raise an exception (note that we use a function from `pylint` directly). We have no control of those, hence the catch-all `try`-`except`.

drdavella: We should probably honor `# noqa` as well for this codemod but that doesn't have to happen in this PR.

drdavella (approved):

src/core_codemods/remove_unused_imports.py
andrecsilva: We do (and have tests for it). I've also fixed an issue in this PR where the regular expression that detects the `# noqa` pragma was case sensitive.

PR #106, opened 2023-10-27
Alias detection solution for codemods

andrecsilva:
## Overview
Added new alias detection solution for a few codemods.

## Description
Added new alias detection solution for a few codemods (`harden-pyyaml`, `secure-tempfile`, and `upgrade-ssl-context-minimum-version`). This solution will detect whenever a module is aliased and prefer to use the alias for new references to module attributes. Those use a more general tool now included in `NameResolutionMixin`.

The `harden-ruamel` and `https-connection` codemods had  working solutions. I've enabled / added new tests for these.

Other codemods (`url-sandbox`, `use-defusedxml`) will replace module calls with another call from a drop-in replacement module. 
Those will ignore aliases and refer to the replacement module call directly. Since we are referring a third-party module on those, I've elected not to leave the current solution to make the change more explicit.

### DIFF
--- a/src/codemodder/codemods/utils_mixin.py
+++ b/src/codemodder/codemods/utils_mixin.py
@@ -86,19 +86,18 @@ class NameResolutionMixin(MetadataDependent):
                                 return (import_node, alias)
         return None
 
-    def get_aliased_prefix_name(self, node: cst.CSTNode, name: str):
+    def get_aliased_prefix_name(self, node: cst.CSTNode, name: str) -> Optional[str]:
         """
         Returns the alias of name if name is imported and used as a prefix for this node.
         """
         maybe_import = self.get_imported_prefix(node)
         maybe_name = None
-        if maybe_import:
-            imp, ia = maybe_import
-            match imp:
-                case cst.Import():
-                    imp_name = get_full_name_for_node(ia.name)
-                    if imp_name == name and ia.asname:
-                        maybe_name = ia.asname.name.value
+        if maybe_import and matchers.matches(maybe_import[0], matchers.Import()):
+            _, ia = maybe_import
+            imp_name = get_full_name_for_node(ia.name)
+            if imp_name == name and ia.asname:
+                # AsName is always a Name for ImportAlias
+                maybe_name = ia.asname.name.value
         return maybe_name
 
     def find_assignments(
--- a/src/codemodder/codemods/utils_mixin.py
+++ b/src/codemodder/codemods/utils_mixin.py
@@ -6,7 +6,6 @@ from libcst.metadata import (
     Assignment,
     BaseAssignment,
     ImportAssignment,
-    Scope,
     ScopeProvider,
 )
 from libcst.metadata.scope_provider import GlobalScope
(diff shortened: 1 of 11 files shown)


src/codemodder/codemods/utils_mixin.py
```diff
@@ -58,6 +63,44 @@ def _is_direct_call_from_imported_module(
...
+                        maybe_assignment, ImportAssignment
+                    ):
+                        import_node = maybe_assignment.node
+                        for alias in import_node.names:
+                            if maybe_assignment.name in (
+                                alias.evaluated_alias,
+                                alias.evaluated_name,
+                            ):
+                                return (import_node, alias)
+        return None
```
clavedeluna, line 87: surprised pylint nor mypy didn't complain but all returns of a func should return the same len, so this should be (None, None)

src/codemodder/codemods/utils_mixin.py
```diff
@@ -58,6 +63,44 @@ def _is_direct_call_from_imported_module(
...
+    def get_aliased_prefix_name(self, node: cst.CSTNode, name: str):
+        """
+        Returns the alias of name if name is imported and used as a prefix for this node.
+        """
+        maybe_import = self.get_imported_prefix(node)
+        maybe_name = None
+        if maybe_import:
+            imp, ia = maybe_import
+            match imp:
+                case cst.Import():
```
clavedeluna: I love match/case style but I Think in this particular case maybe an if statemtn or walrus is best since it's less code indentation.

src/core_codemods/harden_pyyaml.py
```diff
@@ -44,6 +47,13 @@ def rule(cls):
 
         """
 
-    def on_result_found(self, _, updated_node):
-        new_args = [*updated_node.args[:1], self.parse_expression("yaml.SafeLoader")]
+    def on_result_found(self, original_node, updated_node):
+        maybe_name = self.get_aliased_prefix_name(original_node, self._module_name)
+        maybe_name = maybe_name or self._module_name
+        if maybe_name and maybe_name == self._module_name:
```
clavedeluna: `if maybe_name` not necessary since above you do `or` which will make `maybe_name` be truthy

src/core_codemods/tempfile_mktemp.py
```diff
@@ -26,6 +29,9 @@ def rule(cls):
         """
 
     def on_result_found(self, original_node, updated_node):
+        maybe_name = self.get_aliased_prefix_name(original_node, self._module_name)
+        maybe_name = maybe_name or self._module_name
+        if maybe_name and maybe_name == self._module_name:
```
clavedeluna: same as above on the if check

src/core_codemods/upgrade_sslcontext_minimum_version.py
```diff
@@ -45,6 +48,11 @@ def rule(cls):
         """
 
     def on_result_found(self, original_node, updated_node):
+        maybe_name = self.get_aliased_prefix_name(
+            original_node.value, self._module_name
+        )
+        maybe_name = maybe_name or self._module_name
+        if maybe_name and maybe_name == self._module_name:
+            self.add_needed_import(self._module_name)
```
clavedeluna, line 56: these new green lines added are the same (or almost?) in every codemod added so maybe make this a separate method to call to dedupe

clavedeluna (changes requested):
Can you address in your PR description why other codemods didn't get this lovely treatment? Grepping `alias` in tests I'm seeing some that could be easy to add: ruamel, defusedxml, https connection, and url sandbox had some aliasing logic too. If you can address what's the plan with those or add them in this PR that would be helpful context.

src/codemodder/codemods/utils_mixin.py
andrecsilva: The return type is annotated as `Optional[Tuple[...]]`, that's the reason.
Returning `(None, None)` would mess with `None` checks since `bool((None,None)) == True`.

src/codemodder/codemods/utils_mixin.py
andrecsilva: You're right here, but using match/case has the advantage of signaling the static analysis tools about the type.

src/core_codemods/upgrade_sslcontext_minimum_version.py
andrecsilva: These address a very specific problem that are shared by those codemods. I don't want to add it to the api because the solution needs `NameResolutionMixin` which depends on `ScopeProvider`.

Metadata is calculated at the start of every transform on a as-need basis (libcst looks at `METADATA_DEPENDENCIES` for that). Integrating it to the API means every codemod would calculate the `ScopeProvider` metadata, making it a bit more expensive.

Maybe creating a class for those types of codemods would be a good idea, but I'd rather do that when we try to rewrite those in pure libcst.

tests/codemods/test_tempfile_mktemp.py
```diff
@@ -1,4 +1,4 @@
-import pytest
+import pytest  # pylint: disable=unused-import
```
drdavella: Just needs to be deleted, not disabled.

src/core_codemods/upgrade_sslcontext_minimum_version.py
drdavella: I agree the duplication here is not ideal. I think we need to revisit the whole concept of utility mixins for this reason, although I understand the reasons it was done this way.

drdavella (approved):
LGTM but please delete the unused `pytest` import.

clavedeluna (approved):

PR #107, opened 2023-10-27
Only run semgrep for codemods with initial results

drdavella:
## Overview
*Only run semgrep for codemods with initial results*

## Description

* This is an optimization that is intended to reduce runtime for the case where many/all codemods are requested
* The idea is to run semgrep once up-front to gather a list of applicable codemods and then to only run semgrep for those codemods that were identified in the initial pass
* There are further optimizations available here but this is a simple change intended to improve performance for this particular case
* Unfortunately this comes at the expense of doubling the semgrep cost of running a single codemod. We will address this later as it is not currently the bottleneck in performance.

### DIFF
--- a/src/codemodder/codemodder.py
+++ b/src/codemodder/codemodder.py
@@ -1,6 +1,7 @@
 from concurrent.futures import ThreadPoolExecutor
 import datetime
 import difflib
+import itertools
 import logging
 import os
 import sys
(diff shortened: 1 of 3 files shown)


andrecsilva (approved):
I'm very skeptical this will have any actual impact on performance, otherwise it looks fine.

drdavella:
@andrecsilva I gathered performance data that indicates that our runtime is completely dominated by invoking semgrep in certain environments. Anything we do to reduce the number of times we invoke semgrep is going to have a huge impact.

PR #110, opened 2023-10-27
Use setuptools_scm to implement tag-based versioning

drdavella:
## Overview
*Use `setuptools_scm` to implement tag-based versioning*

## Description
* This is the first step towards supporting tag-based releases (to PyPI)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Install Dependencies
         run: pip install -r requirements/lint.txt
       - name: Black Format Check
-        run: black --check . --exclude samples/
+        run: black --check .
       - name: Run pylint
         run: make lint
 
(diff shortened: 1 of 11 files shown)


andrecsilva (approved):

.github/workflows/lint.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Install Dependencies
         run: pip install -r requirements/lint.txt
       - name: Black Format Check
-        run: black --check . --exclude samples/
+        run: black --check .
```
clavedeluna, line 34: are we sure this is what we want? Changing the format of our samples can have some unintended effects on some codemods.

drdavella: This is now excluded in `pyproject.toml` since that affects all environments and not just GitHub actions.

PR #115, opened 2023-11-02
Filter semgrep results based on initial findings

drdavella:
## Overview
*Only run semgrep on files that have findings from initial run*

## Description

* This is an optimization designed to make semgrep faster by limiting the number of files it analyzes
  * This is motivated by the fact that semgrep execution is our long pole in certain environments
* Only files that have findings from the initial "discovery" run will be considered for further analysis
* In order to enable this change I made some changes to the structure of processed sarif results

### DIFF
--- a/src/codemodder/codemodder.py
+++ b/src/codemodder/codemodder.py
@@ -20,6 +20,7 @@ from codemodder.context import CodemodExecutionContext
 from codemodder.executor import CodemodExecutorWrapper
 from codemodder.project_analysis.python_repo_manager import PythonRepoManager
 from codemodder.report.codetf_reporter import report_default
+from codemodder.result import ResultSet
 from codemodder.semgrep import run as run_semgrep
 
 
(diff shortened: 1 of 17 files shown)


src/codemodder/codemodder.py
```diff
@@ -102,14 +102,17 @@ def process_file(
 
     line_exclude = file_line_patterns(file_path, cli_args.path_exclude)
     line_include = file_line_patterns(file_path, cli_args.path_include)
-    sarif_for_file = sarif.get(str(file_path)) or {}
+    findings_for_rule = results.results_for_rule_and_file(
```
andrecsilva, line 115: Before, we could query the results for a file x rule in O(1) amortized due to using a dict of a dict to gather the results per rule and file. Now we have to iterate over all possible results for given rule and filter those by file path.

Since this is something we do a lot (once per file) and we expect no results per file, I suggest we organize the results by file path in a dict, as before.

andrecsilva (changes requested):

andrecsilva (approved):

drdavella:
I want to merge #116 and release a bugfix before merging this one.

PR #119, opened 2023-11-08
Flip insecure defined flask session config

clavedeluna:
## Overview
*A codemod that will check if `SESSION_COOKIE_HTTPONLY`, `SESSION_COOKIE_SECURE` or `SESSION_COOKIE_SAMESITE` are set to an insecure value on a flask app config object.*

## Description

* This codemod checks any instances of a Flask app for any calls to its configuration. If any of those calls leads to setting any of the 3 flags to their unsafe value, this codemod will change the value.
* Note that after quite a bit of back and forth, we've decided this codemod will not at this time add set the secure values for any of these configs. We will only handle the case when the user specifically set to an unsafe value. We're aware that some default values are unsafe. 
* I've left some commented out code which did handle ^ above by adding a line at the very end of the module any of the config values which were not set. We decided there is some risk to adding this line at the end of the module so we won't do so right now.

### DIFF
--- a/src/codemodder/project_analysis/file_parsers/setup_py_file_parser.py
+++ b/src/codemodder/project_analysis/file_parsers/setup_py_file_parser.py
@@ -1,11 +1,11 @@
 from codemodder.project_analysis.file_parsers.package_store import PackageStore
+from codemodder.utils.utils import clean_simplestring
 from pathlib import Path
 import libcst as cst
 from libcst import matchers
 from packaging.requirements import Requirement
 
 from .base_parser import BaseParser
-from .utils import clean_simplestring
 
 
 class SetupPyParser(BaseParser):
--- a/src/core_codemods/secure_flask_session_config.py
+++ b/src/core_codemods/secure_flask_session_config.py
@@ -1,0 +1,96 @@
+import libcst as cst
+from libcst.codemod import Codemod, CodemodContext, ContextAwareVisitor
+from libcst.metadata import ParentNodeProvider, ScopeProvider
+
+from libcst import matchers
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import BaseCodemod
+from codemodder.codemods.utils_mixin import NameResolutionMixin
+
+
+class SecureFlaskSessionConfig(BaseCodemod):
+    METADATA_DEPENDENCIES = BaseCodemod.METADATA_DEPENDENCIES + (
+        ParentNodeProvider,
+        ScopeProvider,
+    )
+    NAME = "secure-flask-session-configuration"
+    SUMMARY = "UTODO"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_REVIEW
+    DESCRIPTION = "TODO"
+    REFERENCES = [
+        {
+            "url": "todo",
+            "description": "",
+        }
+    ]
+
+    def transform_module_impl(self, tree: cst.Module) -> cst.Module:
+        flask_visitor = FindConfigCalls(self.context)
+        tree.visit(flask_visitor)
+        if not flask_visitor.flask_app_name:
+            return tree
+
+        # TODO: iterate through flask_visitor.config_access
+        # TODO: use a list of secure config names, values to only write the needed ones out
+        result = self.insert_config_line_endof_mod(tree, flask_visitor.flask_app_name)
+        return result
+
+    def insert_config_line_endof_mod(
+        self, original_node: cst.Module, app_name: str
+    ) -> cst.Module:
+        # TODO: record change
+        # line_number is the end of the module where we will insert the new flag.
+        # pos_to_match = self.node_position(original_node)
+        # line_number = pos_to_match.end.line
+        # self.changes_in_file.append(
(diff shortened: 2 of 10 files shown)


src/core_codemods/secure_flask_session_config.py
```diff
@@ -0,0 +1,186 @@
...
+        self.flask_app_name = ""
+        # Later: if we want to store configs to write later
+        # self.configs_to_write = self.SECURE_SESSION_CONFIGS.copy()
+        self.file_context = file_context
+
+    def _store_flask_app(self, original_node) -> None:
+        flask_app_parent = self.get_metadata(ParentNodeProvider, original_node)
+        match flask_app_parent:
+            case cst.AnnAssign() | cst.Assign():
+                flask_app_attr = flask_app_parent.targets[0].target
```
andrecsilva: This won't work with `AnnAssign`, as the target value is stored in the `target` attribute. You should separate the cases.

This may help:
```python
    def _extract_targets_of_assignment(
        self, assignment: cst.AnnAssign | cst.Assign | cst.WithItem | cst.NamedExpr
    ) -> list[cst.BaseExpression]:
        match assignment:
            case cst.AnnAssign():
                if assignment.target:
                    return [assignment.target]
            case cst.Assign():
                return [t.target for t in assignment.targets]
            case cst.NamedExpr():
                return [assignment.target]
            case cst.WithItem():
                if assignment.asname:
                    return [assignment.asname.name]
        return []
```

src/core_codemods/secure_flask_session_config.py
```diff
@@ -0,0 +1,186 @@
...
+                self.flask_app_name = flask_app_attr.value
```
andrecsilva: This assumes that the target is a `Name`, but the target may be any slew of different types of expressions like `Subscript`:(`l[0] = Flask(...)`) , `Attribute`:(`a.b = Flask(...)`), etc. So you should make it fail in the other cases.

My suggestion is to break this detection part into a new method `is_assigned_to_variable` with return `Optional[Name]`.

src/core_codemods/secure_flask_session_config.py
```diff
@@ -0,0 +1,186 @@
...
+        app_config_node = cst.Attribute(value=app_name, attr=config)
+        update = cst.Name(value="update")
+        return matchers.matches(
+            original_node.func, matchers.Attribute(value=app_config_node, attr=update)
+        )
+
+    def _is_config_subscript(self, original_node: cst.Assign):
+        config = cst.Name(value="config")
+        app_name = cst.Name(value=self.flask_app_name)
+        app_config_node = cst.Attribute(value=app_name, attr=config)
```
andrecsilva: When using `matchers.matches`, you should use the `matchers` version of the nodes instead (i.e. `matchers.Name` instead of `cst.Name`).

See: https://libcst.readthedocs.io/en/latest/matchers.html#libcst.matchers.matches

src/core_codemods/secure_flask_session_config.py
```diff
@@ -0,0 +1,186 @@
...
+                safe_value = self._get_secure_config_val(key)
+                self.report_change(original_node)
+                return updated_node.with_changes(value=safe_value)
+        return updated_node
+
+    def _is_config_update_call(self, original_node: cst.Call):
+        config = cst.Name(value="config")
+        app_name = cst.Name(value=self.flask_app_name)
+        app_config_node = cst.Attribute(value=app_name, attr=config)
+        update = cst.Name(value="update")
```
andrecsilva: See comment in `_is_config_subscript`. Same issue.

andrecsilva (changes requested):
Some minor issues, otherwise it looks good.

src/codemodder/project_analysis/file_parsers/setup_py_file_parser.py
drdavella, line 1: I think this file doesn't belong in this PR, it looks like it's from another ticket.

drdavella (approved):
Thanks for all of the back and forth on this, I know it ended up being fairly tricky. 

Based on the expected behavior in the tests this looks good to me. My only substantive comment is to remove the file that doesn't belong in this PR. Once @andrecsilva's other comments are addressed this should be good to go.

src/codemodder/project_analysis/file_parsers/setup_py_file_parser.py
clavedeluna: This change does belong here, I didn't add the file, I just moved a function to utils so I had to change the import

src/core_codemods/secure_flask_session_config.py
clavedeluna: I assume you wrote that function and it isn't from somewhere I should cite, so I'm taking it as is.

src/codemodder/project_analysis/file_parsers/setup_py_file_parser.py
drdavella: Ah sorry my bad.

src/core_codemods/secure_flask_session_config.py
andrecsilva: I wrote it for the file resource leak codemod. This PR will probably be merged before, so go ahead.

andrecsilva (approved):

PR #123, opened 2023-11-09
convert django debug codemod to use codemod api

clavedeluna:
## Overview
Convert one codemod to use codemod API instead of basecodemod directly. I recognized the opportunity to make this codemod much smaller. Instead of having one codemod class and one transformer, I made it all into one small class. Since this codemod only cares about settings.py files, we can use the fact that `visit_Module` checks if we should even visit the module. If so, this codemod will run at `leave_Assign`

If we like this change, I think I can make the same for django cookie secure off codemod


andrecsilva (approved):

clavedeluna:
@andrecsilva sorry for asking for review again. I recognized an optimization using visit_Module

andrecsilva (approved):
Good call on the optimization.  Maybe we could use a `should_visit_module()` method in the API to make it a bit more clear and independent on libcst.

PR #124, opened 2023-11-09
Support for f-strings in SQLQueryParameterization

andrecsilva:
## Overview
Added support for query parameterization contained in f-strings.

## Description

Aside from the main change, the following are of note:
- Added a new utility: `BaseType` to represent literal types in python. It comes along with a method to infer an expression type from a few common cases
- Changed how multiple expressions in a single injection are parameterized. The reason is to avoid problems with type conversions. For example, given the following:
```
cursor.execute("SELECT * FROM users where name ='" + name + '_and_' + lastname + "'")
```
will now produce:
```
cursor.execure("SELECT * FROM users where name =?", ('{0}_and_{1}'.format(name, lastname), ))
```
while before it would result in:
```
cursor.execure("SELECT * FROM users where name =?", (name + '_and_' + lastname, ))
```

The `.format` method was chosen for compatibility. While f-strings would look better, they require python 3.6+.

### DIFF
--- a/src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
+++ b/src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
@@ -1,11 +1,28 @@
+from typing import Union
 import libcst as cst
-from libcst import CSTTransformer
+from libcst import CSTTransformer, RemovalSentinel, SimpleString
 
 
 class RemoveEmptyStringConcatenation(CSTTransformer):
     """
     Removes concatenation with empty strings (e.g. "hello " + "") or "hello" ""
     """
+
+    def leave_FormattedStringExpression(
+        self,
+        original_node: cst.FormattedStringExpression,
+        updated_node: cst.FormattedStringExpression,
+    ) -> Union[
+        cst.BaseFormattedStringContent,
+        cst.FlattenSentinel[cst.BaseFormattedStringContent],
+        RemovalSentinel,
+    ]:
+        expr = original_node.expression
+        match expr:
+            case SimpleString():  # type: ignore
+                if expr.raw_value == "":
+                    return RemovalSentinel.REMOVE
+        return updated_node
 
     def leave_BinaryOperation(
         self, original_node: cst.BinaryOperation, updated_node: cst.BinaryOperation
--- a/src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
+++ b/src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
@@ -19,9 +19,8 @@ class RemoveEmptyStringConcatenation(CSTTransformer):
     ]:
         expr = original_node.expression
         match expr:
-            case SimpleString():  # type: ignore
-                if expr.raw_value == "":
-                    return RemovalSentinel.REMOVE
+            case SimpleString() if expr.raw_value == "":  # type: ignore
+                return RemovalSentinel.REMOVE
         return updated_node
 
     def leave_BinaryOperation(
--- a/src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
+++ b/src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
@@ -39,19 +39,22 @@ class RemoveEmptyStringConcatenation(CSTTransformer):
     def handle_node(
         self, updated_node: cst.BinaryOperation | cst.ConcatenatedString
     ) -> cst.BaseExpression:
-        match updated_node.left:
-            # TODO f-string cases
-            case cst.SimpleString() if updated_node.left.raw_value == "":
-                match updated_node.right:
-                    case cst.SimpleString() if updated_node.right.raw_value == "":
-                        return cst.SimpleString(value='""')
-                    case _:
-                        return updated_node.right
-        match updated_node.right:
-            case cst.SimpleString() if updated_node.right.raw_value == "":
-                match updated_node.left:
-                    case cst.SimpleString() if updated_node.left.raw_value == "":
-                        return cst.SimpleString(value='""')
-                    case _:
-                        return updated_node.left
+        left = updated_node.left
+        right = updated_node.right
+        if self._is_empty_string_literal(left):
+            if self._is_empty_string_literal(right):
+                return cst.SimpleString(value='""')
+            return right
+        if self._is_empty_string_literal(right):
+            if self._is_empty_string_literal(left):
+                return cst.SimpleString(value='""')
+            return left
         return updated_node
+
+    def _is_empty_string_literal(self, node):
+        match node:
+            case cst.SimpleString() if node.raw_value == "":
+                return True
+            case cst.FormattedString() if not node.parts:
+                return True
+        return False
--- a/src/codemodder/codemods/utils.py
+++ b/src/codemodder/codemods/utils.py
@@ -16,42 +16,40 @@ class BaseType(Enum):
     STRING = 3
     BYTES = 4
 
-    @classmethod
-    # pylint: disable-next=R0911
-    def infer_expression_type(cls, node: cst.BaseExpression) -> Optional["BaseType"]:
-        """
-        Tries to infer if the type of a given expression is one of the base literal types.
-        """
-        # The current implementation could be enhanced with a few more cases
-        match node:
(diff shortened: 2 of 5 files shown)


src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
```diff
@@ -1,12 +1,29 @@
...
+        original_node: cst.FormattedStringExpression,
+        updated_node: cst.FormattedStringExpression,
+    ) -> Union[
+        cst.BaseFormattedStringContent,
+        cst.FlattenSentinel[cst.BaseFormattedStringContent],
+        RemovalSentinel,
+    ]:
+        expr = original_node.expression
+        match expr:
+            case SimpleString():  # type: ignore
```
clavedeluna: is it possible to do `case Simplestring() and expr.raw_value == ...` in one line? This is one case when the multiple indent is kinda exsessive and just having an if statement seems cleaner than using match.

src/core_codemods/sql_parameterization.py
```diff
@@ -428,3 +504,18 @@ def leave_Call(self, original_node: cst.Call) -> None:
                                 expr.value
                             ):
                                 self.calls[original_node] = query_visitor.leaves
+
+
+def _extract_prefix_raw_value(self, node) -> Optional[Tuple[str, str]]:
```
clavedeluna, line 510: kinda similar to `clean_simplestring` which is in another util module. At some point we should try to combine

src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
andrecsilva: Good call on this. I've been writing those type checks on autopilot, there is quite a few of them for this codemod.

src/codemodder/codemods/utils.py
```diff
@@ -1,10 +1,59 @@
...
+    """
+
+    NUMBER = 1
+    LIST = 2
+    STRING = 3
+    BYTES = 4
+
+    @classmethod
+    # pylint: disable-next=R0911
+    def infer_expression_type(cls, node: cst.BaseExpression) -> Optional["BaseType"]:
```
drdavella: I think it might be slightly more idiomatic to do the following:
```python
from typing_extensions import Self


def .... -> Optional[Self]
```
I think the `Self` type is included in newer versions of the built-in `typing` module.

src/core_codemods/sql_parameterization.py
```diff
@@ -128,52 +153,117 @@ def transform_module_impl(self, tree: cst.Module) -> cst.Module:
...
-                    current_end.prefix
-                    + current_end.quote
-                    + new_raw_value
-                    + current_end.quote
-                )
-                self.changed_nodes[end] = current_end.with_changes(value=new_value)
-            case cst.FormattedStringText():
-                # TODO formatted string case
-                pass
+            if isinstance(
```
drdavella, line 184: Just a nitpick but feels like this would read slightly cleaner with a match?

src/codemodder/codemods/utils.py
drdavella: Also I don't have a major objection but this pattern of using a class method feels a bit Java-ish to me. I think it probably just makes more sense as a module-level utility method.

src/core_codemods/sql_parameterization.py
```diff
@@ -64,25 +74,42 @@ def __init__(
...
-            right=middle[index],
+        format_pieces: list[str] = []
+        format_expr_count = 0
+        args = []
+        if len(new_middle) == 1:
+            # TODO maybe handle conversion here?
+            return new_middle[0]
+        for e in new_middle:
+            exception = False
+            if isinstance(e, cst.SimpleString | cst.FormattedStringText):
```
drdavella, line 103: `match`?

src/core_codemods/sql_parameterization.py
```diff
@@ -128,52 +153,117 @@ def transform_module_impl(self, tree: cst.Module) -> cst.Module:
...
+        t = _extract_prefix_raw_value(self, current_end)
+        prefix, raw_value = t if t else ("", "")
+        if "r" in prefix:
+            quote_span = list(raw_quote_pattern.finditer(raw_value))[0]
+        else:
+            quote_span = list(quote_pattern.finditer(raw_value))[0]
+
+        new_raw_value = raw_value[quote_span.end() :]
+        append_raw_value = raw_value[: quote_span.start()]
+        match current_end:
```
drdavella: This seems almost exactly the same as the block starting with `match current_start:` above. I might suggest factoring this out into a separate function for readability since this method is already quite long.

src/core_codemods/sql_parameterization.py
```diff
@@ -309,88 +399,74 @@ def leave_Module(self, original_node: cst.Module):
...
-                    matches = list(self.raw_quote_pattern.finditer(node.raw_value))
-                else:
-                    matches = list(self.quote_pattern.finditer(node.raw_value))
-                return matches[0].start() == 0 if matches else False
-            case cst.FormattedStringText():
-                # TODO may be in the middle i.e. f"'{exp}_home'"
-                # be careful of f"name='literal'", it needs one but not two
-                return False
-        return False
+        t = _extract_prefix_raw_value(self, node)
```
drdavella, line 439: Other than the final condition, this method exactly duplicates the body of `_is_literal_start`, which suggests it should be factored out.

drdavella (approved):
LGTM overall, thanks for all of the hard work.

I do think there is some opportunity here to reduce duplication and improve readability, which would be nice to do before merging.

My longer term concern is that this is pretty complicated code, which makes it fairly difficult to understand as a reviewer/maintainer. It is also going to continue to grow as we support additional string processing and probably other SQL drivers. 

Right now it feels like there is not a clear separation of concerns between the code that processes string literals and the core codemod itself. I think going forward it would be good to break these things into separate components where possible before this becomes too unwieldy.

src/core_codemods/sql_parameterization.py
andrecsilva: `match` would create an extra indent here with no additional benefit.

src/core_codemods/sql_parameterization.py
andrecsilva: Same as the previous comment, `match` would create an extra indent with no benefit. As a personal rule, I tend to use match to capture complex patterns or enforce types in `case` body. Mostly to appease the static analysis. 

I'm not against to just use `match` in those cases. Maybe we could talk about adopting an informal coding standard here.

src/core_codemods/sql_parameterization.py
andrecsilva: There is not much benefit in doing that. To elaborate, let's ponder what would be the return of the newly factored out method.

The logical, straightforward answer would be `Optional[Tuple[...]]` since we extract the `prefix` and `raw_value` and check for failing conditions (i.e. the `b` prefix). But since this is also a may-fail method, we would need to add an extra check for this which would result in a similar duplicated code.

This type of chaining of may-fail methods are prevalent here and the overall code could be cleaned up quite a using maybe monad as a programming pattern. As much as I'd like do use it, I've been avoiding it (at least explicitly) since would cost readability quite a bit as monads are seemingly infamously hard to grasp.

andrecsilva:
> LGTM overall, thanks for all of the hard work.
> 
> I do think there is some opportunity here to reduce duplication and improve readability, which would be nice to do before merging.
> 
> My longer term concern is that this is pretty complicated code, which makes it fairly difficult to understand as a reviewer/maintainer. It is also going to continue to grow as we support additional string processing and probably other SQL drivers.
> 
> Right now it feels like there is not a clear separation of concerns between the code that processes string literals and the core codemod itself. I think going forward it would be good to break these things into separate components where possible before this becomes too unwieldy.

I've added a comment explaining the major steps of the transformations. For now I'm expecting the code to change quite a bit when we add support for other string format options, so I'm avoiding factoring out some of the string processing methods.

src/core_codemods/sql_parameterization.py
drdavella: That sounds good to me. I'm a little unclear on where `match` is beneficial myself. Although I do agree in certain places it appeases the type checker and in other places it enables destructuring which is quite useful.

PR #125, opened 2023-11-09
Django session codemod to use codemod API

clavedeluna:
## Overview
*Rely on codemod API*

Cleaner django session codemod and refactor


src/codemodder/codemods/utils.py
```diff
@@ -91,3 +91,10 @@ def get_function_name_node(call: cst.Call) -> Optional[cst.Name]:
         case cst.Attribute():
             return call.func.attr
     return None
+
+
+def is_assigned_to_True(original_node: cst.Assign):
```
andrecsilva, line 137: To make this more general, it should cover the `AnnAssign` (assign with annotations, e.g. `x:int=1`) case. Just note that `AnnAssign` may not always have a non-None `target` attribute.

andrecsilva (approved):

PR #131, opened 2023-11-14
Generate CodeTF when no files/codemods match

drdavella:
## Overview
*Generate CodeTF with empty results when no files or codemods are applied*

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -5,7 +5,7 @@ coverage:
         target: 90%
     patch:
       default:
-        target: 100%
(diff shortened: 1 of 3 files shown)


codecov.yaml
```diff
@@ -5,7 +5,7 @@ coverage:
         target: 90%
     patch:
       default:
-        target: 100%
+        target: 90%
```
andrecsilva, line 8: The most exciting change in this PR.

andrecsilva (approved):

PR #135, opened 2023-11-15
New codemod: use-generator

drdavella:
## Overview
*Implement new codemod: `use-generator`*

## Description

* This was inspired by a [pylint rule](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/use-a-generator.html)
* It is implemented purely in terms of `libcst` constructs
* I would like to add some kind of scope check that ensures the functions being checked are actually builtin

### DIFF
--- a/src/codemodder/codemods/utils_mixin.py
+++ b/src/codemodder/codemods/utils_mixin.py
@@ -5,6 +5,7 @@ from libcst.helpers import get_full_name_for_node
 from libcst.metadata import (
     Assignment,
     BaseAssignment,
+    BuiltinAssignment,
     ImportAssignment,
     ScopeProvider,
 )
@@ -148,6 +149,15 @@ class NameResolutionMixin(MetadataDependent):
         if len(assignments) == 1:
             return next(iter(assignments))
         return None
+
+    def is_builtin_function(self, node: cst.Call):
+        """
+        Given a `Call` node, find if it refers to a builtin function
+        """
+        maybe_assignment = self.find_single_assignment(node)
+        if maybe_assignment and isinstance(maybe_assignment, BuiltinAssignment):
+            return matchers.matches(node.func, matchers.Name())
+        return False
 
 
 def iterate_left_expressions(node: cst.BaseExpression):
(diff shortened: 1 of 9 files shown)


src/core_codemods/docs/pixee_python_use-generator.md
```diff
@@ -0,0 +1,13 @@
+Imagine that someone handed you a pile of 100 apples and then asked you to count how many of them were green without putting any of them down. You'd probably find this quite challenging and you'd struggle to hold the pile of apples at all. Now imagine someone handed you the apples one at a time and asked you to just count the green ones. This would be a much easier task.
```
andrecsilva, line 1: Mostly a nitpick, but this is the kind of language I'd avoid in documentations. Be direct and on point.

src/core_codemods/use_generator.py
```diff
@@ -0,0 +1,49 @@
...
+            "url": "https://docs.python.org/3/glossary.html#term-list-comprehension",
+            "description": "",
+        },
+    ]
+
+    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
+        match original_node.func:
+            # NOTE: could also support things like `list` and `tuple`
+            # but it's a less compelling use case
+            case cst.Name("any" | "all" | "sum" | "min" | "max"):
```
andrecsilva, line 31: As you have pointed it out, those could use an extra step to check if they are builtin functions. You can do this using the `ScopeProvider`. Query the metadata for assignments for this `Name` node and check if it is a `BuiltinAssignment`.

andrecsilva (changes requested):
Looks pretty good, just needs the builtin check.

src/core_codemods/docs/pixee_python_use-generator.md
drdavella: Maybe this is a little too whimsical but it's important to remember that a big part of our vision here involves educating developers and security professionals on best practices. This means we actually need to make our content as engaging as possible. I'm not necessarily saying I've succeeded in that here, but I'm going to leave it as-is and see what feedback we receive.

src/codemodder/codemods/utils_mixin.py
```diff
@@ -149,6 +150,15 @@ def find_single_assignment(
             return next(iter(assignments))
         return None
 
+    def is_builtin_function(self, node: cst.Call):
+        """
+        Given a `Call` node, find if it refers to a builtin function
+        """
+        maybe_assignment = self.find_single_assignment(node)
+        if maybe_assignment and isinstance(maybe_assignment, BuiltinAssignment):
+            return matchers.matches(node.func, matchers.Name())
```
andrecsilva, line 159: I'm 90% sure this is redundant, as in any function that that matches the `if` predicate will have `node.func` as `Name()`.

andrecsilva (approved):

src/codemodder/codemods/utils_mixin.py
drdavella: That's fair. I'll leave it for now but we could revisit.

PR #141, opened 2023-11-17
Updates to fix-mutable-params codemod

drdavella:
## Overview
*Updates to `fix-mutable-params`*

## Description

* Do not modify type annotations that are already `Optional`
* Do not modify the body of `abstractmethod`s


src/core_codemods/fix_mutable_params.py
```diff
@@ -133,6 +140,14 @@ def _build_new_body(self, new_var_decls, body):
         new_body.extend(body[offset:])
         return new_body
 
+    def _is_abstractmethod(self, node: cst.FunctionDef) -> bool:
+        for decorator in node.decorators:
+            match decorator.decorator:
+                case cst.Name("abstractmethod"):
```
drdavella, line 146: Could possibly be more accurate and guarantee this is imported from `abc`, but it's probably pretty reliable as-is.

andrecsilva (approved):

PR #143, opened 2023-11-20
Add support for aiohttp_jinja2 to enable-jinja2-autoescape

drdavella:
## Overview
*Add support for `aiohttp_jinja2` to `enable-jinja2-autoescape` codemod*

## Description
* I discovered an example of this in an open source app and it fits nicely into the existing codemod


andrecsilva (approved):

clavedeluna (approved):
Could make the aiohttp tests a separate class but 🤷

PR #144, opened 2023-11-21
Add requirements to setup.py codemod

clavedeluna:
## Overview
*Add a codemod that updates a `setup.py` module to add arbitrary requirements to the `install_requires` argument of `setup` call*

## Description

* This codemod is in isolation but the goal is to use it for dependency management. When codemod A needs to add a dependency, if the project uses setup.py, we would use this codemod
* Note right now this is limited to **adding** dependencies, not to removing or merging them.


src/codemodder/codemods/utils.py
```diff
@@ -113,6 +113,10 @@ def is_django_settings_file(file_path: Path):
     return False
 
 
+def is_setup_py_file(file_path: Path):
```
drdavella, line 116: I think this is probably pretty specific to the setup.py codemod and maybe doesn't belong here.

tests/dependency_management/test_setup_py_codemod.py
```diff
@@ -0,0 +1,218 @@
...
+            packages=find_packages("src"),
+            package_dir={"": "src"},
+            python_requires=">3.6",
+            install_requires="some-package",
+        )
+        """
+        tmp_file_path = Path(tmpdir / "setup.py")
+        self.run_and_assert_filepath(tmpdir, tmp_file_path, before, before)
+
+    @pytest.mark.skip("Need to add support.")
```
drdavella, line 174: Obviously not for this ticket but I think this will definitely be possible for the simple cases.

src/codemodder/dependency_management/setup_py_codemod.py
```diff
@@ -0,0 +1,68 @@
...
+
+class SetupPyAddDependencies(BaseCodemod, NameResolutionMixin):
+    NAME = "setup-py-add-dependencies"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    SUMMARY = "Add Dependencies to `setup.py` `install_requires`"
+    DESCRIPTION = SUMMARY
+    REFERENCES: list = []
+
+    def __init__(
+        self, codemod_context: CodemodContext, file_context: FileContext, dependencies
```
drdavella: Could we add a type annotation for `dependencies`?

src/codemodder/dependency_management/setup_py_codemod.py
```diff
@@ -0,0 +1,68 @@
...
+            new_args.append(new)
+        return new_args
+
+    def add_dependencies_to_arg(self, arg: cst.Arg) -> cst.Arg:
+        new_dependencies = [
+            cst.Element(value=cst.SimpleString(value=f'"{str(dep)}"'))
+            for dep in self.dependencies
+        ]
+        # TODO: detect if elements are separated by newline in source code.
+        return cst.Arg(
```
drdavella, line 62: I think we need to call `self.add_change` here (or maybe in the calling method) since the caller of this codemod will need to build a `ChangeSet` to report in CodeTF.

drdavella (approved):
A few minor comments but looks really good overall.

src/codemodder/codemods/utils.py
clavedeluna: In an effort to generalize and since I don't quite yet know how the repo manager will work, I think it's worth keeping it here for now.

src/codemodder/dependency_management/setup_py_codemod.py
clavedeluna: The purpose of this PR is to show we can take a setup.py file and update install_requires, without much else. This change will come in an upcoming PR when we merge with codemodder / dependency manager.

PR #147, opened 2023-11-22
File Resource Leak Codemod

andrecsilva:
## Overview

Adds a new codemod that closes file resources requested with `open`.

## Description

- Added a new codemod: `FileResourceLeak`.
- Added  `AncestorPatternsMixin` a new mixin library for querying patterns involving ancestors of a node.


src/core_codemods/docs/pixee_python_file-resource-leak.md
```diff
@@ -0,0 +1,12 @@
+This codemod wraps assignments of `open` calls in a with statement. Without explicit closing, these resources will be "leaked", and won't be re-claimed until garbage collection. In situations where these resources are leaked rapidly (either through malicious repetitive action or unusually spiky usage), connection pool or file handle exhaustion will occur. These types of failures tend to be catastrophic, resulting in downtime and many times affect downstream applications.
```
clavedeluna:
```suggestion
This codemod wraps assignments of `open` calls in a with statement. Without explicit closing, these resources will be "leaked" and won't be re-claimed until garbage collection. In situations where these resources are leaked rapidly (either through malicious repetitive action or unusually spiky usage), connection pool or file handle exhaustion will occur. These types of failures tend to be catastrophic, resulting in downtime and many times affect downstream applications.
```

tests/codemods/test_file_resource_leak.py
```diff
@@ -0,0 +1,166 @@
...
+        input_code = """\
+        file = open('test.txt', 'r')
+        file.read()
+        pass
+        """
+        expected = """\
+        with open('test.txt', 'r') as file:
+            file.read()
+        pass
+        """
```
clavedeluna, line 70: From this test I gather that the logic here is to indent everything after open call up to the read call, is that right?
So
```
var = 1
file = open('test.txt', 'r')
var = 2
file.read()
pass
```

would result in `var = 2` also being in the indented block, even if author didn't want that.

src/codemodder/codemods/utils.py
```diff
@@ -104,6 +106,40 @@ def on_leave(self, original_node, updated_node):
...
+    def __init__(self, context: CodemodContext) -> None:
+        MetadataDependent.__init__(self)
+        MatcherDecoratableTransformer.__init__(self)
+        self.context = context
+        dependencies = self.get_inherited_dependencies()
+        if dependencies:
+            wrapper = self.context.wrapper
+            if wrapper is None:
+                # pylint: disable-next=broad-exception-raised
+                raise Exception(
```
clavedeluna: I would suggest `ValueError` since it's related to something being None, and remove the pylint disable

src/core_codemods/file_resource_leak.py
```diff
@@ -0,0 +1,424 @@
...
+)
+from codemodder.codemods.base_visitor import UtilsMixin
+from codemodder.codemods.utils import MetadataPreservingTransformer
+from codemodder.codemods.utils_mixin import AncestorPatternsMixin, NameResolutionMixin
+from codemodder.file_context import FileContext
+from functools import partial
+
+
+class FileResourceLeak(BaseCodemod, UtilsMixin, Codemod):
+    SUMMARY = "Automatically close resources"
```
clavedeluna:
```suggestion
    SUMMARY = "Automatically Close Resources"
```

src/core_codemods/file_resource_leak.py
```diff
@@ -0,0 +1,424 @@
...
+)
+from libcst.metadata import (
+    BuiltinAssignment,
+    ParentNodeProvider,
+    PositionProvider,
+    ScopeProvider,
+)
+from codemodder.change import Change
+from codemodder.codemods.base_codemod import (
+    BaseCodemod,
```
clavedeluna: Any reason not to use api.BaseCodemod?

src/core_codemods/file_resource_leak.py
```diff
@@ -0,0 +1,424 @@
...
+                self._extract_targets_of_assignment(maybe_assigned)
+            )
+            for n in named_targets:
+                n_named_targets, n_other_targets = self._find_name_assignment_targets(n)
+                named_targets.extend(n_named_targets)
+                other_targets.extend(n_other_targets)
+            return named_targets, other_targets
+        return ([], [])
+
+    def _extract_targets_of_assignment(
```
clavedeluna: can now use codemodder.utils.extract_targets_of_assignment

clavedeluna (changes requested):
I still need to more fully review this beast of a codemod. Great job working through all this

tests/codemods/test_file_resource_leak.py
andrecsilva: We wrap all the statements until the last statement that references the `with` variable.
The situation in your example is deliberate. We don't really want to change the order of variables willy nilly.
Scope wise, having it inside the `with` block is fine as python has function level scope, that is, after the transformation, `var` will still be reachable anywhere else.

src/codemodder/codemods/utils.py
andrecsilva: This was actually copied directly from `libcst`'s `ContextAwareVisitor.__init__`. The only reason this class exists is because `ContextAwareTransformer` will handle metadata differently.

No reason we cannot improve as per your suggestion, just giving some context.

clavedeluna (approved):

src/codemodder/scripts/generate_docs.py
```diff
@@ -138,6 +138,10 @@ class DocMetadata:
         importance="Medium",
         guidance_explained="Our change fixes explicitly insecure session configuration for a Flask application. However, there may be valid cases to use these insecure configurations, such as for testing or backward compatibility.",
     ),
+    "file-resource-leak": DocMetadata(
```
drdavella: It seems like most of our codemod names are framed in terms of the fix and not the problem. A simple suggestion would be `fix-file-resource-leak` but I'm open to suggestions.

src/core_codemods/file_resource_leak.py
```diff
@@ -0,0 +1,402 @@
...
+)
+from codemodder.codemods.api import BaseCodemod
+from codemodder.codemods.utils import MetadataPreservingTransformer
+from codemodder.codemods.utils_mixin import AncestorPatternsMixin, NameResolutionMixin
+from codemodder.file_context import FileContext
+from functools import partial
+
+
+class FileResourceLeak(BaseCodemod):
+    NAME = "file-resource-leak"
```
drdavella: See comment above about the name.

drdavella (approved):
This is a very nice codemod with great test cases. I just have a minor comment about the name.

PR #148, opened 2023-11-23
Integrate Repo Manager

clavedeluna:
## Overview
*Turn on python repo manager initial analysis and refactor dependency manager to handle more than just requirements.txt parsing*

## Description

* This PR turns on PRM to run before any codemod runs. When `repo_manager.parse_project()` is run, codemodder will run all available file parsers and determine which is the appropriate location to add dependencies for this project.
* Then, when a codemod needs to add a dependency, it adds the dependency to the file context. At the end of a codemod analyzing all files, the execution context will call on the dependency manager.
* The dependency manager is now responsible for determining where to write the new dependencies. It relies on the fact that PRM already has determined the `dependencies_store`. It then picks the appropriate writer.
* The only enabled writer right now is `RequirementsTxtWriter`, which contains much of what dependency manager used to do (it used to just work with requirements.txt).
* One important thing to note is that at this each codemod that needs to add a dependency is re-parsing the dependency file. Later on we can optimize, such as decide not to parse the dependencies at all at startup.

## Future work
* Add more writers. Need to enable the setup.py writer (which is a codemod) and there is a separate PR for pyproject.toml.
* More complex heuristics for picking where to write dependencies. Many projects can have setup.py and requirements.txt, among other files. We need to update `PythonRepoManager.dependencies_store` to not just pick the first store found.

### DIFF
--- a/src/codemodder/dependency_management/base_dependency_writer.py
+++ b/src/codemodder/dependency_management/base_dependency_writer.py
@@ -21,12 +21,12 @@ class DependencyWriter(metaclass=ABCMeta):
     ) -> Optional[ChangeSet]:
         pass
 
-    def add(self, dependencies: list[Dependency]) -> Optional[list[Dependency]]:
+    def add(self, dependencies: list[Dependency]) -> list[Dependency]:
         """add any number of dependencies to the end of list of dependencies."""
         new = []
         for new_dep in dependencies:
             requirement: Requirement = new_dep.requirement
             if requirement not in self.dependency_store.dependencies:
-                self.dependency_store.dependencies.append(requirement)
+                self.dependency_store.dependencies.add(requirement)
                 new.append(new_dep)
         return new
--- a/src/codemodder/dependency_management/base_dependency_writer.py
+++ b/src/codemodder/dependency_management/base_dependency_writer.py
@@ -1,0 +1,19 @@
+from abc import ABCMeta, abstractmethod
+from pathlib import Path
+from typing import Optional
+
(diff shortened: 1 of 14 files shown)


src/codemodder/dependency_management/base_dependency_writer.py
```diff
@@ -0,0 +1,32 @@
...
+        self.path = Path(dependency_store.file)
+        self.parent_directory = parent_directory
+
+    @abstractmethod
+    def write(
+        self, dependencies: list[Dependency], dry_run: bool = False
+    ) -> Optional[ChangeSet]:
+        pass
+
+    def add(self, dependencies: list[Dependency]) -> Optional[list[Dependency]]:
```
andrecsilva: `Optional[list[Dependency]]` -> `list[Dependency]`.
I usually expect a `None` with `Optional[...]` but here it just returns an empty list.

src/codemodder/dependency_management/base_dependency_writer.py
```diff
@@ -0,0 +1,32 @@
...
+        """add any number of dependencies to the end of list of dependencies."""
+        new = []
+        for new_dep in dependencies:
+            requirement: Requirement = new_dep.requirement
+            if requirement not in self.dependency_store.dependencies:
```
andrecsilva, line 29: `PackageStore.dependencies` is a list. This particular test will take `O(n)`.
Suggestion:
Change `PackageStore.dependencies` to be a set.

andrecsilva (approved):

PR #150, opened 2023-11-28
pyproject.toml writer

clavedeluna:
## Overview
*Add a pyproject.toml dependency writer*

## Description

* Following the example of the requirements.txt writer, add one for pyproject.toml, IFF it has a [project] dependencies section in the file
* the one interesting difference for this writer is that since we're using `tomlkit` to parse and update the toml file, we didn't have a way to get the line numbers that changed. For that, I created a custom function that uses the output of `difflib.unified_diff` and parses out the `@@ -x,y +a,b @@` and any `+ ...` lines.

### DIFF
--- a/src/codemodder/dependency_management/dependency_manager.py
+++ b/src/codemodder/dependency_management/dependency_manager.py
@@ -4,6 +4,7 @@ from codemodder.dependency import Dependency
 from codemodder.dependency_management.requirements_txt_writer import (
     RequirementsTxtWriter,
 )
+from codemodder.dependency_management.pyproject_writer import PyprojectWriter
 from codemodder.project_analysis.file_parsers.package_store import PackageStore
 from pathlib import Path
 
@@ -25,6 +26,10 @@ class DependencyManager:
         match self.dependencies_store.type:
             case "requirements.txt":
                 return RequirementsTxtWriter(
+                    self.dependencies_store, self.parent_directory
+                ).write(dependencies, dry_run)
+            case "pyproject.toml":
+                return PyprojectWriter(
                     self.dependencies_store, self.parent_directory
                 ).write(dependencies, dry_run)
             case "setup.py":
--- a/src/codemodder/dependency_management/pyproject_writer.py
+++ b/src/codemodder/dependency_management/pyproject_writer.py
@@ -4,7 +4,7 @@ from copy import deepcopy
 from codemodder.dependency import Dependency
 from codemodder.change import Action, Change, ChangeSet, PackageAction, Result
 from codemodder.dependency_management.base_dependency_writer import DependencyWriter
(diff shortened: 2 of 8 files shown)


andrecsilva (approved):

src/codemodder/dependency_management/pyproject_writer.py
```diff
@@ -0,0 +1,55 @@
...
+        diff, added_line_nums = create_diff_and_linenums(
+            tomlkit.dumps(original).split("\n"), tomlkit.dumps(pyproject).split("\n")
+        )
+
+        if not dry_run:
+            with open(self.path, "w", encoding="utf-8") as f:
+                tomlkit.dump(pyproject, f)
+
+        changes = [
+            Change(
```
drdavella, line 33: I think this should be factored out into the base class now since it includes some fiddly logic that will be pretty much the same for each writer.

tests/dependency_management/test_pyproject_writer.py
```diff
@@ -0,0 +1,209 @@
+from textwrap import dedent
+import pytest
+
+from codemodder.dependency_management.pyproject_writer import PyprojectWriter
+from codemodder.dependency import DefusedXML, Security
+from codemodder.project_analysis.file_parsers.package_store import PackageStore
+
+
```
drdavella, line 8: It would be good to have a test case somewhere showing whether or not we preserve comments since this is going to be pretty important.

src/codemodder/dependency_management/dependency_manager.py
```diff
@@ -27,6 +28,10 @@ def write(
                 return RequirementsTxtWriter(
                     self.dependencies_store, self.parent_directory
                 ).write(dependencies, dry_run)
+            case "pyproject.toml":
```
drdavella, line 31: I feel pretty strongly that these kinds of values should be represented by an enum. The fact that it is being used as cases for `match` really emphasizes that to me.

src/codemodder/dependency_management/pyproject_writer.py
```diff
@@ -0,0 +1,55 @@
...
+    ) -> Optional[ChangeSet]:
+        pyproject = self._parse_file()
+        original = deepcopy(pyproject)
+
+        try:
+            pyproject["project"]["dependencies"].extend(
+                [f"{dep.requirement}" for dep in dependencies]
+            )
+        except tomlkit.exceptions.NonExistentKey:
+            return None
```
drdavella, line 22: I think we probably ought to have some kind of debug logging either here or at the caller.

drdavella (commented):
@clavedeluna overall this work looks really nice but I'm adding some comments here since I didn't have the opportunity to review before this was merged.

PR #152, opened 2023-11-29
Setup py writer

clavedeluna:
## Overview
*Integrate the setup.py writer to run and refactor*

## Description

* We had a basic setup.py writer but it hadn't yet aligned with the other writers nor been enabled
* Some refactors

Work we still need:
- more robust file error handling for all parsers
- implement the heuristic for picking which dependency file to use: 
> if there’s a pyproject.toml with dependencies, make updates there. Otherwise, if there’s a setup.py with dependencies, make the updates there. Finally, if neither of those are present, try to find requirements.txt and make the update there.


clavedeluna:
@drdavella I've addressed all your comments from previouis PR https://github.com/pixee/codemodder-python/pull/150#discussion_r1407959018 in this PR

The only one that isn't as clean is refactoring the Change list. Since each writer has a tiny difference in how we get the line number. If you have a better idea lmk.

src/codemodder/dependency_management/dependency_manager.py
```diff
@@ -24,14 +28,21 @@ def write(
...
                 return PyprojectWriter(
                     self.dependencies_store, self.parent_directory
                 ).write(dependencies, dry_run)
-            case "setup.py":
-                pass
+            case FileType.SETUP_PY:
+                # pylint: disable-next=cyclic-import
+                from codemodder.dependency_management.setup_py_writer import (
+                    SetupPyWriter,
+                )
```
clavedeluna:
```
integration_tests/test_use_walrus_if.py:1:0: R0401: Cyclic import (codemodder.codemods.base_codemod -> codemodder.semgrep -> codemodder.context -> codemodder.dependency_management -> codemodder.dependency_management.dependency_manager -> codemodder.dependency_management.setup_py_writer) (cyclic-import)
```
pretty annoyed with this. the issue is that setup_py_writer has a codemod and uses `from codemodder.codemods.api import BaseCodemod`. Tried a few quick things but wasn't convinced I liked any or some didn't work. Thoughts?

drdavella:
> The only one that isn't as clean is refactoring the Change list. Since each writer has a tiny difference in how we get the line number. If you have a better idea lmk.

@clavedeluna would it be possible to pass the line number as a parameter to a factored out method? If so I think it's a worthwhile change because I really want to avoid having to make updates to this in multiple places.

src/codemodder/dependency_management/dependency_manager.py
drdavella: I think the only other way to fix this is maybe by importing `DependencyManager` from within `CodemodExecutionContext.process_dependencies`. Maybe that's a little cleaner? Either way it's not ideal but I'm also not too worried about it.

andrecsilva: Not sure here, maybe rewrite `SetupPyAddDependencies` as a pure libcst codemod, that is, use `libcst.Codemod` instead of `BaseCodemod`.

clavedeluna: > Not sure here, maybe rewrite SetupPyAddDependencies as a pure libcst codemod, that is, use libcst.Codemod instead of BaseCodemod.

having this codemod rely on our own base is really useful because we have already solved some of the things this codemod needs (updating args)

I'll either move the import or just ignore for now.

andrecsilva (approved):

drdavella (approved):
Good stuff 👍

PR #156, opened 2023-12-01
Codemod for numpy NaN equality bug

andrecsilva:
## Overview
This codemod will change any equality to `numpy.nan` using `==` to use `numpy.isnan()` instead.

### DIFF
--- a/src/core_codemods/docs/pixee_python_numpy-nan-equality.md
+++ b/src/core_codemods/docs/pixee_python_numpy-nan-equality.md
@@ -1,4 +1,4 @@
-Comparisons against `numpy.nan` always results in `False`. Thus comparing an expression directly against `numpy.nan` is always unintended. The correct way to compare a value for `NaN` is to use the `numpy.isnan` function.
+Comparisons against `numpy.nan` always result in `False`. Thus comparing an expression directly against `numpy.nan` is always unintended. The correct way to compare a value for `NaN` is to use the `numpy.isnan` function.
 
 Our changes look something like this:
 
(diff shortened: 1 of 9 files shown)


src/core_codemods/docs/pixee_python_numpy-nan-equality.md
```diff
@@ -0,0 +1,12 @@
+Comparisons against `numpy.nan` always results in `False`. Thus comparing an expression directly against `numpy.nan` is always unintended. The correct way to compare a value for `NaN` is to use the `numpy.isnan` function.
```
drdavella:
```suggestion
Comparisons against `numpy.nan` always result in `False`. Thus comparing an expression directly against `numpy.nan` is always unintended. The correct way to compare a value for `NaN` is to use the `numpy.isnan` function.
```

tests/codemods/test_numpy_nan_equality.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        expected = """\
+        import numpy as np
+        if np.isnan(a):
+            pass
+        """
+        self.run_and_assert(tmpdir, dedent(input_code), dedent(expected))
+        assert len(self.file_context.codemod_changes) == 1
+
+    def test_multiple_comparisons(self, tmpdir):
```
drdavella, line 54: I know it's sort of implicit in this and the next test case but it would be good to have an explicit test case with `np.nan` on the LHS of the comparison.

src/core_codemods/numpy_nan_equality.py
```diff
@@ -0,0 +1,120 @@
...
+from codemodder.codemods.base_codemod import ReviewGuidance
+
+from codemodder.codemods.utils_mixin import NameResolutionMixin
+
+NodeWithTrueName = namedtuple("NodeWithTrueName", ["node", "name"])
+
+
+class NumpyNanEquality(BaseCodemod, NameResolutionMixin):
+    NAME = "numpy-nan-equality"
+    SUMMARY = "Uses numpy.isnan() instead of ==."
```
drdavella:
```suggestion
    SUMMARY = "Replace == comparison with numpy.isnan()"
```

drdavella (approved):
Another very nice codemod overall. Just some minor comments.

PR #158, opened 2023-12-04
Set content-type for Django Json Response Type Codemod

andrecsilva:
## Overview
Sets `content-type` to `application-json` for `HttpResponse()` calls in Django if the argument is a json string.

### DIFF
--- a/src/core_codemods/django_json_response_type.py
+++ b/src/core_codemods/django_json_response_type.py
@@ -6,7 +6,7 @@ from codemodder.codemods.api import SemgrepCodemod
 
 class DjangoJsonResponseType(SemgrepCodemod):
     NAME = "django-json-response-type"
-    SUMMARY = "Restricts the type to `json/application` for django.http.HttpResponse with json objects as argument."
+    SUMMARY = "Set content type to `json/application` for `django.http.HttpResponse` with JSON data"
     REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
     DESCRIPTION = "Sets `content_type` to `json/application`."
     REFERENCES = [
@@ -42,6 +42,10 @@ class DjangoJsonResponseType(SemgrepCodemod):
                 cst.Arg(
                     value=cst.parse_expression('"application/json"'),
                     keyword=cst.Name("content_type"),
+                    equal=cst.AssignEqual(
+                        whitespace_before=cst.SimpleWhitespace(""),
+                        whitespace_after=cst.SimpleWhitespace(""),
+                    ),
                 ),
             ],
         )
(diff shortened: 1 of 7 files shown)


tests/codemods/test_django_json_response_type.py
```diff
@@ -0,0 +1,92 @@
...
+            json_response = json.dumps({ "user_input": request.GET.get("input") })
+            return HttpResponse(json_response)
+        """
+        expected = """\
+        from django.http import HttpResponse
+        import json
+
+        def foo(request):
+            json_response = json.dumps({ "user_input": request.GET.get("input") })
+            return HttpResponse(json_response, content_type = "application/json")
```
drdavella: Sorry I know this is a relatively small thing but the spaces on either side of the `=` in a kwarg assignment is not considered as stylistically correct and will be flagged by many tools. We should make sure to format it correctly. I think this should be possible via whitespace nodes in the libcst representation.

src/core_codemods/django_json_response_type.py
```diff
@@ -0,0 +1,47 @@
+import libcst as cst
+
+from codemodder.codemods.base_codemod import ReviewGuidance
+from codemodder.codemods.api import SemgrepCodemod
+
+
+class DjangoJsonResponseType(SemgrepCodemod):
+    NAME = "django-json-response-type"
+    SUMMARY = "Restricts the type to `json/application` for django.http.HttpResponse with json objects as argument."
```
drdavella:
```suggestion
    SUMMARY = "Set content type to `json/application` for `django.http.HttpResponse` with JSON data"
```

src/core_codemods/docs/pixee_python_django-json-response-type.md
```diff
@@ -0,0 +1,11 @@
+For a `HttpReponse` in Django, if the `content_type` is not set, it default to `text/html`. If the JSON contains user supplied input without being sanitized first, a malicious user may supply HTTP code which leaves the application vulnerable to cross-site scripting (XSS). This transformation restricts the response type to `application/json` to avoid this vulnerability. Our changes look something like this:
```
drdavella: Just a suggestion.
```suggestion
The default `content_type` for `HttpResponse` in Django is `'text/html'`. This is true even when the response contains JSON data.

If the JSON contains (unsanitized) user-supplied input, a malicious user may supply HTML code which leaves the application vulnerable to cross-site scripting (XSS). 

This fix explicitly sets the response type to `application/json` when the response body is JSON data to avoid this vulnerability. Our changes look something like this:
```

drdavella (changes requested):
Nice codemod overall but please see comments about whitespace and suggestions about metadata.

drdavella (approved):

PR #161, opened 2023-12-05
Increased test timeout in CI

andrecsilva:


drdavella (approved):
Is this because of the Python 3.12 tests? It bothers me that we keep having to do this. It would be good to get to the root cause and see what we can do to bring this back down. Especially since the tests run relatively quickly locally.

andrecsilva:
> Is this because of the Python 3.12 tests? It bothers me that we keep having to do this. It would be good to get to the root cause and see what we can do to bring this back down. Especially since the tests run relatively quickly locally.

Yes, I've had it timeout after a merge to main today, hence the increase. It may be worth doing some investigation on this.

PR #162, opened 2023-12-05
Added flask-json-response-type codemod

andrecsilva:
## Overview
Added flask-json-response-type codemod.

## Description

* Added flask version of django-json-response-type codemod.
* ~~The reason why this one is just not a case of django-json-response-type codemod stems from the fact that we would need to identify which pattern rule was active when making the change. This would require the use of the ScopeProvider and would make it expensive.~~

### DIFF
--- a/src/core_codemods/docs/pixee_python_flask-json-response-type.md
+++ b/src/core_codemods/docs/pixee_python_flask-json-response-type.md
@@ -1,0 +1,16 @@
+The default `mimetype` for `make_response` in Flask is `'text/html'`. This is true even when the response contains JSON data.
+If the JSON contains (unsanitized) user-supplied input, a malicious user may supply HTML code which leaves the application vulnerable to cross-site scripting (XSS). 
+This fix explicitly sets the response type to `application/json` when the response body is JSON data to avoid this vulnerability. Our changes look something like this:
+
+```diff
+from flask import make_response, Flask
+import json
+
+app = Flask(__name__)
+
+@app.route("/test")
+def foo(request):
+    json_response = json.dumps({ "user_input": request.GET.get("input") })
+-   return make_response(json_response)
++   return (make_response(json_response), {'Content-Type':'application/json'}
+```
--- a/src/core_codemods/docs/pixee_python_flask-json-response-type.md
+++ b/src/core_codemods/docs/pixee_python_flask-json-response-type.md
@@ -12,5 +12,5 @@ app = Flask(__name__)
 def foo(request):
     json_response = json.dumps({ "user_input": request.GET.get("input") })
 -   return make_response(json_response)
-+   return (make_response(json_response), {'Content-Type':'application/json'}
++   return make_response(json_response, {'Content-Type':'application/json'})
 ```
--- a/tests/codemods/test_flask_json_response_type.py
+++ b/tests/codemods/test_flask_json_response_type.py
@@ -35,6 +35,32 @@ class TestFlaskJsonResponseType(BaseCodemodTest):
         self.run_and_assert(tmpdir, dedent(input_code), dedent(expected))
         assert len(self.file_context.codemod_changes) == 1
 
+    def test_simple_return_json(self, tmpdir):
+        input_code = """\
+        from flask import Flask
+        import json
+
+        app = Flask(__name__)
+
+        @app.route("/test")
+        def foo(request):
+            json_response = json.dumps({ "user_input": request.GET.get("input") })
+            return json_response
+        """
+        expected = """\
+        from flask import Flask
+        import json
+
+        app = Flask(__name__)
+
+        @app.route("/test")
+        def foo(request):
+            json_response = json.dumps({ "user_input": request.GET.get("input") })
+            return (json_response, {'Content-Type': 'application/json'})
+        """
+        self.run_and_assert(tmpdir, dedent(input_code), dedent(expected))
+        assert len(self.file_context.codemod_changes) == 1
+
     def test_alias(self, tmpdir):
         input_code = """\
         from flask import make_response as response, Flask
--- a/tests/codemods/test_flask_json_response_type.py
+++ b/tests/codemods/test_flask_json_response_type.py
@@ -30,7 +30,61 @@ class TestFlaskJsonResponseType(BaseCodemodTest):
         @app.route("/test")
         def foo(request):
             json_response = json.dumps({ "user_input": request.GET.get("input") })
-            return (make_response(json_response), {'Content-Type': 'application/json'})
+            return make_response(json_response, {'Content-Type': 'application/json'})
+        """
+        self.run_and_assert(tmpdir, dedent(input_code), dedent(expected))
+        assert len(self.file_context.codemod_changes) == 1
+
+    def test_simple_indirect(self, tmpdir):
+        input_code = """\
+        from flask import make_response, Flask
+        import json
+
+        app = Flask(__name__)
+
+        @app.route("/test")
+        def foo(request):
+            json_response = json.dumps({ "user_input": request.GET.get("input") })
+            response = make_response(json_response)
+            return response
+        """
+        expected = """\
+        from flask import make_response, Flask
+        import json
+
+        app = Flask(__name__)
+
+        @app.route("/test")
+        def foo(request):
+            json_response = json.dumps({ "user_input": request.GET.get("input") })
+            response = make_response(json_response, {'Content-Type': 'application/json'})
+            return response
+        """
+        self.run_and_assert(tmpdir, dedent(input_code), dedent(expected))
+        assert len(self.file_context.codemod_changes) == 1
+
+    def test_simple_tuple_arg(self, tmpdir):
+        input_code = """\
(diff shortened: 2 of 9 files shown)


drdavella (changes requested):
This is a nice idea for a codemod. However, my main substantive comment is that `make_response` does not appear to accept a kwarg called `mimetype`: https://flask.palletsprojects.com/en/2.3.x/api/#flask.Flask.make_response
```
 File "/Users/danieldavella/sandbox/flask_json.py", line 12, in return_json_str
    return make_response(json.dumps(response), mimetype="application/json")
TypeError: make_response() got an unexpected keyword argument 'mimetype'
```

There are a probably a few different ways to implement this codemod. One would involve explicitly setting the header on the generated response:
```python
response = make_response(json.dumps(...))
response.headers['Content-Type'] = 'application/json'
```

Another (imo cleaner) way would involve using `werkzeug.wrappers.Response` to set the content type. `werkzeug` is an explicit dependency of `flask` so there is no issue with adding this import:
```python
from werkzeug.wrappers import Response

def return_json_str():
    response_data = {"message": "This is a JSON response"}
    response_json = json.dumps(response_data)
    return Response(response_json, content_type='application/json')
```

It's also worth mentioning that in newer versions of `flask` it is possible to return a `dict` directly from a view and the response will be properly interpreted as JSON data.

I also think there are also a few other cases we can handle for `flask`. You can feel free to add them to this PR or create a ticket/add it in another PR.

1. It is possible to return a string directly from a view in Flask:
```python
@app.route("/return-json-str")
def return_json_str():
    return json.dumps({ "user_input": request.GET.get("input") })
```

2. The `make_request` function also takes a tuple argument where the body is the first element:
```python
def foo():
    data = json.dumps({ "user_input": request.GET.get("input") })
    headers = {}
    return make_response((data, 200, headers))
```

tests/codemods/test_flask_json_response_type.py
```diff
@@ -0,0 +1,238 @@
...
+        expected = """\
+        from flask import make_response, Flask
+        import json
+
+        app = Flask(__name__)
+
+        @app.route("/test")
+        def foo(request):
+            json_response = json.dumps({ "user_input": request.GET.get("input") })
+            return (make_response(json_response), {'Content-Type': 'application/json'})
```
drdavella: Wouldn't the simplest solution in this case to be to just return `json_response` since Flask will automatically set the proper context type?

tests/codemods/test_flask_json_response_type.py
```diff
@@ -0,0 +1,238 @@
...
+        input_code = """\
+        from flask import Flask
+        import json
+
+        app = Flask(__name__)
+
+        @app.route("/test")
+        def foo(request):
+            json_response = json.dumps({ "user_input": request.GET.get("input") })
+            return json_response
```
drdavella, line 102: I think this does not need to be modified as flask will automatically set the correct content type in this case.

tests/codemods/test_flask_json_response_type.py
```diff
@@ -0,0 +1,238 @@
+from core_codemods.flask_json_response_type import FlaskJsonResponseType
+from tests.codemods.base_codemod_test import BaseCodemodTest
+from textwrap import dedent
+
+
+class TestFlaskJsonResponseType(BaseCodemodTest):
```
drdavella, line 6: What about a test case where a variety of different headers are already set? We need to make sure the only modified header is `Content-Type`.

tests/codemods/test_flask_json_response_type.py
drdavella: Actually, according to the `flask` documentation, `make_response` will also correctly handle the content type by calling `jsonify` on a `dict`: https://flask.palletsprojects.com/en/3.0.x/api/#flask.Flask.make_response

src/core_codemods/docs/pixee_python_flask-json-response-type.md
```diff
@@ -0,0 +1,16 @@
...
+from flask import make_response, Flask
+import json
+
+app = Flask(__name__)
+
+@app.route("/test")
+def foo(request):
+    json_response = json.dumps({ "user_input": request.GET.get("input") })
+-   return make_response(json_response)
++   return (make_response(json_response), {'Content-Type':'application/json'}
```
drdavella: See other comments regarding the overall conditions for this change, but I think it would be more idiomatic and more aligned with the purpose of `make_response` to call it like this:
```python
return make_response(json_response, {'Content-Type':'application/json'})
```

See the documentation and specifically the `tuple` parameter case: https://flask.palletsprojects.com/en/3.0.x/api/#flask.Flask.make_response

drdavella (changes requested):
Now that this codemod has been restricted only to decorated Flask view functions, we need to revisit the conditions where this fix is actually necessary. Flask will call `jsonify` by default on `dict` objects that are returned from view functions. Furthermore, it appears that `make_response` also calls `jsonify` on `dict` objects. 

This would suggest that the set of problematic conditions is confined to cases where the wrong content type is being set explicitly, either with `make_response`, or in a tuple returned by a view function, or when used to construct a `Response` object.

src/core_codemods/docs/pixee_python_flask-json-response-type.md
andrecsilva: The reason for the tuple response is because it is more general. It will also work for the direct `json.dumps` response. I'll adjust the transformation for the `make_response` case.

tests/codemods/test_flask_json_response_type.py
andrecsilva: It won't. The return of `json.dumps()` is a string and `flask` will only apply `jsonify` for dict returns.

andrecsilva:
> Now that this codemod has been restricted only to decorated Flask view functions, we need to revisit the conditions where this fix is actually necessary. Flask will call `jsonify` by default on `dict` objects that are returned from view functions. Furthermore, it appears that `make_response` also calls `jsonify` on `dict` objects.
> 
> This would suggest that the set of problematic conditions is confined to cases where the wrong content type is being set explicitly, either with `make_response`, or in a tuple returned by a view function, or when used to construct a `Response` object.

First, note that the return of `json.dumps()` is a string, not a dict. This codemod will only act when the actual response is a result of `json.dumps(...)` either through `make_response`, a tuple or directly. In those cases, `jsonify` won't actually be called since the response is a string.

If you need to convince yourself, simply run the following small flask app:
```python
from flask import Flask, jsonify, make_response
import json

app = Flask(__name__)

@app.route("/")
def hello_world():
    json_response = json.dumps({'data':'Hello World dict'})
    return make_response(json_response))

```
with 
```bash 
FLASK_APP=flask_test flask run
```
and check the `Content-Type` header from the response with your browser or:
```
curl -iX GET http://localhost:5000
```

The most idiomatic response is to actually remove the `json.dumps` call and either return its argument directly or explicitly call `jsonify`. The reason I've opted to use a tuple response is twofold: it's simpler and will only require local modifications on the `return` statement while also being more explicit about the change, and `jsonify` will call `app.json.dumps` which may actually not be the same as `json.dumps` (unlikely, but still possible). 
See:
https://flask.palletsprojects.com/en/3.0.x/api/#flask.json.dumps
https://github.com/pallets/flask/blob/089f6a1c50ea2b18287106ecd357522bab200510/src/flask/json/provider.py#L74

drdavella:
Yes you're right, sorry I forgot that the output of `json.dumps` was a string. I agree that the most idiomatic behavior in this case would be to remove the call to `json.dumps` entirely, although that is probably a difficult change to make. Also I understand the rationale for returning the tuple but it just doesn't look like a clean change when `make_response` is involved, so hopefully we can make it such that those cases don't overlap.

I realize this codemod has become substantially more complicated than you or I originally anticipated, so I'd be okay with descoping it if that makes it easier for you and we can revisit/refine it later.

andrecsilva:
> Yes you're right, sorry I forgot that the output of `json.dumps` was a string. I agree that the most idiomatic behavior in this case would be to remove the call to `json.dumps` entirely, although that is probably a difficult change to make. Also I understand the rationale for returning the tuple but it just doesn't look like a clean change when `make_response` is involved, so hopefully we can make it such that those cases don't overlap.
> 
> I realize this codemod has become substantially more complicated than you or I originally anticipated, so I'd be okay with descoping it if that makes it easier for you and we can revisit/refine it later.

I'm already working on making the changes to `make_response` as you suggested. This codemod certainly was way more touble than expected, but some of the new functions added will help with other codemods in the future.

tests/samples/flask_json_response_type.py
```diff
@@ -0,0 +1,9 @@
+from flask import make_response, Flask
+import json
+
+app = Flask(__name__)
+
+@app.route("/test")
+def foo(request):
+    json_response = json.dumps({ "user_input": request.GET.get("input") })
+    return make_response(json_response)
```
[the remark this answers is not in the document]
drdavella: 😆

drdavella (approved):
Very nice, thanks for powering through on this one 😅

PR #164, opened 2023-12-06
Fix edge case for `use-walrus-if`; implement without semgrep

drdavella:
## Overview
*Restrict cases where codemod is applicable; implement without semgrep*

## Description

* In order to avoid some weird and hard-to-understand code changes, this restricts the cases where this codemod is applicable to the following operators: `==`, `!=`, `is`, and `is not`
* I also rewrote the codemod without semgrep which yields considerable performance improvements
  * I think there's probably a way to do everything from within `on_leave` but it's a bigger change and it works pretty well with the preexisting architecture

### DIFF
--- a/src/core_codemods/use_walrus_if.py
+++ b/src/core_codemods/use_walrus_if.py
@@ -1,3 +1,5 @@
+from collections import namedtuple
+import itertools
 from typing import List, Tuple, Optional
 
 import libcst as cst
@@ -6,11 +8,20 @@ from libcst import matchers as m
 from libcst.metadata import ParentNodeProvider, ScopeProvider
 
 from codemodder.codemods.base_codemod import ReviewGuidance
-from codemodder.codemods.api import SemgrepCodemod
+from codemodder.codemods.api import BaseCodemod
 
 
-class UseWalrusIf(SemgrepCodemod):
-    METADATA_DEPENDENCIES = SemgrepCodemod.METADATA_DEPENDENCIES + (
+FoundAssign = namedtuple("FoundAssign", ["assign", "target", "value"])
+
+
+def pairwise(iterable):
+    a, b = itertools.tee(iterable)
+    next(b, None)
+    return zip(a, b)
+
+
+class UseWalrusIf(BaseCodemod):
+    METADATA_DEPENDENCIES = BaseCodemod.METADATA_DEPENDENCIES + (
         ParentNodeProvider,
         ScopeProvider,
     )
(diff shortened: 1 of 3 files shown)


src/core_codemods/use_walrus_if.py
```diff
@@ -27,52 +34,85 @@ class UseWalrusIf(SemgrepCodemod):
...
+                                    operator=(
+                                        cst.Is()
+                                        | cst.IsNot()
+                                        | cst.Equal()
+                                        | cst.NotEqual()
+                                    )
+                                )
+                            ],
+                        ):
+                            # TODO: use actual scope analysis?
```
andrecsilva: Since you guarantee those are consecutive statements, there is no need to do more complex scope analysis.

src/core_codemods/use_walrus_if.py
```diff
@@ -27,52 +34,85 @@ class UseWalrusIf(SemgrepCodemod):
...
+    def _build_named_expr(self, target, value, parens=True):
+        return cst.NamedExpr(
+            target=target,
+            value=value,
+            lpar=[cst.LeftParen()] if parens else [],
+            rpar=[cst.RightParen()] if parens else [],
+        )
 
-    def visit_If(self, node):
+    def on_visit(self, node: cst.CSTNode) -> Optional[bool]:
```
andrecsilva, line 80: It seems the purpose of this method is to identify two consecutive statements each that satisfies a predicate.

The way it is, it is a bit hard to understand right now. My suggestion is to break this into a few methods: (1) `handle_block`, (2) `is_name_assign() -> Optional[cst.Assign]`, (3) `is_named_comparison_if -> Optional[cst.If]`. As always, suggestions are optional.

Functions (2) and (3) are straightforward and just test the desired properties.  For (1), there are three nodes that may contain a sequence of statements (actually 4 since `cst.SimpleStatementLine` may contain more than one statement, but we should ignore that for now): `cst.Module`, `cst.IndentedBlock` and `cst.SimpleStatementSuite`. They all have the `body` attribute which contains the statements in order. So you can use `visit_Module`, `visit_SimpleStatementSuite` and `visit_IndentedBlock` all which will simply call `handle_block`.

Thus the implemenation of `handle_block` would look like:
```
for s1, s2 in pairwise(original_node.body):
  assign = is_name_assign(s1)
  if_stmt = is_named_comparison_if(s2)
  if assign and if_stmt:
     <do what you need>
```

andrecsilva (approved):

src/core_codemods/use_walrus_if.py
drdavella: 👍 that's what I was thinking too. There's definitely a more sophisticated version of this codemod that identifies non-adjacent statements using scope analysis. But that's obviously more difficult to implement and I also think it's a lot less safe in general since reordering statements can have unanticipated side effects. I'm happy with this restriction for now.

src/core_codemods/use_walrus_if.py
drdavella: Good suggestion. This method is a bit unwieldy right now. The `match` statements are really useful but they add a lot of indentation.

PR #169, opened 2023-12-07
New codemod: fix-deprecated-abstractproperty

drdavella:
## Overview
*Add new codemod for `fix-deprecated-abstractproperty`*

## Description

* This replaces deprecated `@abstractproperty` with `@property` and `@abstractmethod`
* Also made a minor fix to `numpy-nan-equality`

### DIFF
--- a/src/codemodder/codemods/utils_mixin.py
+++ b/src/codemodder/codemods/utils_mixin.py
@@ -17,24 +17,41 @@ from libcst.metadata.scope_provider import GlobalScope
 class NameResolutionMixin(MetadataDependent):
     METADATA_DEPENDENCIES: Tuple[Any, ...] = (ScopeProvider,)
 
+    def _find_imported_name(self, node: cst.Name):
+        match self.find_single_assignment(node):
+            case ImportAssignment(
+                name=node.value,
+                node=(
+                    cst.Import(names=names) | cst.ImportFrom(names=names)
+                ) as import_node,
+            ) as assignment:
+                match names:
+                    case cst.ImportStar():
+                        return node.value
+
+                for alias in names:
+                    if assignment.name in (
+                        alias.evaluated_alias,
+                        alias.evaluated_name,
+                    ):
+                        return self.base_name_for_import(import_node, alias)
+        return node.value
+
     def find_base_name(self, node):
         """
-        Given a node, solve its name to its basest form. For now it can only solve names that are imported. For example, in what follows, the base name for exec.capitalize() is sys.executable.capitalize.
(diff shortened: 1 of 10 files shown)


src/codemodder/codemods/utils_mixin.py
```diff
@@ -17,24 +17,41 @@
...
+        match self.find_single_assignment(node):
+            case ImportAssignment(
+                name=node.value,
+                node=(
+                    cst.Import(names=names) | cst.ImportFrom(names=names)
+                ) as import_node,
+            ) as assignment:
+                match names:
+                    case cst.ImportStar():
+                        return node.value
```
andrecsilva, line 30: This will never happen. You cannot attribute an assignment for a given to an import star because it can be ambiguous. Consider:
```python
from a import *
from b import *
foo()
```
Where does `foo()` come from?
If you need to convince yourself, try resolving the assignment for `Path` in this minimal example:
```python
from pathlib import *
Path("\")
```

src/codemodder/codemods/utils_mixin.py
```diff
@@ -17,24 +17,41 @@
 class NameResolutionMixin(MetadataDependent):
     METADATA_DEPENDENCIES: Tuple[Any, ...] = (ScopeProvider,)
 
+    def find_imported_name(self, node: cst.Name):
```
andrecsilva: I would make this "private". The original idea was to make `find_base_name` a single stop for everything.

src/core_codemods/fix_deprecated_abstractproperty.py
```diff
@@ -0,0 +1,42 @@
...
+            "url": "https://docs.python.org/3/library/abc.html#abc.abstractproperty",
+            "description": "",
+        },
+    ]
+
+    def leave_Decorator(
+        self, original_node: cst.Decorator, updated_node: cst.Decorator
+    ):
+        if base_name := self.find_base_name(original_node.decorator):
+            if base_name == "abc.abstractproperty":
```
andrecsilva: Suggestion:
Join those `if`s  with an `and`.

andrecsilva (approved):
Some minor issues.

src/codemodder/codemods/utils_mixin.py
drdavella: It can never happen but the type of `ImportFrom` allows `ImportStar` as the `name` so this is necessary for type correctness.

drdavella: Basically I could not figure out how else to restrict this using patterns in the match.

PR #172, opened 2023-12-11
Simplified numpy-nan-equality and added inequality support

andrecsilva:


tests/codemods/test_numpy_nan_equality.py
```diff
@@ -23,6 +23,20 @@ def test_simple(self, tmpdir):
         self.run_and_assert(tmpdir, dedent(input_code), dedent(expected))
         assert len(self.file_context.codemod_changes) == 1
 
+    def test_simple_inequality(self, tmpdir):
```
drdavella, line 26: Just for a sanity check could we add a few more test cases with more complex clauses. For example, things like:
```python
if not x and a != numpy.nan
```
```python
if not (a != numpy.nan):
```
```python
if a != numpy.nan and b != numpy.nan
```

Not necessarily for this ticket but we might want to simplify the result of some of these boolean expressions. I'm not sure whether LibCST provides a built-in codemod for that. If not, maybe it would be a good one for us to write.

drdavella (changes requested):

andrecsilva:
I've decided to drop support for multiple comparisons in the codemod. This means it will now ignore comparisons like this:
```
numpy.nan == foo() == numpy.nan
```

The reason is that these comparisons are a bit more permissive than I've initially thought. Also, we may be duplicating expressions with side-effects in them. Previously, the above expression would result in:
```
numpy.isnan(foo()) and numpy.isnan(foo())
```
which means `foo()` would be called twice, repeating any unintended side-effects.

tests/codemods/test_numpy_nan_equality.py
```diff
@@ -23,77 +23,114 @@ def test_simple(self, tmpdir):
...
-        if numpy.isnan(a):
+    def test_simple_inequality_2(self, tmpdir):
+        input_code = """\
+        import numpy
+        if not (a != numpy.nan):
+            pass
+        """
+        expected = """\
+        import numpy
+        if not (not numpy.isnan(a)):
```
drdavella, line 48: Simplifying boolean expressions feels like it might be a fun and useful codemod?

drdavella (approved):

tests/codemods/test_numpy_nan_equality.py
andrecsilva: Should make a fine codemod, specially as a transformation after codemods like this one.

PR #180, opened 2023-12-14
Added heuristics about which file to update in DependencyManager

andrecsilva:
- Also added a report listing if dependencies were added successfully or not.
- Added some robustness to RequirementsTxtWriter (this is mostly a band-aid and needs proper, more general, handling).

### DIFF
--- a/integration_tests/test_dependency_manager.py
+++ b/integration_tests/test_dependency_manager.py
@@ -3,6 +3,7 @@ import subprocess
 import shutil
 import pytest
 from integration_tests.base_test import SAMPLES_DIR, CleanRepoMixin
+from textwrap import dedent
 
 
 class TestDependencyManager(CleanRepoMixin):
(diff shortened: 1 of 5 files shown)


integration_tests/test_dependency_manager.py
```diff
@@ -0,0 +1,66 @@
+import os
+import subprocess
+import shutil
+import pytest
+from integration_tests.base_test import SAMPLES_DIR, CleanRepoMixin
+
+
+class TestDependencyManager(CleanRepoMixin):
```
drdavella, line 9: Shouldn't we have test cases here for updating `pyproject.toml` and `setup.py` as well? In those cases we could just add some simple content to those temporary files within individual test cases.

src/codemodder/context.py
```diff
@@ -85,28 +86,41 @@ def get_failed_files(self):
...
         # pylint: disable-next=cyclic-import
         from codemodder.dependency_management import DependencyManager
 
-        dm = DependencyManager(dependencies_store, self.directory)
-        if (changeset := dm.write(list(dependencies), self.dry_run)) is not None:
-            self.add_results(codemod_id, [changeset])
+        for package_store in store_list:
+            dm = DependencyManager(package_store, self.directory)
+            if (changeset := dm.write(list(dependencies), self.dry_run)) is not None:
+                self.add_results(codemod_id, [changeset])
```
drdavella, line 118: Maybe a debug log here saying which file is being written?

drdavella (changes requested):
Apologies for the delayed review. It looks good overall but I think we might need some more tests to enumerate a few more cases.

src/codemodder/context.py
andrecsilva: This information is being store in the `record` variable. The debug message with the file the dependency has been written in is in the new `record_dependency_update` function at `codemodder.py`

drdavella (approved):

PR #181, opened 2023-12-15
Added exception-without-raise codemod

andrecsilva:
## Overview
Added a codemod that raises exceptions that were created but not raised

### DIFF
--- a/src/codemodder/scripts/generate_docs.py
+++ b/src/codemodder/scripts/generate_docs.py
@@ -162,6 +162,10 @@ If you want to allow those protocols, change the incoming PR to look more like t
         importance="Medium",
         guidance_explained="This change will only restrict the response type and will not alter the response data itself. Thus we deem it safe.",
     ),
+    "exception-without-raise": DocMetadata(
+        importance="Low",
+        guidance_explained="An statement with an exception has no effect. Raising the exception is most likely the intended effect and thus we deem it safe.",
+    ),
 }
 
 
--- a/src/codemodder/scripts/generate_docs.py
+++ b/src/codemodder/scripts/generate_docs.py
@@ -164,7 +164,7 @@ If you want to allow those protocols, change the incoming PR to look more like t
     ),
     "exception-without-raise": DocMetadata(
         importance="Low",
-        guidance_explained="An statement with an exception has no effect. Raising the exception is most likely the intended effect and thus we deem it safe.",
+        guidance_explained="A statement with an exception by itself has no effect. Raising the exception is most likely the intended effect and thus we deem it safe.",
     ),
 }
 
--- a/src/codemodder/utils/utils.py
+++ b/src/codemodder/utils/utils.py
@@ -1,12 +1,22 @@
 import libcst as cst
+from functools import cache
 
 
-def list_subclasses(kls) -> set[str]:
-    l = set()
-    for subkls in kls.__subclasses__():
-        l.add(subkls.__name__)
-        l = l | list_subclasses(subkls)
-    return l
+@cache
+def list_subclasses(base_kls) -> set[str]:
+    all_subclasses = set()
+    stack = [base_kls]
+    while stack:
+        kls = stack.pop()
+        all_subclasses.add(kls)
+        stack.extend(kls.__subclasses__())
+    return all_subclasses
+
+
+def full_qualified_name_from_class(cls) -> str:
+    if cls.__module__ == "builtins":
+        return cls.__qualname__
+    return f"{cls.__module__}.{cls.__qualname__}"
 
 
 def clean_simplestring(node: cst.SimpleString | str) -> str:
--- a/src/codemodder/utils/utils.py
+++ b/src/codemodder/utils/utils.py
@@ -1,6 +1,14 @@
 import libcst as cst
 
 
+def list_subclasses(kls) -> set[str]:
+    l = set()
+    for subkls in kls.__subclasses__():
+        l.add(subkls.__name__)
+        l = l | list_subclasses(subkls)
+    return l
+
+
 def clean_simplestring(node: cst.SimpleString | str) -> str:
     if isinstance(node, str):
         return node.strip('"')
--- a/src/core_codemods/docs/pixee_python_exception-without-raise.md
+++ b/src/core_codemods/docs/pixee_python_exception-without-raise.md
@@ -1,0 +1,9 @@
+An statement with an exception without raising it is most likely an error. Our changes look something like this:
+
+```diff
+try:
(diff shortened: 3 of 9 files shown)


tests/codemods/test_exception_without_raise.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        expected = """\
+        raise ValueError
+        """
+        self.run_and_assert(tmpdir, dedent(input_code), dedent(expected))
+        assert len(self.file_context.codemod_changes) == 1
+
+    def test_alias(self, tmpdir):
+        input_code = """\
+        from libcst import CSTValidationError as error
```
drdavella: This only works because `libcst` is a dependency of codemodder itself, correct? We won't be able to detect custom exception types that are defined within the project under analysis and/or third-party dependencies of that project?

If Sonar is smarter than us then we can (eventually) use their report to fix the issues that we can't ourselves detect.

src/codemodder/utils/utils.py
```diff
@@ -1,6 +1,14 @@
 import libcst as cst
 
 
+def list_subclasses(kls) -> set[str]:
```
drdavella: For `BaseException` this can be known up-front and will be constant for the entire process lifetime. This suggests we should either compute this up-front or do it lazily with a cache. You could use `@functools.lru_cache` but since this is called recursively it might be necessary to do a bit of refactoring.

src/codemodder/utils/utils.py
```diff
@@ -1,6 +1,14 @@
...
+    l = set()
+    for subkls in kls.__subclasses__():
+        l.add(subkls.__name__)
+        l = l | list_subclasses(subkls)
```
drdavella: Very nitpicky but I think you could use `|=` here.

src/core_codemods/docs/pixee_python_exception-without-raise.md
```diff
@@ -0,0 +1,9 @@
+An statement with an exception without raising it is most likely an error. Our changes look something like this:
```
drdavella:
```suggestion
This codemod fixes cases where an exception is referenced by itself in a statement without being raised. This most likely indicates a bug: you probably meant to actually raise the exception. 

Our changes look something like this:
```

src/core_codemods/exception_without_raise.py
```diff
@@ -0,0 +1,41 @@
...
+    def leave_SimpleStatementLine(
+        self,
+        original_node: cst.SimpleStatementLine,
+        updated_node: cst.SimpleStatementLine,
+    ) -> Union[
+        cst.BaseStatement, cst.FlattenSentinel[cst.BaseStatement], cst.RemovalSentinel
+    ]:
+        match original_node:
+            case cst.SimpleStatementLine(
+                body=[cst.Expr(cst.Name() | cst.Attribute() as name)]
```
drdavella, line 32: What about `Call`?

tests/codemods/test_exception_without_raise.py
```diff
@@ -0,0 +1,46 @@
+from tests.codemods.base_codemod_test import BaseCodemodTest
+from core_codemods.exception_without_raise import ExceptionWithoutRaise
+from textwrap import dedent
+
+
+class TestExceptionWithoutRaise(BaseCodemodTest):
```
drdavella, line 6: What about this case?
```python
ValueError("Bad value!")
```

src/codemodder/scripts/generate_docs.py
```diff
@@ -162,6 +162,10 @@ class DocMetadata:
         importance="Medium",
         guidance_explained="This change will only restrict the response type and will not alter the response data itself. Thus we deem it safe.",
     ),
+    "exception-without-raise": DocMetadata(
+        importance="Low",
+        guidance_explained="An statement with an exception has no effect. Raising the exception is most likely the intended effect and thus we deem it safe.",
```
```suggestion
        guidance_explained="A statement with an exception by itself has no effect. Raising the exception is most likely the intended effect and thus we deem it safe.",
```

src/core_codemods/exception_without_raise.py
```diff
@@ -0,0 +1,41 @@
...
+from codemodder.codemods.api import BaseCodemod
+from codemodder.codemods.base_codemod import ReviewGuidance
+
+from codemodder.codemods.utils_mixin import NameResolutionMixin
+from codemodder.utils.utils import list_subclasses
+
+
+class ExceptionWithoutRaise(BaseCodemod, NameResolutionMixin):
+    NAME = "exception-without-raise"
+    SUMMARY = "Added raise statement to exception creation"
```
drdavella:
```suggestion
    SUMMARY = "Ensure bare exception statements are raised"
```

src/core_codemods/exception_without_raise.py
```diff
@@ -0,0 +1,41 @@
...
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = SUMMARY
+    REFERENCES = [
+        {
+            "url": "https://docs.python.org/3/tutorial/errors.html#raising-exceptions",
+            "description": "",
+        },
+    ]
+    CHANGE_DESCRIPTION = "Wrapped exception in a raise statement"
```
drdavella:
```suggestion
    CHANGE_DESCRIPTION = "Raised bare exception statement"
```

drdavella (changes requested):
Looks good overall but I think there might be some other cases we want to consider. Also some small comment and a suggestion for the description.

tests/codemods/test_exception_without_raise.py
andrecsilva: > This only works because libcst is a dependency of codemodder itself, correct? 

Yes and I see that as a feature.

> We won't be able to detect custom exception types that are defined within the project under analysis and/or third-party dependencies of that project?

We can't really detect if a particular name is a class that derives from `BaseException` unless we somehow parse its declaration. This is unlikely to happen as we only use the module itself as context. Also, looking at sonar's implementation of the rule, it seems to only check for builtin exceptions

src/core_codemods/exception_without_raise.py
andrecsilva: Good catch, I forgot about this case.

andrecsilva:
I tweaked the codemod to now match against the fully qualified name to be more precise. This means that it will miss some cases when the class is declared in a different module than what it was imported.
For example: you may import `CSTValidationError` from `libcst` directly, but it is defined in `libcst._nodes.base.CSTValidationError`.

drdavella (approved):

PR #186, opened 2023-12-21
Added literal-or-new-object-identity codemod

andrecsilva:
## Overview
Added a new codemod that replaces `is` and `is not` comparisons with literals and new objects with `==`

### DIFF
--- a/src/core_codemods/docs/pixee_python_literal-or-new-object-identity.md
+++ b/src/core_codemods/docs/pixee_python_literal-or-new-object-identity.md
@@ -1,0 +1,9 @@
+The `is` and `is not` operator will only return `True` when the expression have the same `id`, that is, `a is b` is equivalent to `id(a) == id(b)`. New objects and literals have their own identities and thus shouldn't be compared with using the `is` or `is not` operators.
+
+Our changes look something like this:
+
+```diff
+def foo(l):
+-    return l is [1,2,3]
++    return l == [1,2,3]
+```
--- a/src/core_codemods/docs/pixee_python_literal-or-new-object-identity.md
+++ b/src/core_codemods/docs/pixee_python_literal-or-new-object-identity.md
@@ -1,4 +1,4 @@
-The `is` and `is not` operator will only return `True` when the expression have the same `id`, that is, `a is b` is equivalent to `id(a) == id(b)`. New objects and literals have their own identities and thus shouldn't be compared with using the `is` or `is not` operators.
+The `is` and `is not` operator will only return `True` when the expression have the same `id`. In other words, `a is b` is equivalent to `id(a) == id(b)`. New objects and literals have their own identities and thus shouldn't be compared with using the `is` or `is not` operators.
 
 Our changes look something like this:
 
--- a/src/core_codemods/literal_or_new_object_identity.py
+++ b/src/core_codemods/literal_or_new_object_identity.py
@@ -1,0 +1,63 @@
+import libcst as cst
+from codemodder.codemods.api import BaseCodemod
+from codemodder.codemods.base_codemod import ReviewGuidance
+
+from codemodder.codemods.utils_mixin import NameResolutionMixin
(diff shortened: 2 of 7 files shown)


src/core_codemods/docs/pixee_python_literal-or-new-object-identity.md
```diff
@@ -0,0 +1,9 @@
+The `is` and `is not` operator will only return `True` when the expression have the same `id`, that is, `a is b` is equivalent to `id(a) == id(b)`. New objects and literals have their own identities and thus shouldn't be compared with using the `is` or `is not` operators.
```
drdavella:
```suggestion
The `is` and `is not` operator will only return `True` when the expression have the same `id`. In other words, `a is b` is equivalent to `id(a) == id(b)`. New objects and literals have their own identities and thus shouldn't be compared with using the `is` or `is not` operators.
```

src/core_codemods/literal_or_new_object_identity.py
```diff
@@ -0,0 +1,63 @@
...
+                    "list",
+                    "tuple",
+                    "set",
+                )
+        return False
+
+    def leave_Comparison(
+        self, original_node: cst.Comparison, updated_node: cst.Comparison
+    ) -> cst.BaseExpression:
+        if self.filter_by_path_includes_or_excludes(self.node_position(original_node)):
```
drdavella, line 37: I realized I've forgotten this filter in many of my own recent codemods so I'll need to go back and make updates.

src/core_codemods/literal_or_new_object_identity.py
```diff
@@ -0,0 +1,63 @@
...
+            match original_node:
+                case cst.Comparison(
+                    left=left, comparisons=[cst.ComparisonTarget() as target]
+                ):
+                    if isinstance(target.operator, cst.Is | cst.IsNot):
```
drdavella, line 42: Might be ever so slightly cleaner if you match here? But possibly I'm just obsessed with match right now.

tests/codemods/test_literal_or_new_object_identity.py
```diff
@@ -0,0 +1,174 @@
+from core_codemods.literal_or_new_object_identity import LiteralOrNewObjectIdentity
+from tests.codemods.base_codemod_test import BaseCodemodTest
+from textwrap import dedent
+
+
+class TestLiteralOrNewObjectIdentity(BaseCodemodTest):
```
drdavella, line 6: Do we need test cases with the literal object on the LHS of the expression?

drdavella (approved):
Approving with comments. My main substantive comment/question is about expressions with literals on the LHS.

drdavella:
One additional comment but Sonar calls this out as a non-compliant example as well:
```python
    mylist = []  # mylist is assigned a new object
    param is mylist  # Noncompliant: always False
```
I know it's harder for us to do constant propagation without semgrep but we might want to add a ticket for this.

src/core_codemods/literal_or_new_object_identity.py
andrecsilva: Not really, I'm not using the operator in any place. Using match would add an extra indentation which is why I avoid in those simple cases.

src/core_codemods/literal_or_new_object_identity.py
andrecsilva: Another reason to have another look at the API. This should be automated in these simple cases.

andrecsilva:
> One additional comment but Sonar calls this out as a non-compliant example as well:
> 
> ```python
>     mylist = []  # mylist is assigned a new object
>     param is mylist  # Noncompliant: always False
> ```
> 
> I know it's harder for us to do constant propagation without semgrep but we might want to add a ticket for this.

I'll do this in a future PR. We kinda have the solution to that already but I need to do some testing here.

PR #191, opened 2023-12-22
Do not duplicate existing dependency with different version

drdavella:
## Overview
*Do not duplicate existing dependency with different version*

## Description

* This fixes a regression in the dependency manager
* We should never update dependency versions or add versions that conflict with existing dependencies


andrecsilva (commented):
Shouldn't we check for the version against the one we are trying to add?  What if we are trying to add a higher version than the one already in?

I guess that may break things, but then again so is adding code with dependency to late versions.

drdavella:
@andrecsilva I think we're currently taking the stance that if the dependency is already present then it doesn't need to be added, and we will allow dependabot to manage updates to the dependencies themselves. If we encounter a situation where it becomes necessary to update the version, then we can adjust our logic here.

In either case this fixes a regression since expressing the same dependency twice with different versions will cause most package managers to fail.

andrecsilva (approved):

PR #192, opened 2023-12-22
New codemod: `use-set-literal`

drdavella:
## Overview
*A simple codemod that simplifies set literals*


src/core_codemods/use_set_literal.py
```diff
@@ -0,0 +1,28 @@
...
+            case cst.Name("set"):
+                if self.is_builtin_function(original_node):
+                    match original_node.args:
+                        case [cst.Arg(value=cst.List(elements=elements))]:
+                            # Can't use set literal for empty set
+                            if len(elements) == 0:
+                                self.add_change(original_node, self.CHANGE_DESCRIPTION)
+                                return updated_node.with_changes(args=[])
+
+                            self.add_change(original_node, self.CHANGE_DESCRIPTION)
```
clavedeluna: can move this line to line 20 so it doesn't get called twice

src/core_codemods/use_set_literal.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
+        match original_node.func:
+            case cst.Name("set"):
+                if self.is_builtin_function(original_node):
+                    match original_node.args:
+                        case [cst.Arg(value=cst.List(elements=elements))]:
+                            # Can't use set literal for empty set
+                            if len(elements) == 0:
+                                self.add_change(original_node, self.CHANGE_DESCRIPTION)
```
clavedeluna: should also be able to call `self.report_change(original_node)` which I think is nicer but they're kinda dupes of each other

tests/codemods/test_use_set_literal.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def test_simple(self, tmpdir):
+        original_code = """
+        x = set([1, 2, 3])
+        """
+        expected_code = """
+        x = {1, 2, 3}
+        """
+        self.run_and_assert(tmpdir, original_code, expected_code)
+
```
clavedeluna, line 17: worth adding `assert len(self.file_context.codemod_changes) == 1` (or 0) to each test case

src/core_codemods/use_set_literal.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class UseSetLiteral(BaseCodemod, NameResolutionMixin):
+    NAME = "use-set-literal"
+    SUMMARY = "Use Set Literals Instead of Sets from Lists"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = "Replace sets from lists with set literals"
+    REFERENCES: list = []
+
+    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
```
andrecsilva, line 14: It's lacking the line include/excludes check.

andrecsilva (changes requested):

andrecsilva (approved):

PR #193, opened 2023-12-27
codemod to set `shell=False` for subprocess commands

clavedeluna:
## Overview
*Codemod that checks any of the `subprocess` functions / class instantiations that have a `shell` kwarg and ensures it is set to `True`*

This is the most secure fix, but there are some valid cases for using shell=True hence the review guidance.

### DIFF
--- a/src/codemodder/scripts/generate_docs.py
+++ b/src/codemodder/scripts/generate_docs.py
@@ -179,8 +179,8 @@ If you want to allow those protocols, change the incoming PR to look more like t
         guidance_explained="Since literals and new objects have their own identities, comparisons against them using `is` operators are most likely a bug and thus we deem the change safe.",
     ),
     "subprocess-shell-false": DocMetadata(
-        importance="Medium",
-        guidance_explained="There are valid use cases for `shell=True` such as executing a validated string command or using shell functionality like globs or wildcard.",
+        importance="High",
+        guidance_explained="In most cases setting `shell=False` is correct and leads to much safer code. However there are valid use cases for `shell=True` when using shell functionality like pipes or wildcard is required. In such cases it is important to run only trusted, validated commands.",
     ),
 }
 
--- a/src/codemodder/scripts/generate_docs.py
+++ b/src/codemodder/scripts/generate_docs.py
@@ -178,6 +178,10 @@ If you want to allow those protocols, change the incoming PR to look more like t
         importance="Low",
         guidance_explained="Since literals and new objects have their own identities, comparisons against them using `is` operators are most likely a bug and thus we deem the change safe.",
     ),
+    "subprocess-shell-false": DocMetadata(
+        importance="Medium",
+        guidance_explained="There are valid use cases for `shell=True` such as executing a validated string command or using shell functionality like globs or wildcard.",
+    ),
 }
 
 
--- a/src/core_codemods/docs/pixee_python_subprocess-shell-false.md
+++ b/src/core_codemods/docs/pixee_python_subprocess-shell-false.md
@@ -1,0 +1,11 @@
+This codemod sets the `shell` keyword argument to `False` in `subprocess` module function calls that have set it to `True`.
+
+Setting `shell=True` will execute the provided command through the system shell which can lead to shell injection vulnerabilities.
+
+The changes from this codemod look like this:
+
+```diff
+ import subprocess
+- subprocess.run("echo 'hi'", shell=True)
++ subprocess.run("echo 'hi'", shell=False)
+```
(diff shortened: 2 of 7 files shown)


src/core_codemods/subprocess_shell_false.py
```diff
@@ -0,0 +1,31 @@
...
+        f"subprocess.{func}"
+        for func in {"run", "call", "check_output", "check_call", "Popen"}
+    ]
+
+    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
+        if self.find_base_name(original_node.func) in self.SUBPROCESS_FUNCS:
+            for arg in original_node.args:
+                if matchers.matches(
+                    arg.keyword, matchers.Name("shell")
+                ) and matchers.matches(arg.value, matchers.Name("True")):
```
andrecsilva: Suggestion:
```
matchers.matches(arg, matchers.Arg(keyword = matchers.Name('shell'), value = matchers.Name('True')))
```

src/core_codemods/subprocess_shell_false.py
```diff
@@ -0,0 +1,31 @@
...
+    SUMMARY = "Use `shell=False` in `subprocess` Function Calls"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW
+    DESCRIPTION = "Set `shell` keyword argument to `False`"
+    REFERENCES: list = []
+    SUBPROCESS_FUNCS = [
+        f"subprocess.{func}"
+        for func in {"run", "call", "check_output", "check_call", "Popen"}
+    ]
+
+    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
```
andrecsilva, line 32: It's lacking the include/exclude check.

andrecsilva (changes requested):

tests/codemods/test_subprocess_shell_false.py
```diff
@@ -0,0 +1,54 @@
...
+
+class TestSubprocessShellFalse(BaseCodemodTest):
+    codemod = SubprocessShellFalse
+
+    def test_name(self):
+        assert self.codemod.name() == "subprocess-shell-false"
+
+    @each_func
+    def test_import(self, tmpdir, func):
+        input_code = f"""import subprocess
```
drdavella: Test cases now automatically adjust indentation so this is much more readable if rendered like this:
```python
    input_code = f"""
    import subprocess
    subprocess.{func}(args, shell=True)
    """
```

src/codemodder/scripts/generate_docs.py
```diff
@@ -178,6 +178,10 @@ class DocMetadata:
         importance="Low",
         guidance_explained="Since literals and new objects have their own identities, comparisons against them using `is` operators are most likely a bug and thus we deem the change safe.",
     ),
+    "subprocess-shell-false": DocMetadata(
+        importance="Medium",
```
```suggestion
        importance="High",
```

src/core_codemods/subprocess_shell_false.py
```diff
@@ -0,0 +1,31 @@
...
+from codemodder.codemods.utils_mixin import NameResolutionMixin
+from codemodder.codemods.api.helpers import NewArg
+
+
+class SubprocessShellFalse(BaseCodemod, NameResolutionMixin):
+    NAME = "subprocess-shell-false"
+    SUMMARY = "Use `shell=False` in `subprocess` Function Calls"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW
+    DESCRIPTION = "Set `shell` keyword argument to `False`"
+    REFERENCES: list = []
```
drdavella: Please add the following references:
* https://docs.python.org/3/library/subprocess.html#security-considerations
* https://en.wikipedia.org/wiki/Code_injection#Shell_injection
* https://stackoverflow.com/a/3172488

src/codemodder/scripts/generate_docs.py
```diff
@@ -178,6 +178,10 @@ class DocMetadata:
...
+        guidance_explained="There are valid use cases for `shell=True` such as executing a validated string command or using shell functionality like globs or wildcard.",
```
```suggestion
        guidance_explained="In most cases setting `shell=False` is correct and leads to much safer code. However there are valid use cases for `shell=True` when using shell functionality like pipes or wildcard is required. In such cases it is important to run only trusted, validated commands.",
```

src/core_codemods/docs/pixee_python_subprocess-shell-false.md
```diff
@@ -0,0 +1,11 @@
+This codemod sets the `shell` keyword argument to `False` in `subprocess` module function calls that have set it to `True`.
+
+Setting `shell=True` will execute the provided command through the system shell which can lead to shell injection vulnerabilities.
```
drdavella:
```suggestion
Setting `shell=True` will execute the provided command through the system shell which can lead to shell injection vulnerabilities. In the worst case this can give an attacker the ability to run arbitrary commands on your system. In most cases using `shell=False` is sufficient and leads to much safer code.
```

drdavella (changes requested):
Looks good but I have some requests for updates to metadata.

Also there's the question of whether this should be retroactively applied to our own `security` API.

drdavella (approved):

andrecsilva (approved):

PR #194, opened 2023-12-27
New codemod to remove `global` lines at module level

clavedeluna:
## Overview
*A codemod that removes a line that uses `global` keyword at the module level*

## Additional Details
* Not to be confused with a future codemod that may add or remove a `global` keyword line within non-global scopes if a user missed them

### DIFF
--- a/src/core_codemods/remove_module_global.py
+++ b/src/core_codemods/remove_module_global.py
@@ -1,0 +1,22 @@
+import libcst as cst
+from libcst.metadata import GlobalScope, ScopeProvider
+from typing import Union
+from codemodder.codemods.api import BaseCodemod, ReviewGuidance
+from codemodder.codemods.utils_mixin import NameResolutionMixin
+
+
+class RemoveModuleGlobal(BaseCodemod, NameResolutionMixin):
+    NAME = "remove-module-global"
+    SUMMARY = "Remove Module-level Global Call"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = "Remove Lines with `global` keyword at Module Level"
+    REFERENCES: list = []
+
+    def leave_Global(
+        self, original_node: cst.Global, _
+    ) -> Union[cst.Global, cst.RemovalSentinel,]:
+        scope = self.get_metadata(ScopeProvider, original_node)
+        if isinstance(scope, GlobalScope):
+            self.report_change(original_node)
+            return cst.RemovalSentinel.REMOVE
+        return original_node
--- a/src/core_codemods/remove_module_global.py
+++ b/src/core_codemods/remove_module_global.py
@@ -7,9 +7,9 @@ from codemodder.codemods.utils_mixin import NameResolutionMixin
 
 class RemoveModuleGlobal(BaseCodemod, NameResolutionMixin):
     NAME = "remove-module-global"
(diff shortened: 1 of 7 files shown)


src/core_codemods/remove_module_global.py
```diff
@@ -0,0 +1,22 @@
...
+class RemoveModuleGlobal(BaseCodemod, NameResolutionMixin):
+    NAME = "remove-module-global"
+    SUMMARY = "Remove Module-level Global Call"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = "Remove Lines with `global` keyword at Module Level"
+    REFERENCES: list = []
+
+    def leave_Global(
+        self, original_node: cst.Global, _
+    ) -> Union[cst.Global, cst.RemovalSentinel,]:
```
andrecsilva, line 17: It's lacking the line include/exclude check.

andrecsilva (changes requested):

src/core_codemods/remove_module_global.py
```diff
@@ -0,0 +1,22 @@
+import libcst as cst
+from libcst.metadata import GlobalScope, ScopeProvider
+from typing import Union
+from codemodder.codemods.api import BaseCodemod, ReviewGuidance
+from codemodder.codemods.utils_mixin import NameResolutionMixin
+
+
+class RemoveModuleGlobal(BaseCodemod, NameResolutionMixin):
+    NAME = "remove-module-global"
+    SUMMARY = "Remove Module-level Global Call"
```
drdavella:
```suggestion
    SUMMARY = "Remove Module-level `global` Keyword"
```

src/core_codemods/remove_module_global.py
```diff
@@ -0,0 +1,22 @@
...
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = "Remove Lines with `global` keyword at Module Level"
```
drdavella:
```suggestion
    DESCRIPTION = "Remove `global` keyword at module level"
```

drdavella (changes requested):

src/core_codemods/remove_module_global.py
clavedeluna: so glad you're realizing this bc I clearly definitely forgot it multiple times.

src/core_codemods/remove_module_global.py
clavedeluna: This comment applies to both of your suggestions for wording. Your suggestions are shorter but I think remove the clarity. We aren't just removing the global keyword, but the entire line where global is.
So we are removing `global var`, not just `global` and keeping `var`

If you still prefer your suggestions I will apply them.

src/core_codemods/remove_module_global.py
drdavella: I did too, which is how we noticed it in the first place 😅

src/core_codemods/remove_module_global.py
drdavella: The previous comment fixed the reference to "call" which isn't correct because it's not a call. The description in both places could be "Remove `global` usage at module level".

drdavella (approved):

andrecsilva (approved):

PR #197, opened 2023-12-30
new codemod to combine two startswith/endswith

clavedeluna:
## Overview
*A codemod to combine two `or` calls to `startswith` or `endswith`*

## Description

* This can later on be generalized for multiple `or` calls

### DIFF
--- a/src/codemodder/scripts/generate_docs.py
+++ b/src/codemodder/scripts/generate_docs.py
@@ -194,6 +194,10 @@ If you want to allow those protocols, change the incoming PR to look more like t
         importance="Medium",
         guidance_explained="Breakpoints are generally used only for debugging and can easily be forgotten before deploying code.",
     ),
+    "combine-startswith-endswith": DocMetadata(
+        importance="Low",
+        guidance_explained="Simplifying expressions involving `startswith` or `endswith` calls is safe.",
+    ),
 }
 
 
--- a/src/core_codemods/combine_startswith_endswith.py
+++ b/src/core_codemods/combine_startswith_endswith.py
@@ -1,0 +1,63 @@
+import libcst as cst
+from libcst import matchers as m
+from codemodder.codemods.api import BaseCodemod, ReviewGuidance
+from codemodder.codemods.utils_mixin import NameResolutionMixin
+
+
+class CombineStartswithEndswith(BaseCodemod, NameResolutionMixin):
+    NAME = "combine-startswith-endswith"
+    SUMMARY = "Simplify Boolean Expressions Using `startswith` and `endswith`"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = "Use tuple of matches instead of boolean expression"
+    REFERENCES: list = []
+
+    def leave_BooleanOperation(
+        self, original_node: cst.BooleanOperation, updated_node: cst.BooleanOperation
+    ) -> cst.CSTNode:
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return original_node
+
+        if self.matches_startswith_endswith_or_pattern(original_node):
+            left_call = cst.ensure_type(updated_node.left, cst.Call)
+            right_call = cst.ensure_type(updated_node.right, cst.Call)
+
+            self.report_change(original_node)
+
+            new_arg = cst.Arg(
+                value=cst.Tuple(
+                    elements=[
+                        cst.Element(value=left_call.args[0].value),
+                        cst.Element(value=right_call.args[0].value),
+                    ]
+                )
+            )
+
+            return cst.Call(func=left_call.func, args=[new_arg])
+
+        return updated_node
+
+    def matches_startswith_endswith_or_pattern(
+        self, node: cst.BooleanOperation
+    ) -> bool:
+        # Match the pattern: x.startswith("...") or x.startswith("...")
+        # and the same but with endswith
+        args = [m.Arg(value=m.SimpleString())]
+        startswith = m.Call(
+            func=m.Attribute(value=m.Name(), attr=m.Name("startswith")),
+            args=args,
+        )
+        endswith = m.Call(
+            func=m.Attribute(value=m.Name(), attr=m.Name("endswith")),
+            args=args,
+        )
+        startswith_or = m.BooleanOperation(
+            left=startswith, operator=m.Or(), right=startswith
+        )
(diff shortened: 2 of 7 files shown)


src/core_codemods/combine_startswith_endswith.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class CombineStartswithEndswith(BaseCodemod, NameResolutionMixin):
+    NAME = "combine-startswith-endswith"
+    SUMMARY = "Combine Two `startswith` or `endswith` Calls"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = "Combine two calls to either `startswith` or `endswith` joined by an `or` statement."
+    REFERENCES: list = []
+
+    def leave_BooleanOperation(
```
andrecsilva, line 14: Lacking line include/exclude check.

src/core_codemods/combine_startswith_endswith.py
```diff
@@ -0,0 +1,49 @@
...
+        return updated_node
+
+    def matches_startswith_endswith_or_pattern(
+        self, node: cst.BooleanOperation
+    ) -> bool:
+        # Match the pattern: x.startswith("...") or x.startswith("...")
+        # and the same but with endswith
+        call = m.Call(
+            func=m.Attribute(
+                value=m.Name(), attr=m.Name("startswith") | m.Name("endswith")
```
andrecsilva: Those will match:
```python
x.startswith('foo') or x.endswith('bar')
...
x.startswith('foo') or y.startswith('bar')
```
We don't want that.

andrecsilva (changes requested):

src/core_codemods/combine_startswith_endswith.py
```diff
@@ -38,12 +43,19 @@ def matches_startswith_endswith_or_pattern(
...
+            args=args,
+        )
+        startswith_or = m.BooleanOperation(
+            left=startswith, operator=m.Or(), right=startswith
         )
-        return m.matches(
-            node, m.BooleanOperation(left=call, operator=m.Or(), right=call)
+        endswith_or = m.BooleanOperation(left=endswith, operator=m.Or(), right=endswith)
+        return m.matches(node, startswith_or | endswith_or) and m.matches(
+            node.left.func.value, node.right.func.value
```
andrecsilva: When using `matchers.matches` you should use the mock nodes from the `matchers` module as recommended by the docs:
https://libcst.readthedocs.io/en/latest/matchers.html#libcst.matchers.matches
Suggestion:
```python
node.left.func.value.value == node.right.func.value.value
```

andrecsilva (changes requested):

andrecsilva (approved):

src/core_codemods/docs/pixee_python_combine-startswith-endswith.md
```diff
@@ -0,0 +1,11 @@
+This codemod updates places where two separate calls to either `startswith` or `endswith` are joined by an `or` statement.
```
drdavella:
```suggestion
Many developers are not necessarily aware that the `startswith` and `endswith` methods of `str` objects can accept a tuple of strings to match. This means that there is a lot of code that uses boolean expressions such as `x.startswith('foo') or x.startswith('bar')` instead of the simpler expression `x.startswith(('foo', 'bar'))`.

This codemod simplifies the boolean expressions where possible which leads to cleaner and more concise code.
```

src/core_codemods/combine_startswith_endswith.py
```diff
@@ -0,0 +1,63 @@
+import libcst as cst
+from libcst import matchers as m
+from codemodder.codemods.api import BaseCodemod, ReviewGuidance
+from codemodder.codemods.utils_mixin import NameResolutionMixin
+
+
+class CombineStartswithEndswith(BaseCodemod, NameResolutionMixin):
+    NAME = "combine-startswith-endswith"
+    SUMMARY = "Combine Two `startswith` or `endswith` Calls"
```
drdavella:
```suggestion
    SUMMARY = "Simplify Boolean Expressions Using `startswith` and `endswith`"
```

src/core_codemods/combine_startswith_endswith.py
```diff
@@ -0,0 +1,63 @@
...
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = "Combine two calls to either `startswith` or `endswith` joined by an `or` statement."
```
drdavella:
```suggestion
    DESCRIPTION = "Use tuple of matches instead of boolean expression"
```

src/core_codemods/docs/pixee_python_combine-startswith-endswith.md
```diff
@@ -0,0 +1,11 @@
+This codemod updates places where two separate calls to either `startswith` or `endswith` are joined by an `or` statement. 
+The replacement code takes advantage of the fact that these two methods can accept a tuple as an argument.
```
drdavella:
```suggestion
```

src/codemodder/scripts/generate_docs.py
```diff
@@ -194,6 +194,10 @@ class DocMetadata:
         importance="Medium",
         guidance_explained="Breakpoints are generally used only for debugging and can easily be forgotten before deploying code.",
     ),
+    "combine-startswith-endswith": DocMetadata(
+        importance="Low",
+        guidance_explained="Combining two `startswith` or `endswith` calls is safe and cleans up code.",
```
```suggestion
        guidance_explained="Simplifying expressions involving `startswith` or `endswith` calls is safe.",
```

drdavella (changes requested):
Looks great but requesting some updates on metadata.

clavedeluna:
@drdavella once ou approve you can add it to the merge q

drdavella (approved):

drdavella (changes requested):
@clavedeluna can you please rebase this PR to get rid of the upstream merge from `main` and squash the commit history?

clavedeluna:
@drdavella done. is there a way to make the merge queue squash commits? IIRC the idea for using it was to avoid having to do manual rebase so frequently, but if we use the GH features for committing then it defeats the purpose.

drdavella:
@clavedeluna honestly I'm not sure what it would have done in this case but it's probably just best to avoid merging upstream branches into topic branches to avoid any non-linear history.

drdavella (approved):

PR #204, opened 2024-01-04
new setupcfg writer

clavedeluna:
## Overview
*Adds a setup.cfg file writer for new dependencies*

## Description

* This writer relies on `configparser` only to detect the existing dependencies in the setup.cfg file. We can't rely on it to update and write to file because `configparser` doesn't retain some formatting and comment lines.
* I tried some other parsers to see if they could work, but none did. Instead, this is an attempt to manually write the new dependencies to file with some formatting heuristics and as much reasonable error catching as I could think.

### DIFF
--- a/src/codemodder/dependency_management/setupcfg_writer.py
+++ b/src/codemodder/dependency_management/setupcfg_writer.py
@@ -1,0 +1,124 @@
+import configparser
+
+from typing import Optional
+from codemodder.dependency import Dependency
+from codemodder.change import ChangeSet
+from codemodder.dependency_management.base_dependency_writer import DependencyWriter
+from codemodder.diff import create_diff_and_linenums
+from codemodder.logging import logger
+
+
+import re
+
+
+def find_leading_whitespace(s):
+    match = re.match(r"(\s+)", s)
+    if match:
+        return match.group(1)
+    return ""
+
+
+def added_line_nums_strategy(lines, i):
+    return lines[i]
+
+
+class SetupCfgWriter(DependencyWriter):
+    def add_to_file(
+        self, dependencies: list[Dependency], dry_run: bool = False
+    ) -> Optional[ChangeSet]:
+        config = configparser.ConfigParser()
+
+        try:
+            config.read(self.path)
+        except configparser.ParsingError:
+            logger.debug("Unable to parse setup.cfg file.")
+            return None
+
+        if "options" not in config or not (
+            defined_dependencies := config["options"].get("install_requires", "")
+        ):
+            logger.debug("Unable to add dependencies to setup.cfg file.")
+            return None
+
+        with open(self.path, "r", encoding="utf-8") as f:
+            original_lines = f.readlines()
+
+        new_lines = self.build_new_lines(
+            original_lines, defined_dependencies, dependencies
+        )
+        if not new_lines:
+            logger.debug("Unable to add dependencies to setup.cfg file.")
+            return None
+
+        if not dry_run:
+            try:
+                with open(self.path, "w", encoding="utf-8") as f:
+                    f.writelines(new_lines)
+            except Exception:
+                logger.debug("Unable to add dependencies to setup.cfg file.")
+                return None
+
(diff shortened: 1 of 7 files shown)


clavedeluna:
I made this PR ready to review despite the codecov failure which is valid but due to my attempt to catch exceptions. After review I"ll go back and add some testing or skip coverage in certain lines.

andrecsilva (approved):

drdavella (commented):
Just curious but did you investigate whether this project would help simplify the implementation? https://github.com/Preocts/commented-configparser

clavedeluna:
> Just curious but did you investigate whether this project would help simplify the implementation? https://github.com/Preocts/commented-configparser

I tested with commented-configparser and am very slightly disappointed that it's only a tiny bit buggy. It does retain the comments in most cases, but in some cases it moves comments to different lines. It also added some trailing spaces to the new setup.cfg file. So here's a diff:

```diff
 #comment
 [metadata] 
name = my_package
 version = attr: my_package.VERSION
 author = Josiah Carberry
 author_email = josiah_carberry@brown.edu
 description = My package description
 long_description = file: README.rst, CHANGELOG.rst, LICENSE.rst
 keywords = one, two
 # comment should stay in ln 10
 license = BSD-3-Clause
-classifiers =
+classifiers =
+# comment should stay in ln 16
        Framework :: Django
        Programming Language :: Python :: 3

-# comment should stay in ln 16
```
changes are:
1. added a leading spaces after classifiers =
2. moved comment in ln 16 to ln13, while not changing comment in ln1

@drdavella lmk if you want to accept these changes and I can switch to using commented config or we can keep code as is

drdavella:
@clavedeluna thanks for the info. I'm sorry it didn't work out of the box. Let's accept this as-is for now and we can revisit later since I don't think this is the most critical work for us right now. One other possibility down the road would be to vendor in that third-party code (assuming a permissive license) and maybe tweak it to get the behavior we need.

src/codemodder/dependency_management/setupcfg_writer.py
```diff
@@ -0,0 +1,124 @@
...
+        self,
+        original_lines: list[str],
+        defined_dependencies: str,
+        dependencies_to_add: list[Dependency],
+    ) -> Optional[list[str]]:
+        """
+        configparser does not retain formatting or comment lines, so we have to build
+        the output newline manually.
+        """
+        clean_lines = list(map(lambda s: s.strip(), original_lines))
```
drdavella: I like `map` but genuinely curious whether there's any benefit to doing it this way rather than using a comprehension?
```python
clean_lines = [s.strip for s in original_lines]
```

tests/dependency_management/test_setupcfgt_writer.py
```diff
@@ -0,0 +1,302 @@
...
+        install_requires =
+            requests
+            importlib-metadata; python_version<"3.8"
+            security~=1.2.0
+    """
+
+    assert setup_cfg.read() == dedent(updated_setupcfg)
+
+
+def test_dont_add_existing_dependency(tmpdir):
```
drdavella, line 145: We should have another test case where the version of `security` that already exists in the package is different from the one being added. There should be no new dependency added (or version changed) in this case either.

drdavella (approved):
LGTM but it would be good to add the test case mentioned below.

src/codemodder/dependency_management/setupcfg_writer.py
clavedeluna: no other benefit to map, I think the moment I wrote it that's the first thing that came to me. I'll switch it out

tests/dependency_management/test_setupcfgt_writer.py
clavedeluna: This is a really good test case to think about. None of the other writers handle for this either so if it's ok with you it would be good to handle it in a separate PR for all. It would be worth thinking about it a bit more:  do we really not want to update the version? What if the project has `pkg==1.0` but our recommendation is `pkg==2.0`?

Worth thinking about it independently.

PR #205, opened 2024-01-05
harden-pyyaml codemod now also handles inheritance

clavedeluna:
## Overview
*Update harden-pyyaml codemod to handle cases when a custom class inherits an unsafe loader class*

### DIFF
--- a/src/core_codemods/harden_pyyaml.py
+++ b/src/core_codemods/harden_pyyaml.py
@@ -1,3 +1,5 @@
+from typing import Union
+import libcst as cst
 from codemodder.codemods.base_codemod import ReviewGuidance
 from codemodder.codemods.api import SemgrepCodemod
 from codemodder.codemods.utils_mixin import NameResolutionMixin
@@ -6,8 +8,8 @@ from codemodder.codemods.utils_mixin import NameResolutionMixin
 class HardenPyyaml(SemgrepCodemod, NameResolutionMixin):
     NAME = "harden-pyyaml"
     REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
-    SUMMARY = "Use SafeLoader in `yaml.load()` Calls"
-    DESCRIPTION = "Ensures all calls to yaml.load use `SafeLoader`."
+    SUMMARY = "Replace unsafe `pyyaml` loader with `SafeLoader`"
+    DESCRIPTION = "Replace unsafe `pyyaml` loader with `SafeLoader` in calls to `yaml.load` or custom loader classes."
     REFERENCES = [
         {
             "url": "https://owasp.org/www-community/vulnerabilities/Deserialization_of_untrusted_data",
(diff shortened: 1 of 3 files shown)


src/core_codemods/harden_pyyaml.py
```diff
@@ -50,18 +53,70 @@ def rule(cls):
...
-            ),
-        ]
-        return self.update_arg_target(updated_node, new_args)
+    def on_result_found(
+        self,
+        original_node: Union[cst.Call, cst.ClassDef],
+        updated_node: Union[cst.Call, cst.ClassDef],
+    ):
+        match original_node:
+            case cst.Call():
```
drdavella, line 77: It would actually be useful to provide different change descriptions for each of the cases here but I'm not quite sure that `on_result_found` as currently written enables us to do that. At the very least can you add a TODO here?

src/core_codemods/harden_pyyaml.py
```diff
@@ -50,18 +53,70 @@ def rule(cls):
...
+
+    def _update_bases(self, original_node: cst.ClassDef) -> list[cst.Arg]:
+        new = []
+        unsafe_name_matchers = (
+            matchers.Name(value="UnsafeLoader")
+            | matchers.Name(value="Loader")
+            | matchers.Name(value="BaseLoader")
+            | matchers.Name(value="FullLoader")
+        )
+        for base_arg in original_node.bases:
```
drdavella, line 103: This logic isn't quite correct. Try adding another test case with this as the input code:
```python
        from abc import ABC
        import yaml as yam
        from whatever import Loader

        class MyCustomLoader(ABC, Loader, yam.UnsafeLoader):
            def __init__(self, *args, **kwargs):
                super(MyCustomLoader, self).__init__(*args, **kwargs)
```
The `Loader` base in this case should not be replaced.

src/core_codemods/harden_pyyaml.py
```diff
@@ -50,18 +53,70 @@ def rule(cls):
...
+            if matchers.matches(base_arg.value, unsafe_name_matchers):
+                self.add_needed_import(self._module_name, "SafeLoader")
+                self.remove_unused_import(base_arg.value)
+                base_arg = base_arg.with_changes(
+                    value=base_arg.value.with_changes(value="SafeLoader")
+                )
+            if matchers.matches(
```
drdavella: This condition seems not specific enough: wouldn't anything with `yaml.` match here? This is a little artificial but:
```python
from yaml import UnsafeLoader
import whatever as yaml

class MyLoader(UnsafeLoader, yaml.Loader):
    ...
```

drdavella (changes requested):

src/core_codemods/harden_pyyaml.py
clavedeluna: good cal, I assumed the semgrep filtering wouldn't lead to this but because there is one finding at least it does. I'll refine

src/core_codemods/harden_pyyaml.py
```diff
@@ -50,18 +53,80 @@ def rule(cls):
...
+                new.append(base_arg)
+                continue
+
+            if matchers.matches(base_arg.value, unsafe_name_matchers):
+                self.add_needed_import(self._module_name, "SafeLoader")
+                self.remove_unused_import(base_arg.value)
+                base_arg = base_arg.with_changes(
+                    value=base_arg.value.with_changes(value="SafeLoader")
+                )
+            if matchers.matches(
```
drdavella: It took me awhile to understand why this code works, even though the tests are passing. I think this can all be simplified a bit more for clarity:
```python
for base_arg in original_node.bases:
    # Ensures that we only consider valid names
    base_name = self.find_base_name(base_arg.value)
    if base_name not in base_names:
        new.append(base_arg)
        continue

    match base_arg.value:
        case cst.Name():
            self.add_needed_import(self._module_name, "SafeLoader")
            self.remove_unused_import(base_arg.value)
            base_arg = base_arg.with_changes(
                 value=base_arg.value.with_changes(value="SafeLoader")
            )
        case cst.Attribute():
            base_arg = base_arg.with_changes(
                value=base_arg.value.with_changes(attr=cst.Name("SafeLoader"))
            )

    new.append(base_arg)
```

drdavella (changes requested):
Looks good but one more suggestion for simplification/clarity.

andrecsilva (approved):

drdavella (approved):
LGTM just need to fix the pylint failure.

PR #206, opened 2024-01-08
codemod to deprecate logging.warn

clavedeluna:
## Overview
*codemod that will fix `logging.warn` or `{logger}.warn` to `....warning`*

### DIFF
--- a/src/core_codemods/docs/pixee_python_fix-deprecated-logging-warn.md
+++ b/src/core_codemods/docs/pixee_python_fix-deprecated-logging-warn.md
@@ -1,0 +1,13 @@
+The `warn` method from `logging` has been [deprecated](https://docs.python.org/3/library/logging.html#logging.Logger.warning) since Python 3.3. 
+
+Our changes look like the following:
+```diff
+ import logging
+
+- logging.warn("hello")
++ logging.warning("hello")
+ ...
+ log = logging.getLogger("my logger")
+- log.warn("hello")
++ log.warning("hello") 
+```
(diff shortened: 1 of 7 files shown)


andrecsilva (approved):

src/core_codemods/docs/pixee_python_fix-deprecated-logging-warn.md
```diff
@@ -0,0 +1,13 @@
+The `warn` method from `logging` has been [deprecated](https://docs.python.org/3/library/logging.html#logging.Logger.warning) since Python 3.3.
```
drdavella:
```suggestion
The `warn` method from `logging` has been [deprecated](https://docs.python.org/3/library/logging.html#logging.Logger.warning) in favor of `warning` since Python 3.3. Since the old method `warn` has been retained for a long time, there are a lot of developers that are unaware of this change and consequently a lot of code using the older method.
```

tests/codemods/test_fix_deprecated_logging_warn.py
```diff
@@ -0,0 +1,87 @@
...
+        [
+            (
+                """from logging import warn as warn_func
+warn_func('something')""",
+                """from logging import warning
+warning('something')""",
+            ),
+            (
+                """from logging import getLogger as make_logger
+logger = make_logger('anything')
```
drdavella: I think the `dedent` should apply to this as well, which means we could have more consistent indentation. Not sure though.

tests/codemods/test_fix_deprecated_logging_warn.py
```diff
@@ -0,0 +1,87 @@
+import pytest
+from core_codemods.fix_deprecated_logging_warn import FixDeprecatedLoggingWarn
+from tests.codemods.base_codemod_test import BaseSemgrepCodemodTest
+
+
+class TestFixDeprecatedLoggingWarn(BaseSemgrepCodemodTest):
```
drdavella, line 6: It would be good to have another test with a call to `warn` that includes additional arguments just to make sure we preserve those arguments in the rewrite.
```python
logger.warn("something %s and %s", foo, bar)
```

Also I'm curious whether this works in this case:
```python
import logger
log = logging.getLogger('foo')

def some_function(logger):
    logger.warn("hi")
    
some_function(log)
```

It might be good to add the test and we can just xfail it for now if it doesn't work.

drdavella (changes requested):
Looks really good overall, just a few small requests.

drdavella (approved):

PR #208, opened 2024-01-09
add include/exclude check to codemods that did not have it

clavedeluna:
## Overview
* some non-semgrep codemods missed the include/exclude check*

## Description
* Right now any codemod that inherits directly from BaseCodemod is susceptible to being implemented without this check because the API doesn't enforce it. 
* It's hard right now to enforce adding this check because a codemod could implement any arbitrary number of `leave_Node` methods
* I tried for a reasonable amount of time to add a pytest check to alert us if we wrote a codemod that did not pass thru `filter_by_path_includes_or_excludes`, but this is harder than it looks. I was hoping to add it to the existing BaseTest class so any new codemods would benefit from it. I have some basic code that doesn't work very well, so maybe I can iterate on it later on.

### DIFF
--- a/src/core_codemods/django_receiver_on_top.py
+++ b/src/core_codemods/django_receiver_on_top.py
@@ -23,11 +23,8 @@ class DjangoReceiverOnTop(BaseCodemod, NameResolutionMixin):
     ) -> Union[
         cst.BaseStatement, cst.FlattenSentinel[cst.BaseStatement], cst.RemovalSentinel
     ]:
-        if not self.filter_by_path_includes_or_excludes(
-            self.node_position(original_node)
-        ):
-            return original_node
-
+        # TODO: add filter by include or exclude that works for nodes
+        # that that have different start/end numbers.
         maybe_receiver_with_index = None
         for i, decorator in enumerate(original_node.decorators):
             true_name = self.find_base_name(decorator.decorator)
--- a/src/core_codemods/django_receiver_on_top.py
+++ b/src/core_codemods/django_receiver_on_top.py
@@ -23,6 +23,11 @@ class DjangoReceiverOnTop(BaseCodemod, NameResolutionMixin):
     ) -> Union[
         cst.BaseStatement, cst.FlattenSentinel[cst.BaseStatement], cst.RemovalSentinel
     ]:
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return original_node
+
         maybe_receiver_with_index = None
         for i, decorator in enumerate(original_node.decorators):
             true_name = self.find_base_name(decorator.decorator)
--- a/src/core_codemods/fix_deprecated_abstractproperty.py
+++ b/src/core_codemods/fix_deprecated_abstractproperty.py
@@ -22,7 +22,7 @@ class FixDeprecatedAbstractproperty(BaseCodemod, NameResolutionMixin):
         if not self.filter_by_path_includes_or_excludes(
             self.node_position(original_node)
         ):
-            return original_node
+            return updated_node
 
         if (
             base_name := self.find_base_name(original_node.decorator)
--- a/src/core_codemods/fix_deprecated_abstractproperty.py
+++ b/src/core_codemods/fix_deprecated_abstractproperty.py
@@ -19,6 +19,11 @@ class FixDeprecatedAbstractproperty(BaseCodemod, NameResolutionMixin):
     def leave_Decorator(
         self, original_node: cst.Decorator, updated_node: cst.Decorator
     ):
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return original_node
+
         if (
             base_name := self.find_base_name(original_node.decorator)
         ) and base_name == "abc.abstractproperty":
--- a/src/core_codemods/https_connection.py
+++ b/src/core_codemods/https_connection.py
@@ -24,7 +24,12 @@ class HTTPSConnectionModifier(ImportedCallModifier[Set[str]]):
         return new_args
 
     def update_attribute(self, true_name, original_node, updated_node, new_args):
-        del true_name, original_node
+        del true_name
+        if not self.filter_by_path_includes_or_excludes(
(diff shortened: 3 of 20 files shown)


src/core_codemods/https_connection.py
```diff
@@ -24,7 +24,12 @@ def updated_args(self, original_args):
         return new_args
 
     def update_attribute(self, true_name, original_node, updated_node, new_args):
-        del true_name, original_node
+        del true_name
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return original_node
+
```
andrecsilva: The filtering call here is not needed as it is covered by the `leave_Call` in the parent class.

src/core_codemods/https_connection.py
```diff
@@ -34,6 +39,11 @@ def update_attribute(self, true_name, original_node, updated_node, new_args):
 
     def update_simple_name(self, true_name, original_node, updated_node, new_args):
         del true_name
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return original_node
+
```
andrecsilva: Same as before.

src/core_codemods/use_walrus_if.py
```diff
@@ -134,6 +139,11 @@ def leave_If(self, original_node, updated_node):
 
     def leave_Assign(self, original_node: cst.Assign, updated_node: cst.Assign):
         del updated_node
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return original_node
+
```
andrecsilva: As we discussed before, we only filter the nodes that "anchor" the changes. In this case the `If` node is the one being used for this purpose. Note that the change is reported with the `If` node position. Thus you don't need this check.

src/core_codemods/use_walrus_if.py
```diff
@@ -147,6 +157,11 @@ def leave_SimpleStatementLine(self, original_node, updated_node):
 
         This feels like a bug in libCST but we'll work around it for now.
         """
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return original_node
+
```
andrecsilva: Same as the other comment.

andrecsilva (changes requested):

src/core_codemods/django_receiver_on_top.py
```diff
@@ -23,6 +23,11 @@ def leave_FunctionDef(
     ) -> Union[
         cst.BaseStatement, cst.FlattenSentinel[cst.BaseStatement], cst.RemovalSentinel
     ]:
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return original_node
+
```
clavedeluna: in attempting to add a unit test here I discovered that this include/exclude check will not work because a function def node has a different start/end numbers. Here's an example:

For code
```
        from django.dispatch import receiver

        @csrf_exempt
        @receiver(request_finished)
        def foo():
            pass
```
If a user wanted to exclude a specific line here, say they pass in to exclude ln5 which is line `def foo():` (not even the line for the decorator), when we hit  `leave_FunctionDef`, we need to match `pos.start.line == line and pos.end.line == line`. However, the position for this leave_FunctionDef node is `CodeRange(start=CodePosition(line=5, column=0), end=CodePosition(line=6, column=8))` (ends at ln6)

This is evidence that we need much more robust include/exclude behavior for these types of cases that are more subtle. I'm not quite prepared to implement this yet. @drdavella thoughts? Should I remove this include/exclude check for this (and any other case I find has this issue) for now?

drdavella: @clavedeluna yes you can remove those checks for now but please make sure to add TODOs. None of our tooling is currently taking advantage of line-based filtering as far as I'm aware so the impact should be minimal for now.

src/core_codemods/fix_deprecated_abstractproperty.py
```diff
@@ -19,6 +19,11 @@ class FixDeprecatedAbstractproperty(BaseCodemod, NameResolutionMixin):
     def leave_Decorator(
         self, original_node: cst.Decorator, updated_node: cst.Decorator
     ):
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return original_node
```
andrecsilva: Sorry for not realizing this before, but we should return `updated_node` here just in case some node within the `original_node` subtree was changed. Otherwise we may revert those previous changes.

I don't think it will affect the current codemods that contains this snippet, but this will probably be copy/pasted a lot, so it should include the "safe" version.

src/core_codemods/remove_unnecessary_f_str.py
```diff
@@ -34,6 +34,11 @@ def _check_formatted_string(
         _original_node: cst.FormattedString,
         updated_node: cst.FormattedString,
     ):
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(_original_node)
+        ):
+            return _original_node
```
andrecsilva: Same

src/core_codemods/use_generator.py
```diff
@@ -25,6 +25,11 @@ class UseGenerator(BaseCodemod, NameResolutionMixin):
     ]
 
     def leave_Call(self, original_node: cst.Call, updated_node: cst.Call):
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return original_node
```
andrecsilva: Same

andrecsilva (changes requested):

src/core_codemods/fix_deprecated_abstractproperty.py
clavedeluna: I noticed we don't always return updated_node like the docs indicate we should. I kinda wanted this to read "if we are excluding this node bc of line, don't change". so returning original_node reads "no change happened" but  it's true that's not always the case.

andrecsilva (approved):

drdavella (approved):

PR #209, opened 2024-01-10
Prepend builtins base name

andrecsilva:
## Overview
find_base_name will now prepend builtins for builtin names.
~~A caveat of this change is that `libcst` considers any name that is not imported or originates in the module as builtin. This means that in:~~
```python
foo()
```
~~`foo` is considered builtin.~~
Disregard this caveat, it's actually working correctly. I assume that it was behaving like this because of a test case I had to change.

### DIFF
--- a/src/codemodder/codemods/utils_mixin.py
+++ b/src/codemodder/codemods/utils_mixin.py
@@ -35,6 +35,9 @@ class NameResolutionMixin(MetadataDependent):
                         alias.evaluated_name,
                     ):
                         return self.base_name_for_import(import_node, alias)
+            case BuiltinAssignment():
+                return "builtins." + node.value
+
         return node.value
 
     def find_base_name(self, node):
--- a/src/codemodder/utils/utils.py
+++ b/src/codemodder/utils/utils.py
@@ -14,8 +14,6 @@ def list_subclasses(base_kls) -> set[str]:
 
 
 def full_qualified_name_from_class(cls) -> str:
-    if cls.__module__ == "builtins":
-        return cls.__qualname__
     return f"{cls.__module__}.{cls.__qualname__}"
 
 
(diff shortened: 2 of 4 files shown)


drdavella (commented):
@andrecsilva can you explain the motivation for this change? Given the caveat that you mentioned, I am not entirely convinced that this is the behavior we want.

andrecsilva:
> @andrecsilva can you explain the motivation for this change? Given the caveat that you mentioned, I am not entirely convinced that this is the behavior we want.

The goal is to be closer to the `__module__ . __qualname__` naming pattern for Python objects. For builtin objects/functions, the `__module__` equals `builtins`. 
Until now, `find_base_name` was simply returning the function name for builtin functions. This means you need an additional check if the function is builtin. For example, if you had locally redefined a builtin name, say, `print`, `find_base_name` would still return `print` despite not being builtin.
~~The caveat is not a problem as long as you're acting on trees that comes from compilable python files and not small snippets of it. Even worse, the caveat still exists even without those changes (any name without import/local definition will receive the `BuiltinAssignment` from the `ScopeProvider`).~~
~~If you think this is a problem, I can add an extra check for inclusion in `dir(__builtins__)`.~~
The above is wrong, see the edit to the description and comment below.

clavedeluna:
Given the caveat, my personal preference would be to not accept this change.

andrecsilva:
> Given the caveat, my personal preference would be to not accept this change.

I'm happy to say that I was mistaken and it is actually working as intended. You can check for yourself with the following small example:
@drdavella, @clavedeluna 
```python
import libcst as cst
from libcst.codemod import Codemod, CodemodContext
from codemodder.codemods.utils_mixin import NameResolutionMixin


class A(Codemod, NameResolutionMixin):
    def transform_module_impl(self, tree):
        node = tree.body[0].body[0].value
        print(node)
        print(self.is_builtin_function(node))
        print(self.find_base_name(node))
        return tree

tree = cst.parse_module("foo()")
A(CodemodContext()).transform_module(tree)

```

clavedeluna (approved):
nice! 👏

drdavella (approved):

PR #211, opened 2024-01-11
New codemod: enable global CSRF protection for Flask apps

andrecsilva:
## Overview
Adds a new codemod that enables global CSRF protection for Flask apps.

### DIFF
--- a/src/codemodder/dependency.py
+++ b/src/codemodder/dependency.py
@@ -37,7 +37,7 @@ License: [{self._license.name}]({self._license.url}) ✅ \
         return hash(self.requirement)
 
 
-FLaskWTF = Dependency(
+FlaskWTF = Dependency(
     Requirement("flask-wtf~=1.2.0"),
     description="""\
             This package integrates WTForms into Flask. WTForms provides data validation and and CSRF protection which helps harden applications.
--- a/src/codemodder/dependency.py
+++ b/src/codemodder/dependency.py
@@ -37,6 +37,19 @@ License: [{self._license.name}]({self._license.url}) ✅ \
         return hash(self.requirement)
 
 
+FLaskWTF = Dependency(
+    Requirement("flask-wtf~=1.2.0"),
+    description="""\
+            This package integrates WTForms into Flask. WTForms provides data validation and and CSRF protection which helps harden applications.
+""",
+    _license=License(
+        "BSD-3-Clause",
+        "https://opensource.org/license/BSD-3-clause/",
+    ),
+    oss_link="https://github.com/wtforms/flask-wtf/",
+    package_link="https://pypi.org/project/Flask-WTF/",
+)
+
 DefusedXML = Dependency(
     Requirement("defusedxml~=0.7.1"),
     description="""\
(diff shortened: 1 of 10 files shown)


tests/codemods/test_flask_enable_csrf_protection.py
```diff
@@ -0,0 +1,97 @@
...
+        input_code = """\
+        from flask import Flask
+        app = Flask(__name__)
+        """
+        expected = """\
+        from flask import Flask
+        from flask_wtf.csrf import CSRFProtect
+
+        app = Flask(__name__)
+        csrf_app = CSRFProtect(app)
```
drdavella, line 22: ~If we don't change all subsequent references to `app`, then this change actually has no effect, right?~ Actually this doesn't appear to be true if I understand the docs correctly.

src/core_codemods/docs/pixee_python_flask-enable-csrf-protection.md
```diff
@@ -0,0 +1,10 @@
+Flask views using `FlaskForm` have CSRF protection enabled by default. However other views may use AJAX to perform unsafe HTTP methods. FlaskWTF provides a way to enable CSRF protection globally for all views of a Flask app.
```
drdavella, line 3: This is a case where I think we need a fairly detailed description to educate the user on CSRF vulnerabilities and the benefits of this change. I think we also need a pretty detailed description of the risks of such a change.

src/codemodder/dependency.py
```diff
@@ -37,6 +37,19 @@ def __hash__(self):
         return hash(self.requirement)
 
 
+FLaskWTF = Dependency(
```
drdavella: Small thing but `Flask` instead of `FLask`.

drdavella (commented):
The codemod itself looks good but I am honestly a little unsure about the implications (security and otherwise) of such a change. It seems like this is likely to break any existing web forms that aren't using CSRF validation. Also if I understand the docs correctly it looks like this requires a configuration update in the Flask app in order to set a secret key. If this is done improperly we might cause more problems than we solve.

In short: there might be an interesting codemod here but it probably involves more than just adding a dependency and enabling the CSRF protection plugin.

andrecsilva:
> The codemod itself looks good but I am honestly a little unsure about the implications (security and otherwise) of such a change. It seems like this is likely to break any existing web forms that aren't using CSRF validation. Also if I understand the docs correctly it looks like this requires a configuration update in the Flask app in order to set a secret key. If this is done improperly we might cause more problems than we solve.
> 
> In short: there might be an interesting codemod here but it probably involves more than just adding a dependency and enabling the CSRF protection plugin.

I share your concerns about breakage and spent some time researching and thinking about this. However it is unlikely we are able to fix the broken views ourselves. They usually require the addition of a hidden field within the templates. We can't really parse the templates and they may be scattered into other modules.

It's a half-step solution mostly. Ultimately, I thought it was worth a codemod since this either reduces effort for the full solution or brings attention of CSRF issues to unaware folks. Maybe the documentation should reflect this.

About the configuration, the secret key is, by default, the same as the flask secret key. It may be the case that the user needs to configure the secret_key within Flask themselves. I'm not confident that we should automatically include/suggest one though.

drdavella:
@andrecsilva I think I'm okay with accepting this codemod as-is for now but I'm not sure I'll enable it in the platform quite yet.

It would be very difficult to implement, but also very impressive, if we could automatically fix the affected forms.

drdavella (approved):

PR #212, opened 2024-01-15
Codemod to simplify checks against empty sequences

clavedeluna:
## Overview
*Codemod that changes expressions like `x =! []` or `x == ()` in if, assert, assignment, or standalone statements*

## Description

* This codemod is inspired by a similar [pylint rule](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/use-implicit-booleaness-not-comparison.html)
* I added as many tests as I could come up with and inspired some by pylint. For now I left the test cases for multiple comparisons in one statement as xfailed because they're a little bit tricker than it looks and I wanted to go ahead and get this out.


src/core_codemods/fix_empty_sequence_comparison.py
```diff
@@ -0,0 +1,67 @@
...
+                                        operator=cst.Not(),
+                                        expression=comp_var,
+                                    )
+                                )
+
+        return original_node
+
+    def _is_empty_sequence(self, node: cst.BaseExpression):
+        match node:
+            case cst.List(elements=[]) | cst.Dict(elements=[]) | cst.Tuple(elements=[]):
```
drdavella, line 71: It could be in a different ticket but it might also be nice to handle the following (admittedly less likely) cases:
* `list()`
* `dict()`
* `tuple()`
* `set()`

In each case we'd need to validate the name to make sure it refers to the builtin.

src/core_codemods/docs/pixee_python_fix-empty-sequence-comparison.md
```diff
@@ -0,0 +1,11 @@
+Empty sequences in a boolean comparison expression are considered falsy so you can use implicit boolean comparisons instead of
+comparing against empty sequences directly
```
drdavella: Nitpick but there shouldn't be a line break here.

src/core_codemods/docs/pixee_python_fix-empty-sequence-comparison.md
```diff
@@ -0,0 +1,11 @@
+Empty sequences in a boolean comparison expression are considered falsy so you can use implicit boolean comparisons instead of
```
drdavella:
```suggestion
Empty sequences in Python always evaluate to `False`. This means that comparison expressions that use empty sequences can sometimes be simplified. In these cases no explicit comparison is required: instead we can rely on the [truth value](https://docs.python.org/3/library/stdtypes.html#truth-value-testing) of the object under comparison. This is sometimes referred to as "implicit" comparison. Using implicit boolean comparison expressions is considered best practice and can lead to better code.
```

src/core_codemods/fix_empty_sequence_comparison.py
```diff
@@ -0,0 +1,67 @@
...
+class FixEmptySequenceComparison(
+    BaseCodemod, NameResolutionMixin, AncestorPatternsMixin
+):
+    NAME = "fix-empty-sequence-comparison"
+    SUMMARY = "Replace Comparisons to Empty Sequence with Implicit Boolean Comparison"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
+    DESCRIPTION = (
+        "Replace comparisons to empty sequence with implicit boolean comparison."
+    )
+    REFERENCES: list = []
```
drdavella: https://docs.python.org/3/library/stdtypes.html#truth-value-testing

src/core_codemods/fix_empty_sequence_comparison.py
```diff
@@ -0,0 +1,67 @@
...
+from codemodder.codemods.api import BaseCodemod, ReviewGuidance
+from codemodder.codemods.utils_mixin import NameResolutionMixin, AncestorPatternsMixin
+
+
+class FixEmptySequenceComparison(
+    BaseCodemod, NameResolutionMixin, AncestorPatternsMixin
+):
+    NAME = "fix-empty-sequence-comparison"
+    SUMMARY = "Replace Comparisons to Empty Sequence with Implicit Boolean Comparison"
+    REVIEW_GUIDANCE = ReviewGuidance.MERGE_WITHOUT_REVIEW
```
drdavella: This needs to be `MERGE_AFTER_REVIEW` since this has the potential to break existing code.

drdavella (changes requested):
I like this codemod and the implementation is nice. But we need to be a bit careful for the reasons described in the [pylint rule](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/use-implicit-booleaness-not-comparison.html):
>  Following this check blindly in weakly typed code base can create hard to debug issues. If the value can be something else that is falsey but not a sequence (for example ``None``, an empty string, or ``0``) the code will not be equivalent.

It would be safer if we could somehow verify that the object was the same type but that's probably not going to be possible in general.

clavedeluna:
> I like this codemod and the implementation is nice. But we need to be a bit careful for the reasons described in the [pylint rule](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/use-implicit-booleaness-not-comparison.html):
> 
> > Following this check blindly in weakly typed code base can create hard to debug issues. If the value can be something else that is falsey but not a sequence (for example `None`, an empty string, or `0`) the code will not be equivalent.
> 
> It would be safer if we could somehow verify that the object was the same type but that's probably not going to be possible in general.

You're right. Ideally we would be able to limit this check to cases of value we're checking being sets themselves, but that's difficult. Unless we've already somehow solved the problem of "what is the underlying value of x?" Maybe we have it and I can rely on it @andrecsilva ?
```
x = [1,2]
if x != []:
    pass    
```
and not cases:
```
x = 'hi' # or None, or int
if x != []:
    pass    
```

andrecsilva:
> > I like this codemod and the implementation is nice. But we need to be a bit careful for the reasons described in the [pylint rule](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/use-implicit-booleaness-not-comparison.html):
> > > Following this check blindly in weakly typed code base can create hard to debug issues. If the value can be something else that is falsey but not a sequence (for example `None`, an empty string, or `0`) the code will not be equivalent.
> > 
> > 
> > It would be safer if we could somehow verify that the object was the same type but that's probably not going to be possible in general.
> 
> You're right. Ideally we would be able to limit this check to cases of value we're checking being sets themselves, but that's difficult. Unless we've already somehow solved the problem of "what is the underlying value of x?" Maybe we have it and I can rely on it @andrecsilva ?
> 
> ```
> x = [1,2]
> if x != []:
>     pass    
> ```
> 
> and not cases:
> 
> ```
> x = 'hi' # or None, or int
> if x != []:
>     pass    
> ```

We have `resolve_expression` in `utils_mixin.py` for tracking a variable through assignments. Still, it cannot guarantee its value unless it has only a single assignment to it. However, if we already know the value of the variable, what would be the point of the comparison in the first place?

drdavella:
This is certainly tricky. It's a good example of a codemod where an LLM might be able to provide more context/judgment. I'll take the blame for not evaluating the pylint rule carefully enough before giving the go-ahead for this work. I'm okay with accepting this codemod as-is (modulo comments above) under the condition that it is disabled by default.

clavedeluna:
> This is certainly tricky. It's a good example of a codemod where an LLM might be able to provide more context/judgment. I'll take the blame for not evaluating the pylint rule carefully enough before giving the go-ahead for this work. I'm okay with accepting this codemod as-is (modulo comments above) under the condition that it is disabled by default.

I'm not attached to this codemod and I always go for whatever is best for the product, so @drdavella let me know if you'd prefer:
1. Close PR, document come back later in life
2. make merge after review as stated above + not register the codemod  (disabled by default, we don't have another way to enable/disable)

drdavella:
@clavedeluna there is actually one other option: we define a variable called `DEFAULT_EXCLUDED_CODEMODS` in `registry.py`. You could add this one there and I think we'd be good.

src/core_codemods/fix_empty_sequence_comparison.py
clavedeluna: I will document for later

drdavella (approved):

PR #213, opened 2024-01-15
Implement new codemod API

drdavella:
## Overview
*Implement new codemod API*

## Description

* This refactoring was driven by the following goals:
  * Simplified codemod class hierarchy
  * Cleaner metadata validation and more consistent metadata representation
  * More generality for other detectors (e.g. Sonar input) and transformers (e.g. LLMs or other AST/CST implementations)
  * Decoupling of base codemod implementation from LibCST constructs
  * Enable transformers to be shared between different codemods
  * Enable multiple transformations to be applied in sequence by the same codemod
* Codemods now consist of the following attributes:
  * Metadata (e.g. name, review guidance, references, etc.)
  * Detector (optional) which either performs its own analysis, possibly by running a tool like Semgrep, or parses files generated by an external tool
  * A pipeline of one or more transformations
* Some notable changes:
  * All codemods (except for one) use the same (simplified) base class. The one exception is simply for the sake of example
  * No codemods use external Semgrep YAML files

## How to review this PR
* Apologies in advance for the huge PR. This change touched a huge number of files but most of the changes are not that complicated.
* Start by looking at the new codemod implementations. This should give you a sense of the general philosophy behind the changes here.
* Next look at `base_codemod.py` and `libcst_transformer.py`
  * There's not a lot of new stuff here. Most of it has just been rearranged and moved to a new location
* Next look at the new `api.py` file. This contains the simplified API implementation which is intended for most of our current use cases
* Next look at tests. The updates here are pretty uniform and not too complicated


src/codemodder/codemods/base_codemod.py
```diff
@@ -17,109 +22,163 @@ class ReviewGuidance(Enum):
...
-    def add_change(self, node, description: str, start: bool = True):
-        position = self.node_position(node)
-        self.add_change_from_position(position, description, start)
+    Codemods that apply the same transformation but use different detectors
+    should be implemented as distinct codemod classes.
+    """
 
-    def add_change_from_position(
-        self, position: CodeRange, description: str, start: bool = True
+    metadata: Metadata
```
andrecsilva: Having metadata as a class attribute only to expose each member with properties seems a bit pointless.

src/codemodder/codemods/libcst_transformer.py
```diff
@@ -0,0 +1,305 @@
...
+            logger.exception("error parsing file %s", file_path)
+            return None
+
+        tree = source_tree
+        with file_context.timer.measure("transform"):
+            for transformer in self.transformers:
+                tree = transformer.transform(tree, results, file_context)
+
+        if tree.deep_equals(source_tree):
+            return None
```
andrecsilva: Expensive. I would change this to check if any changes were generated for the file.

src/codemodder/codemods/libcst_transformer.py
```diff
@@ -0,0 +1,305 @@
...
+                    source_tree = cst.parse_module(f.read())
+        except Exception:
+            file_context.add_failure(file_path)
+            logger.exception("error parsing file %s", file_path)
+            return None
+
+        tree = source_tree
+        with file_context.timer.measure("transform"):
+            for transformer in self.transformers:
+                tree = transformer.transform(tree, results, file_context)
```
andrecsilva, line 279: Results will be stale after the first transform. You either generate the results (with a detector) before applying each transformer or ignore the results. I suggest making a pipeline class for each case.

If you'd still want to make a pipeline that reuses the results for each transform, then my suggestion is to create a detector that always returns the same results.

src/codemodder/codemods/api.py
```diff
@@ -0,0 +1,52 @@
...
+)
+from codemodder.codemods.libcst_transformer import (
+    LibcstResultTransformer,
+    LibcstTransformerPipeline,
+)
+from codemodder.file_context import FileContext  # pylint: disable=unused-import
+from codemodder.codemods.semgrep import SemgrepRuleDetector
+
+
+class SimpleCodemod(LibcstResultTransformer, metaclass=ABCMeta):
```
andrecsilva, line 20: You went through the trouble of separating codemods into three components: metadata, transformer, and detector, only to muddle them all together again in this class. It's a bit telling it doesn't use the new `BaseCodemod` class at all (and must, instead, rely on an attribute for that).

Not sure why it the codemod class should be concerned with details like `detector_pattern` or `on_result_found`. Those should be the responsibility of both the detector and transformer, respectively. 

Maybe this exists as a convenient way to define everything in the same class, however not only that defeats the purpose of separation to begin with, but it's also not really any different from the old `BaseCodemod` class.

I feel like any codemod class should be simple composition of those three components. So it should look like:
```python
SandboxUrlSemgrepDetector = SemgrepDetector(rule = ...)
SandboxUrlTransformer = LibcstResultTransformer(callback = on_result_found)
SandboxUrlCodemod = BaseCodemod(SandBoxUrlSemgrepDetector, SandboxUrlTransformer, metadata = ...)
LimitReadlineCodemod = SemgrepCodemod(LimitReadlineTransformer, metadata = ..., semgrep_rule = ...)
```
or
```python
class LimitReadline(SemgrepCodemod):
         metadata = ...
         transformer = LimitReadlineTransformer
         semgrep_rule = ...         
```

My suggestion is to first think how you want the codemods to be expressed (e.g. the two examples above) and then work backwards from there.

src/core_codemods/refactor/refactor_new_api.py
```diff
@@ -0,0 +1,297 @@
+import libcst as cst
+
+from core_codemods.api import SimpleCodemod, Metadata, ReviewGuidance
+from codemodder.codemods.utils_mixin import NameResolutionMixin
+
+
+class RefactorNewApi(SimpleCodemod, NameResolutionMixin):
```
clavedeluna, line 7: this clearly took some effort so I can feel the reason to commit it, but do you think we will put it to use later on? maybe add a docstring explaining this was used during this transition.

tests/codemods/base_codemod_test.py
```diff
@@ -19,43 +16,84 @@ class BaseCodemodTest:
...
-        assert self.file_context and self.file_context.dependencies == set([dependency])
+        assert expected_diff == changes.diff
+        assert os.path.relpath(file_path, root) == changes.path
+
+        with open(file_path, "r", encoding="utf-8") as tmp_file:
+            output_code = tmp_file.read()
+
+        assert output_code == dedent(expected)
+
+    def run_and_assert_filepath(  # pylint: disable=too-many-arguments
```
clavedeluna, line 86: you could just add this disable to the top of the file instead of for every method

tests/codemods/test_combine_startswith_endswith.py
```diff
@@ -37,4 +36,3 @@ def test_combine(self, tmpdir, func):
     )
     def test_no_change(self, tmpdir, code):
         self.run_and_assert(tmpdir, code, code)
```
clavedeluna, line 38: I'm not seeing num_changes=0 being passed so just want to make sure that removal of `assert len(self.file_context.codemod_changes) == 0` is indeed covered in new tests refactor

clavedeluna (approved):
very minor non-blocking questions. I'm super impressed how easy this looks on the reviewer side, awesome job Dan!

andrecsilva (changes requested):
There are good ideas that came with the new API, but I don't think the current implementation reflects them.

src/core_codemods/refactor/refactor_new_api.py
drdavella: I can add a docstring. It is one-time use but it's also a nice demonstration of the capabilities of codemods. I intend to blog about it later. It also genuinely saved me some time with refactoring ~35 of the codemods.

tests/codemods/base_codemod_test.py
drdavella: Good point, I actually thought about just disabling it globally since it's not my favorite pylint rule. Any thoughts?

tests/codemods/test_combine_startswith_endswith.py
drdavella: Yes it's covered. The case where `expected == result` explicitly checks for that condition.

src/codemodder/codemods/libcst_transformer.py
drdavella: @andrecsilva this is not new code. I have raised this particular concern before but you specifically mentioned the difficulty of determining whether a transformation has applied which is why we are still doing it this way.

src/codemodder/codemods/base_codemod.py
drdavella: It's not pointless: it provides a separation of concerns and enables metadata fields to be validated cleanly at `__init__`.

src/codemodder/codemods/api.py
drdavella: > Maybe this exists as a convenient way to define everything in the same class, however not only that defeats the purpose of separation to begin with, but it's also not really any different from the old BaseCodemod class.

This is exactly the point of this API and I believe that this is an unfair criticism. Making the simple cases simple to express is necessarily in tension with providing the full generality that we want from the `BaseCodemod` API. We did in fact work backwards from the API that we wanted, which was already reflected in most of these codemod implementations. However as we were all aware, the original implementation was both not powerful enough and also quite convoluted in terms of the underlying class hierarchy. So adding an API layer to make the simple cases simple seems both worthwhile and like a good design.

> I feel like any codemod class should be simple composition of those three components

I am in agreement with you. The `BaseCodemod` API makes this explicit, except that the composition is expressed in terms of a class definition instead of a function call.

If you would like to propose an alternative API for the simple codemod cases, I am open to considering it. In truth I'm not very happy with the underlying implementation of `SimpleCodemod` since it involves some metaprogramming magic. But I think the API should be judged by its use cases which are clean and simple. I would encourage you to rewrite those examples in terms of your proposed API and see whether that leads to cleaner code and we can evaluate it from there.

src/codemodder/codemods/libcst_transformer.py
drdavella: This is fair. That does make sense if the detector belongs to the transformer pipeline.

src/codemodder/codemods/api.py
drdavella: As an addendum: I am not trying to hide the fact that this simplified API was explicitly designed to minimize the changes required from the previous simplified API and the new API. I don't see that as an unworthy goal.

src/codemodder/codemods/libcst_transformer.py
andrecsilva: That's fair, but I remember raising the same point before.

If the tree changed without any `Change` being generated, then either the transformation is not reporting things correctly, or it's changing it in a non-functional way (e.g. formatting). I think we can safely ditch the resulting tree in any case.

src/codemodder/codemods/base_codemod.py
andrecsilva: They can be validated at `__init__` without the properties or at the creation of the `Metadata` object.

My suggestion is to either mark the metadata attribute as "private" or ditch the properties.

drdavella: I can make it private.

src/codemodder/codemods/libcst_transformer.py
drdavella: That's okay. I'll try this change and as long as nothing breaks we should be good.

src/codemodder/codemods/libcst_transformer.py
drdavella: You raise a really good point here, and you're absolutely right about the potential for the results to become stale in general. However, I still think that architecturally the codemod itself is the right place for `detector`. This makes easier to reason about codemods that might share the same transformations but use different detectors.

To be clear: the cases I had in mind for multiple transformers fell into one of two categories:
* Cases where the transformers themselves perform detection (which is perfectly valid)
* Cases involving LLM-based transformations

src/codemodder/codemods/api.py
andrecsilva: > This is exactly the point of this API and I believe that this is an unfair criticism. Making the simple cases simple to express is necessarily in tension with providing the full generality that we want from the BaseCodemod API.

If that's the intention, then it's perfectly fine. It does have the benefits of being easy and convenient to throw everything into a class with a few methods/attributes. 

Having a clear (and enforced) separation between detection/transformation/metadata does have other benefits like modularity and reusability. You can't really reuse any transformation defined within a `SimpleCodemod`.

>I am in agreement with you. The BaseCodemod API makes this explicit, except that the composition is expressed in terms of a class definition instead of a function call.

The point is that the composition is muddled with the `SimpleCodemod` class. It is a `Codemod` and `Transformation` at the same time. For a more clear separation, it should "have" a transformation and not be one. If the intent is clear separation, I don't think it should have any logic pertaining the transformation itself.

> If you would like to propose an alternative API for the simple codemod cases, I am open to considering it. In truth I'm not very happy with the underlying implementation of SimpleCodemod since it involves some metaprogramming magic. But I think the API should be judged by its use cases which are clean and simple. I would encourage you to rewrite those examples in terms of your proposed API and see whether that leads to cleaner code and we can evaluate it from there.

I'm happy with how it turned out. In fact, you may recall that I've suggested something close to this before. The point of the aforementioned comment is to point that the introduction of the `SimpleCodemod` introduces a new way to express codemods and that makes you lose some of the benefits of the new design. Again, if you were already aware of that, then that's fine.

src/codemodder/codemods/libcst_transformer.py
drdavella: I think part of the problem (and maybe the point you were trying to make when we discussed this awhile ago) is that some transformations might be implemented such that in some cases the result is equivalent to the original code. In which case the tree check effectively needs to move to those particular codemods.

src/codemodder/codemods/api.py
drdavella: >  You can't really reuse any transformation defined within a SimpleCodemod.

This is okay with me, but that's good point. I think that cases where the transformation is shared between multiple codemods will need to be refactored.

> The point is that the composition is muddled with the SimpleCodemod class. It is a Codemod and Transformation at the same time. For a more clear separation, it should "have" a transformation and not be one. If the intent is clear separation, I don't think it should have any logic pertaining the transformation itself.

I understand the criticism here and I don't disagree but it is also fundamentally at odds with providing a "concise" simplified API which is another explicit design goal here.

> The point of the aforementioned comment is to point that the introduction of the SimpleCodemod introduces a new way to express codemods and that makes you lose some of the benefits of the new design.

Again, you are not wrong, but I don't think it's so problematic. `SimpleCodemod` is just sugar to make the easy cases easy. It reduces a lot of boilerplate (and refactoring) that would otherwise be required. It just so happens that most of our existing codemods fit into the "simple" category but I expect that there will be many other codemods that will require the lower-level API.

src/codemodder/codemods/libcst_transformer.py
andrecsilva: You can indeed have different trees that produce the same code. It usually happens with differences in whitespace/formatting nodes. But again, I think we could also ditch the resulting tree in this case. After all, it makes no difference.

src/codemodder/codemods/libcst_transformer.py
andrecsilva: I think the burden of composing transformations should be with the codemod itself.

My suggestion is to change the `Codemod` class to accept a list of transformations, and move this logic into the `apply` function of the codemod. This way you could use the detector to guarantee results before each transformation. It also means the `Pipeline` classes of transformations wouldn't be needed.

drdavella: The problem with that suggestion is that there is a phase mismatch between detectors and transformers. Detectors apply to a list of files whereas a transformer applies to a single file. This is necessarily the case due the fact that a tool like Semgrep or Sonar returns results that apply to a collection of files whereas LibCST transformations are applied to a single file at a time. Similarly we would expect LLM-based transformations to apply to one file at a time as well.

andrecsilva (approved):
I think we should fix the pipeline issue before merging this. It will most likely need some structural changes in the code. 
However, since no codemod use it for now, I'm willing to accept this. Just make sure to add a comment raising this issue.

PR #219, opened 2024-01-19
Codemod: remove-assertion-in-pytest-raises

andrecsilva:
## Overview
Adds a codemod that removes assertions from the bottom of `pytest.raises` context manager.

### DIFF
--- a/src/core_codemods/remove_assertion_in_pytest_raises.py
+++ b/src/core_codemods/remove_assertion_in_pytest_raises.py
@@ -91,10 +91,8 @@ class RemoveAssertionInPytestRaisesTransformer(
     ) -> Union[
         cst.BaseStatement, cst.FlattenSentinel[cst.BaseStatement], cst.RemovalSentinel
     ]:
-        if not self.filter_by_path_includes_or_excludes(
-            self.node_position(original_node)
-        ):
-            return updated_node
+        # TODO: add filter by include or exclude that works for nodes
+        # that that have different start/end numbers.
 
         # Are all items pytest.raises?
         if not self._all_pytest_raises(original_node):
--- a/src/core_codemods/remove_assertion_in_pytest_raises.py
+++ b/src/core_codemods/remove_assertion_in_pytest_raises.py
@@ -43,7 +43,6 @@ class RemoveAssertionInPytestRaisesTransformer(
                 case cst.Assert():
                     assert_position = assert_position - 1
                     assert_stmts.append(self._build_simple_statement_line(stmt))
-                    self.report_change(stmt)
                 case _:
                     break
         if assert_position > 0:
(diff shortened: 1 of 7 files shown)


src/core_codemods/remove_assertion_in_pytest_raises.py
```diff
@@ -0,0 +1,157 @@
...
+        assert_stmts.reverse()
+        return assert_stmts, assert_position, new_statement_before_asserts
+
+    def leave_With(
+        self, original_node: cst.With, updated_node: cst.With
+    ) -> Union[
+        cst.BaseStatement, cst.FlattenSentinel[cst.BaseStatement], cst.RemovalSentinel
+    ]:
+        if not self.filter_by_path_includes_or_excludes(original_node):
+            return updated_node
```
clavedeluna: see `test_exclude_line` to add here

src/core_codemods/remove_assertion_in_pytest_raises.py
```diff
@@ -0,0 +1,157 @@
...
+                )
+            return cst.FlattenSentinel([new_with, *assert_stmts])
+
+        return updated_node
+
+
+RemoveAssertionInPytestRaises = CoreCodemod(
+    metadata=Metadata(
+        name="remove-assertion-in-pytest-raises",
+        summary="Moves assertions out of with statement body",
```
clavedeluna: out of pytest.raises scope I would say

tests/codemods/test_remove_assertion_in_pytest_raises.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def test_with_item_not_raises(self, tmpdir):
+        input_code = """\
+        import pytest
+        def foo():
+            with pytest.raises(ZeroDivisionError), open('') as file:
+                1/0
+                assert True
+        """
+        self.run_and_assert(tmpdir, input_code, input_code)
```
clavedeluna, line 191: need test case for `from pytest import raises` and for `from pytest import raises as renamed_raises`

clavedeluna: also would need a no_change test case: 
```
with pytest.raises:
     assert True
     1/0
```

clavedeluna (changes requested):

src/core_codemods/remove_assertion_in_pytest_raises.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def leave_With(
+        self, original_node: cst.With, updated_node: cst.With
+    ) -> Union[
+        cst.BaseStatement, cst.FlattenSentinel[cst.BaseStatement], cst.RemovalSentinel
+    ]:
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return updated_node
```
clavedeluna: you added a `test_exclude_line` but still it says not covered.... weird. 

Well actually, given this is a leave_with node, it's likely the same issue as the other codemods with this todo `# TODO: add filter by include or exclude that works for nodes`. So I think you can go ahead and remove the unit test and just add the todo line

clavedeluna (approved):

PR #223, opened 2024-01-25
Sonar Integration

andrecsilva:
## Overview
Added support for sonar and general SAST codemods

## Description
* Sonar results processing with `SonarResultSet`;
* Sonar detection in codemods with `SonarDetector` (results files are passed via `CodemodExecutionContext`);
* `SASTCodemod` and `SonarCodemod` classes for writing SAST and Sonar codemods;
* New test classes for SASTCodemods;
* Refactored `numpy-nan-equality`;
* First new sonar codemod: `numpy-nan-equality-S6725` (sonar version of `numpy-nan-equality`);
* Changed behavior of CLI. If the `--sarif` or `--sonar-issues-json` are included, only SAST codemods will be loaded by default.


src/codemodder/codemods/base_codemod.py
```diff
@@ -107,11 +108,10 @@ def docs_module(self) -> Traversable:
 
     @cached_property
     def description(self) -> str:
-        if not self._metadata.has_description:
-            return ""
-
-        doc_path = self.docs_module / f"{self.origin}_python_{self.name}.md"
-        return doc_path.read_text()
+        if self._metadata.description == None:
```
drdavella:
```suggestion
        if self._metadata.description is None:
```

Fixing this probably makes it unnecessary to ignore the type check below.

src/codemodder/context.py
```diff
@@ -61,6 +63,10 @@ def __init__(
         self.path_include = path_include
         self.path_exclude = path_exclude
         self.max_workers = max_workers
+        if tool_result_files_map:
```
drdavella: Would probably be a bit cleaner as a one-liner:
```python
```

integration_tests/base_test.py
```diff
@@ -81,10 +82,16 @@ def _assert_run_fields(self, run, output_path):
         assert run["tool"] == "codemodder-python"
         assert run["version"] == __version__
         assert run["elapsed"] != ""
-        assert (
-            run["commandLine"]
-            == f"codemodder {SAMPLES_DIR} --output {output_path} --codemod-include={self.codemod_instance.name} --path-include={self.code_path}"
-        )
+        if self.sonar_issues_json:
+            assert (
```
drdavella: Just a small thing but feels like it might be a bit cleaner with something like:
```python
expected_cmd = base_cmd + ("--sonar-issues-json ..." if self.sonar_issues_json else "")
```

But I just have an irrational aversion to `if` blocks 😆

src/codemodder/codemodder.py
```diff
@@ -136,6 +137,13 @@ def record_dependency_update(dependency_results: dict[Dependency, PackageStore |
             logger.debug("The following dependencies could not be added: %s", str_list)
 
 
+def process_sonar_findings(sonar_json_files: list[str]) -> SonarResultSet:
+    combined_result_set = SonarResultSet()
+    for file in sonar_json_files or []:
+        combined_result_set |= SonarResultSet.from_json(file)
```
drdavella: This is neat.

src/codemodder/codemods/sonar.py
```diff
@@ -0,0 +1,34 @@
...
+            combined_result_set |= SonarResultSet.from_json(file)
+        return combined_result_set
+
+    def apply(
+        self,
+        codemod_id: str,
+        context: CodemodExecutionContext,
+        files_to_analyze: list[Path],
+    ) -> ResultSet:
+        if not self._lazy_cache:
```
drdavella, line 30: Any reason not to use `funcutils.lru_cache`?

src/codemodder/registry.py
```diff
@@ -59,14 +59,24 @@ def match_codemods(
         self,
         codemod_include: Optional[list] = None,
         codemod_exclude: Optional[list] = None,
+        sast_only=False,
     ) -> list[BaseCodemod]:
         codemod_include = codemod_include or []
         codemod_exclude = codemod_exclude or DEFAULT_EXCLUDED_CODEMODS
 
+        if sast_only:
```
drdavella: Feels like you could move the conditional to be within the list comprehension to avoid the duplication.

src/codemodder/scripts/generate_docs.py
```diff
@@ -224,6 +224,14 @@ class DocMetadata:
     ),
 }
 
+METADATA = CORE_METADATA | {
+    "numpy-nan-equality-S6725": DocMetadata(
```
drdavella, line 228: Is this the actual Sonar rule ID?

tests/test_codemod_docs.py
```diff
@@ -12,10 +12,11 @@ def pytest_generate_tests(metafunc):
 
 
 def test_load_codemod_docs_info(codemod: BaseCodemod):
-    if codemod.name in ["order-imports", "refactor-new-api"]:
+    print(codemod.name)
```
drdavella: Leftover print statement.

src/core_codemods/order_imports.py
```diff
@@ -14,7 +14,7 @@ class OrderImports(SimpleCodemod):
         name="order-imports",
         summary="Order Imports",
         review_guidance=ReviewGuidance.MERGE_WITHOUT_REVIEW,
-        has_description=False,
+        description="",
```
drdavella, line 17: Not the biggest deal but I'm not in love with `""` and `None` as the sentinels for these cases. I think `has_description` is clearer.

src/core_codemods/sonar_numpy_nan_equality.py
drdavella, line 1: I'd like to suggest that this should either be in a subdirectory of `core_codemods` or even in an entirely different package.

src/core_codemods/sonar_numpy_nan_equality.py
```diff
@@ -0,0 +1,42 @@
...
+    NumpyNanEquality,
+    NumpyNanEqualityTransformer,
+)
+
+
+class SonarNumpyNanEqualityTransformer(NumpyNanEqualityTransformer):
+    def leave_Comparison(
+        self, original_node: cst.Comparison, updated_node: cst.Comparison
+    ) -> cst.BaseExpression:
+        if self.filter_by_result(self.node_position(original_node)):
```
drdavella: Maybe I'm missing something but shouldn't this filter just be moved to the parent class?

src/core_codemods/sonar_numpy_nan_equality.py
```diff
@@ -0,0 +1,42 @@
...
+            return super().leave_Comparison(original_node, updated_node)
+        return updated_node
+
+
+rules = ["python:S6725"]
+
+SonarNumpyNanEquality = SonarCodemod(
+    metadata=Metadata(
+        name="numpy-nan-equality-S6725",
+        summary="Sonar: " + NumpyNanEquality.summary,
```
drdavella: Not necessarily for this PR but it seems like there should be some kind of shortcut to avoid all of these references to the core codemod class or else we're going to end up with a lot of repeated code.

drdavella (approved):
This looks really nice overall, I'm happy with how this turned out. My comments just consist of some nitpicks and relatively minor questions.

src/codemodder/scripts/generate_docs.py
andrecsilva: `S6725` or more precisely `python:S6725` is the id for the rule. I'm following the naming convention adopted by `codemodder-java`.

src/core_codemods/order_imports.py
andrecsilva: This change was necessary as a means to define the description of a sonar codemod in terms of other codemod. As it was, a codemod would only source it's description from a file or not. Thus, I would have to copy-paste the description to it's own file.

tests/test_codemod_docs.py
andrecsilva: I should make a codemod for that and make it a pre-commit hook.

src/codemodder/codemods/sonar.py
andrecsilva: ~~I remember deliberately doing it like this to test something. But I forgot what it was. I'll just use `lru_cache`.~~
Actually, now I remember. The cache from `lru_cache` depends on the arguments, meaning if  you change the arguments the cache won't apply. This means you would have to process the sonar json for each codemod. Doing it manually like this avoids that.

src/core_codemods/sonar_numpy_nan_equality.py
andrecsilva: The parent class has an empty detector. This means that `self.results` is empty. Both `filter_by_result` and `node_is_selected` would fail in that instance.

src/core_codemods/sonar_numpy_nan_equality.py
andrecsilva: Yeah, I was figuring out how viable it was to derive the metadata from the original codemod as I went. I'll make it easier for the next codemod. Probably with a new constructor for `SonarCodemod`.

src/core_codemods/sonar_numpy_nan_equality.py
drdavella: Makes sense, thanks. Although it also would be useful if we could just define `filter_by_result` to be vacuously true in those cases.

src/core_codemods/order_imports.py
drdavella: I understand, I guess what I'm saying is that keeping `has_description` as a way to differentiate those cases would retain some clarity. But it's not a big deal either way.

integration_tests/test_program.py
```diff
@@ -21,3 +24,21 @@ def test_codemods_include_exclude_conflict(self):
...
+                "--sonar-issues-json",
+                f"{tmp_file_path}",
+                "--dry-run",
+            ],
+            check=False,
+            capture_output=True,
+            text=True,
+        )
+        assert completed_process.returncode == 0
+        assert RemoveAssertionInPytestRaises.id not in completed_process.stdout
```
clavedeluna, line 44: can you explain the goal of this assertion?

andrecsilva: The goal is to make sure no non-sast codemods are loaded by default while the `--sarif` or `--sonar-...` flags are passed. Couldn't think of a better way other than looking for the codemod ids printed to stdout. `RemoveAssertionInPytestRaises` just happens to the one I've used to check it.

clavedeluna (approved):
Would be good to add test coverage to at least some of the locations codecov is suggesting, some seem like a quick unit test will do

PR #224, opened 2024-01-26
Fix path inclusion behavior

drdavella:
## Overview
*Fix path inclusion behavior*

## Description

* While doing some testing I discovered an issue with the behavior of `--path-include`
* Included paths should always be interpreted as relative to the given directory path
* I am going to cherry pick this into a bugfix release since I don't want to release it with the updated API changes

### DIFF
--- a/tests/test_code_directory.py
+++ b/tests/test_code_directory.py
@@ -42,7 +42,7 @@ class TestMatchFiles:
 
     def test_match_excluded(self, dir_structure):
         expected = ["empty_for_testing.py", "insecure_random.py"]
-        files = match_files(dir_structure, ["**/tests/**", "*request.py"])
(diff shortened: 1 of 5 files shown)


andrecsilva (approved):

tests/test_code_directory.py
```diff
@@ -42,7 +42,7 @@ def test_all_py_files_match(self, dir_structure):
 
     def test_match_excluded(self, dir_structure):
         expected = ["empty_for_testing.py", "insecure_random.py"]
-        files = match_files(dir_structure, ["**/tests/**", "*request.py"])
```
clavedeluna, line 45: would it make sense to keep one test with **/tests... ?

drdavella: Possibly but with the new behavior this isn't a valid match anymore.

PR #226, opened 2024-01-29
New way to define a `SonarCodemod` and `filter_by_result` behavior change

andrecsilva:
## Description

* `filter_by_result` and `node_is_selected` will now be vacuously true if the detector is `None`
* Added new way to define `SonarCodemod` from a given `CoreCodemod`
* Refactored `sonar_numpy_nan_equality` and `numpy_nan_equality`

### DIFF
--- a/src/codemodder/codemods/sonar.py
+++ b/src/codemodder/codemods/sonar.py
@@ -1,15 +1,44 @@
 from pathlib import Path
+from codemodder.codemods.base_codemod import Metadata, Reference
 from codemodder.codemods.base_detector import BaseDetector
+from codemodder.codemods.base_transformer import BaseTransformerPipeline
 from codemodder.context import CodemodExecutionContext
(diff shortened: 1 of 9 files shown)


src/codemodder/codemods/sonar.py
```diff
@@ -1,16 +1,45 @@
...
 
+    @classmethod
+    def from_core_codemod(
+        cls,
+        name: str,
+        other: CoreCodemod,
+        rules: list[str],
+        transformer: BaseTransformerPipeline | None = None,
+        new_references: list[Reference] | None = None,
+    ):  # pylint: disable=too-many-arguments
```
drdavella, line 24: Not for this PR but I think we should just reconfigure or disable this pylint rule globally.

drdavella (approved):
Perfect, thank you.

clavedeluna (approved):

PR #227, opened 2024-01-29
Codemod to convert eager logging to lazy logging

clavedeluna:
## Overview
*Lazy logging is preferred as str operations only run if logging will occur*

## Description
This codemod handles cases for logging with `%` and with `+` operators. I tried to test as many cases as I came up with. However, see unit tests carefully as there are some edge cases I decided to leave for a future iteration. Specifically the edge case of logging str concat with `+` but a `%` str literal is present.

### DIFF
--- a/src/core_codemods/docs/pixee_python_lazy-logging.md
+++ b/src/core_codemods/docs/pixee_python_lazy-logging.md
@@ -1,5 +1,5 @@
 This codemod converts "eager" logging into "lazy" logging, which is preferred for performance efficiency and resource optimization.
-Lazy logging defers the actual construction and formatting of log messages until it's confirmed that the message will be logged based on the current log level, thereby avoiding unnecessary computation and memory usage for messages that will be ignored. 
+Lazy logging defers the actual construction and formatting of log messages until it's confirmed that the message will be logged based on the current log level, thereby avoiding unnecessary computation for messages that will not be logged. 
 
 Our changes look something like this:
 
--- a/src/core_codemods/docs/pixee_python_lazy-logging.md
+++ b/src/core_codemods/docs/pixee_python_lazy-logging.md
@@ -1,0 +1,13 @@
+This codemod converts "eager" logging into "lazy" logging, which is preferred for performance efficiency and resource optimization.
+Lazy logging defers the actual construction and formatting of log messages until it's confirmed that the message will be logged based on the current log level, thereby avoiding unnecessary computation and memory usage for messages that will be ignored. 
+
+Our changes look something like this:
+
+```diff
+import logging
+e = "Some error"
+- logging.error("Error occurred: %s" % e)
+- logging.error("Error occurred: " + e)
++ logging.error("Error occurred: %s", e)
++ logging.error("Error occurred: %s", e)
+```
--- a/src/core_codemods/lazy_logging.py
+++ b/src/core_codemods/lazy_logging.py
@@ -49,27 +49,92 @@ class LazyLogging(SimpleCodemod, NameResolutionMixin):
                 $VAR = logging.getLogger(...)
                 ...
             {_log_funcs}
+          - patterns:
+            - pattern: logging.$FUNC(..., $ANYTHING + ..., ...)
+            {_pattern_inside}
+            {_log_funcs}
+          - patterns:
+            - pattern: logging.getLogger(...).$FUNC(..., $ANYTHING + ..., ...)
+            {_pattern_inside}
+            {_log_funcs}
+          - patterns:
+            - pattern: $VAR.$FUNC(..., $ANYTHING + ..., ...)
+            - pattern-inside: |
+                import logging
+                ...
+                $VAR = logging.getLogger(...)
+                ...
+            {_log_funcs}
         """
 
     def on_result_found(self, original_node, updated_node):
         del original_node
+
         match updated_node.func:
             case cst.Name(value="log") | cst.Attribute(attr=cst.Name(value="log")):
                 # logging.log(INFO, ...), log(INFO, ...)
                 first_arg = [updated_node.args[0]]
                 remaining_args = list(updated_node.args[2:])
-                format_string = updated_node.args[1].value.left
-                format_args = updated_node.args[1].value.right
+                binop = updated_node.args[1].value
             case _:
                 first_arg = []
                 remaining_args = list(updated_node.args[1:])
-                format_string = updated_node.args[0].value.left
-                format_args = updated_node.args[0].value.right
+                binop = updated_node.args[0].value
 
-        new_args = [cst.Arg(value=format_string)]
-        if isinstance(format_args, cst.Tuple):
-            for element in format_args.elements:
-                new_args.append(cst.Arg(value=element.value))
-        else:
-            new_args.append(cst.Arg(value=format_args))
+        if set(self.all_operators(binop)) == {cst.Add, cst.Modulo}:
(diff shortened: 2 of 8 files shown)


src/core_codemods/lazy_logging.py
```diff
@@ -0,0 +1,160 @@
...
+            {_pattern_inside}
+            {_log_funcs}
+          - patterns:
+            - pattern: $VAR.$FUNC(..., ... % ..., ...)
+            - pattern-inside: |
+                import logging
+                ...
+                $VAR = logging.getLogger(...)
+                ...
+            {_log_funcs}
```
andrecsilva, line 53: I think the above pattern will cover this case because of constant propagation.

src/core_codemods/lazy_logging.py
```diff
@@ -0,0 +1,160 @@
...
+            {_pattern_inside}
+            {_log_funcs}
+          - patterns:
+            - pattern: $VAR.$FUNC(..., $ANYTHING + ..., ...)
+            - pattern-inside: |
+                import logging
+                ...
+                $VAR = logging.getLogger(...)
+                ...
+            {_log_funcs}
```
andrecsilva, line 69: Same as above.

src/core_codemods/lazy_logging.py
```diff
@@ -0,0 +1,160 @@
...
+            return None
+        left_type = infer_expression_type(self.resolve_expression(binop.left))
+        right_type = infer_expression_type(self.resolve_expression(binop.right))
+        if left_type != right_type or (type_both_sides := left_type) not in {
+            BaseType.STRING,
+            BaseType.BYTES,
+        }:
+            # Cannot concat different types.
+            # Skip logging ints, etc. Eg: `logging.info(2+2)`
+            return None
```
andrecsilva, line 112: You can't concat different types in Python with `+`, so if we assume the code compiles, we could avoid the first check.

src/core_codemods/lazy_logging.py
```diff
@@ -0,0 +1,160 @@
...
+            format_strings = []
+        if format_args is None:
+            format_args = []
+
+        match node:
+            case cst.BinaryOperation(operator=cst.Add()):
+                self.process_concat(node.left, format_strings, format_args)
+                self.process_concat(node.right, format_strings, format_args)
+            case cst.SimpleString():
+                format_strings.append(node.value.strip("\"'"))
```
andrecsilva: I think this won't work if the string itself has any prefix (e.g. `u'some string'`). You can use `raw_value` to remove the quotes and prefix from the value of a `SimpleString`.

However I think you should try to preserve the string here with it's prefixes. Consider, for example:
```python
logger.info("message " + r"I\'m a raw value")
```
If you strip the `r` prefix the value of the string will be changed.

andrecsilva (changes requested):
It needs some consideration on cases with string concatenation having different prefixes. Otherwise looks good.

src/core_codemods/lazy_logging.py
clavedeluna: By above pattern do you mean ln42 `- pattern: logging.getLogger(...).$FUNC(..., ... % ..., ...)`? I'll test it to make sure

src/core_codemods/lazy_logging.py
clavedeluna: This if handles one of the not yet supported cases, skipped test `test_log_format_specifiers`. I also don't think we can assume codemodder will run on code that compiles so I think the check is worth it. Plus it makes it clear that we want both sides to be of the same type.

src/core_codemods/lazy_logging.py
andrecsilva: Yes, I think it covers the highlighted one. But it's not a big issue if it's duplicated.

src/core_codemods/lazy_logging.py
andrecsilva: We do have to assume that it's valid python code (otherwise `libcst` won't even parse it). I guess this is more of a runtime exception case.
It's fine to go without this assumption, but `infer_expression` does assume that for the `BinaryOperation` cases. We can adjust it though.

src/core_codemods/lazy_logging.py
```diff
@@ -0,0 +1,160 @@
...
+                ...
+"""
+    _log_funcs = """\
+- metavariable-pattern:
+                  metavariable: $FUNC
+                  patterns:
+                    - pattern-either:
+                      - pattern: debug
+                      - pattern: info
+                      - pattern: warning
```
drdavella, line 29: Even though it's deprecated `warn` is still a valid log level too.

tests/codemods/test_lazy_logging.py
```diff
@@ -0,0 +1,683 @@
...
+
+    def test_name(self):
+        assert self.codemod.name == "lazy-logging"
+
+    @each_func
+    @pytest.mark.parametrize(
+        "input_code,expected_output",
+        [
+            (
+                """\
```
drdavella: Just a very small thing but because of the way that these test cases are automatically dedented, I don't think the line continuation is necessary here. It ends up looking a little cleaner without.

tests/codemods/test_lazy_logging.py
```diff
@@ -0,0 +1,683 @@
+import pytest
+from tests.codemods.base_codemod_test import BaseSemgrepCodemodTest
+from core_codemods.lazy_logging import LazyLogging
+
+logging_funcs = {"debug", "info", "warning", "error", "critical"}
```
drdavella: Let's include `warn` here as well. I know we have a separate codemod to fix that but it means that the application of these codemods won't be sensitive to ordering.

src/core_codemods/docs/pixee_python_lazy-logging.md
```diff
@@ -0,0 +1,13 @@
+This codemod converts "eager" logging into "lazy" logging, which is preferred for performance efficiency and resource optimization.
+Lazy logging defers the actual construction and formatting of log messages until it's confirmed that the message will be logged based on the current log level, thereby avoiding unnecessary computation and memory usage for messages that will be ignored.
```
drdavella:
```suggestion
Lazy logging defers the actual construction and formatting of log messages until it's confirmed that the message will be logged based on the current log level, thereby avoiding unnecessary computation for messages that will not be logged. 
```

You're not wrong about memory usage, just tightening it up a bit.

drdavella (approved):
A few minor comments but looks really nice overall. I'm approving this on the basis of the test cases which look like what I would expect behaviorally. It seems like @andrecsilva has covered comments about implementation details so as soon as those are addressed this should be good to go.

src/core_codemods/lazy_logging.py
clavedeluna: this does not appear to be the case according to my testing.

src/core_codemods/lazy_logging.py
clavedeluna: 1. libcst does parse it:
```
  File "/Users/dani/codemodder-python/test.py", line 4, in <module>
    logging.info("hi" + 2)
                 ~~~~~^~~
TypeError: can only concatenate str (not "int") to str
```
run this: no error
```
import libcst as cst
source_code = """
import logging
logging.info("hi" + 2)
"""
module = cst.parse_module(source_code)
```

2. I reiterate, this check is needed for cases like `"Scientific Value: " + format(scientific_val, '.2e'))`

andrecsilva (approved):

PR #235, opened 2024-02-01
Adds sonar versions for several codemods

andrecsilva:
## Overview
Adds sonar versions for several codemods

## Description
Added tool specific result location match
The following codemods were refactored and added as a sonar codemod:
* literal-or-new-object-identity
* django-receiver-on-top
* exception-without-raise
* fix-assert-tuple


drdavella (approved):
Really great work!

PR #237, opened 2024-02-05
Codemodder should add a "dependency already present" to codetf result

clavedeluna:
## Overview
*When a dependency that codemodder would add already exists in the project, report a specific codetf description message.*

## Description

* This PR is a bit "shallow" in that in the long-term we really would need to handle a codemod being able to add multiple dependencies, some that may already exist, some that need to be added. However, we decided to not handle for that at this time.
* We also decided to punt on the case for what to do if two codemods run that would add the same dependency. At this time, the current behavior is tested in the new integration test:  the first codemod that runs gets the "we added a new dependency to the project" message, while the second codemod gets a "dependency already present" message. We can come back to this later.

### DIFF
--- a/src/codemodder/context.py
+++ b/src/codemodder/context.py
@@ -140,7 +140,6 @@ class CodemodExecutionContext:  # pylint: disable=too-many-instance-attributes
                 )
                 for dep in dependencies:
                     record[dep] = package_store
-                break
             else:
                 self._dependency_update_by_codemod[codemod_id] = (
                     Update.PRESENT,
--- a/src/codemodder/context.py
+++ b/src/codemodder/context.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-
+from enum import Enum
 import logging
 from pathlib import Path
 import itertools
@@ -11,6 +11,7 @@ from codemodder.dependency import (
     Dependency,
     build_dependency_notification,
     build_failed_dependency_notification,
+    build_dependency_is_present_notification,
 )
 from codemodder.file_context import FileContext
 from codemodder.logging import logger, log_list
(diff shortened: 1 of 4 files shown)


src/codemodder/context.py
```diff
@@ -115,32 +123,48 @@ def process_dependencies(
...
             if (changeset := dm.write(list(dependencies), self.dry_run)) is not None:
                 self.add_results(codemod_id, [changeset])
-                self._dependency_update_by_codemod[codemod_id] = package_store
+                self._dependency_update_by_codemod[codemod_id] = (
+                    Update.ADD,
+                    package_store,
+                )
                 for dep in dependencies:
                     record[dep] = package_store
-                break
```
clavedeluna, line 131: we don't have a good example of needing this break (multiple dependency stores) so I removed it for now

clavedeluna: NVM actually, we do have an integration test that hits this and we do need the `break`. @drdavella the logic here is clear to me what it should be if `store_list` is only one thing, but if it's multiple, it's less clear. Right now if `if (changeset := dm.write(list(dependencies), self.dry_run)) is not None:` ends up being None, we would report
```
self._dependency_update_by_codemod[codemod_id] = (
                    Update.PRESENT,
                    package_store,
                )
```
but all remaining store_lists. Thoughts?

drdavella: I think I see what you're saying. Unfortunately I think the `Optional[ChangeSet]` being returned by `dm.write` does not provide enough granularity for the kinds of conditions we can encounter here. I think there are really three separate conditions:
* This particular package store does not exist, can't be written: continue
* This package store exists and we added a new dependency
* This package store exists and already has the given dependency

This probably means making some changes to the dependency manager and package store return values to properly support it.

PR #241, opened 2024-02-06
remove dedent and \ from unit tests

clavedeluna:
## Overview
*We keep forgetting unit test base_test already dedents so this is just a cleanup*


drdavella:
@clavedeluna you should have written a codemod to do this! 😆

drdavella (approved):

andrecsilva (approved):

PR #243, opened 2024-02-06
Support `httpx` library in requests-verify codemod

clavedeluna:
## Overview
*`httpx` is a requests library with an API almost identical to `requests` so we're adding it in this codemod*

### DIFF
--- a/src/core_codemods/requests_verify.py
+++ b/src/core_codemods/requests_verify.py
@@ -14,13 +14,14 @@ class RequestsVerify(SimpleCodemod):
         review_guidance=ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW,
         references=[
             Reference(url="https://requests.readthedocs.io/en/latest/api/"),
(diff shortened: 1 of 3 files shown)


src/core_codemods/requests_verify.py
```diff
@@ -23,12 +23,18 @@ class RequestsVerify(SimpleCodemod):
...
+            rules:
+              - pattern-either:
+                - patterns:
+                    - pattern: requests.$F(..., verify=False, ...)
+                    - pattern-inside: |
+                        import requests
+                        ...
+                - patterns:
+                    - pattern: httpx.$F(..., verify=False, ...)
+                    - pattern-inside: |
```
clavedeluna, line 36: I wanted to use metavariable-pattern to just do `$LIBRARY.$F....` but using `import $LIBRARY` in pattern-inside did not work?? weird

andrecsilva (approved):

drdavella (approved):

PR #245, opened 2024-02-07
Added support for multiple fixes per block in fix-file-resouce-leak

andrecsilva:
## Overview
`fix-file-resource-leak` will now fix multiple leaks in a single block at once

### DIFF
--- a/src/core_codemods/file_resource_leak.py
+++ b/src/core_codemods/file_resource_leak.py
@@ -1,7 +1,12 @@
 from typing import Optional, Sequence
+from codemodder.codemods.libcst_transformer import (
+    LibcstResultTransformer,
+    LibcstTransformerPipeline,
+)
+from codemodder.result import Result
 from codemodder.utils.utils import extract_targets_of_assignment
 import libcst as cst
-from libcst import ensure_type, matchers
+from libcst import SimpleStatementLine, ensure_type, matchers
 from libcst.codemod import (
     CodemodContext,
     ContextAwareVisitor,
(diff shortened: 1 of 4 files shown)


src/core_codemods/file_resource_leak.py
```diff
@@ -157,59 +172,118 @@ class ResourceLeakFixer(
...
+        # assigned to something that is not a Name?
+        if other_targets:
+            return False
+        # yield, returned, argument of a call, referenced outside of block
+        name_escapes_partial = partial(
+            self._name_escapes_scope, block=block, index=index
+        )
+        # is closed?
+        name_condition = map(
+            # pylint: disable-next=cell-var-from-loop
```
clavedeluna: hmmm either this pylint warning is worth listening to or it's a pylint bug? Either way, maybe a list-comp is easier to read here?

clavedeluna (approved):
Personal preference nit pick would be to use less one-letter variable names as it's harder to grep for them .

src/core_codemods/file_resource_leak.py
andrecsilva: I wanted to avoid building the list since the result would just be fed into `all`. I don't exactly remember the issue with the warning was. It may be invalid by now. I'll try to remove and see if it still complains.

src/core_codemods/file_resource_leak.py
```diff
@@ -21,70 +26,75 @@
...
         ParentNodeProvider,
     )
 
     def __init__(
         self,
         context: CodemodContext,
+        results: list[Result],
         file_context: FileContext,
-        *codemod_args,
+        *codemod_args,  # pylint: disable=unused-argument
```
drdavella: Small thing but I think the idiomatic way to suppress this is with
```python
del *codemod_args
```

Although possibly this argument is no longer needed?

drdavella (approved):

PR #248, opened 2024-02-09
Codemod to fix `asyncio.Task`

clavedeluna:
## Overview
*Replace manual init of asyncio.Task with `create_task`*

## Description

* special case of `loop=not-a-loop` in which I left without changes, I can update accordingly if we do want to change it.


tests/codemods/test_fix_task_instantiation.py
```diff
@@ -0,0 +1,195 @@
...
+        [
+            (
+                """
+                import asyncio
+                asyncio.Task(coro(1, 2), name='task', eager_start=True)
+                """,
+                """
+                import asyncio
+                asyncio.create_task(coro(1, 2), name='task', eager_start=True)
+                """,
```
andrecsilva: This changes the intended behavior of the code. The `eager_start` argument has seemingly no effect on `create_task`.

I think the `ayncio.eager_task_factory` is the intended way to create eager tasks.
https://docs.python.org/3/library/asyncio-task.html#eager-task-factory

andrecsilva (changes requested):
Just a small correction on converting tasks with `eager_start` parameter.

tests/codemods/test_fix_task_instantiation.py
clavedeluna: ah interesting. I only used eager_task as a kwarg test example to demonstrate we correctly handle kwargs. I'll see if I use another kwarg or if we need to special case for eager_task

src/core_codemods/fix_task_instantiation.py
```diff
@@ -0,0 +1,96 @@
...
+from core_codemods.api import Metadata, ReviewGuidance, SimpleCodemod, Reference
+from codemodder.codemods.utils_mixin import NameAndAncestorResolutionMixin
+from codemodder.codemods.utils import BaseType, infer_expression_type
+
+
+class FixTaskInstantiation(SimpleCodemod, NameAndAncestorResolutionMixin):
+    metadata = Metadata(
+        name="fix-task-instantiation",
+        summary="Use high-level `asyncio.create_task` API",
+        review_guidance=ReviewGuidance.MERGE_WITHOUT_REVIEW,
```
drdavella: I think I'd like to hedge on this a bit and call it `MERGE_AFTER_CURSORY_REVIEW`. Feel free to push back but I'm exposing my own unfamiliarity with the `async` module.

drdavella:
I've asked @bdoyal to take a look at the test cases just as another sanity check to make sure we're producing correct code.

src/core_codemods/fix_task_instantiation.py
```diff
@@ -0,0 +1,96 @@
+import libcst as cst
+from typing import Optional
+from core_codemods.api import Metadata, ReviewGuidance, SimpleCodemod, Reference
+from codemodder.codemods.utils_mixin import NameAndAncestorResolutionMixin
+from codemodder.codemods.utils import BaseType, infer_expression_type
+
+
+class FixTaskInstantiation(SimpleCodemod, NameAndAncestorResolutionMixin):
+    metadata = Metadata(
+        name="fix-task-instantiation",
```
drdavella:
```suggestion
        name="fix-async-task-instantiation",
```
Probably should be reflected in the name of the class/file as well 🙏

src/core_codemods/fix_task_instantiation.py
```diff
@@ -0,0 +1,96 @@
...
+        summary="Use high-level `asyncio.create_task` API",
```
drdavella:
```suggestion
        summary="Use `asyncio.create_task` API to instantiate `asyncio.Task`",
```

drdavella (changes requested):
This is a very nice codemod overall. I agree with @andrecsilva that it looks like we need to handle eager tasks slightly differently. Hopefully that's not too difficult. Also I requested some small changes to the metadata.

src/core_codemods/docs/pixee_python_fix-task-instantiation.md
```diff
@@ -0,0 +1,9 @@
+The `asyncio` [documentation](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task) explicitly discourages manual instantiation of a `Task` instance and instead recommends calling `create_task`.
+
```
drdavella: I wish we had a little more to say here. Maybe adding something like:
> This keeps your code in line with recommended best practices and helps to promote correctness and maintainability?

src/core_codemods/fix_task_instantiation.py
```diff
@@ -0,0 +1,96 @@
...
+                if loop_type == BaseType.NONE:
+                    return self.node_create_task(original_node, updated_node)
+                if loop_type in (
+                    BaseType.NUMBER,
+                    BaseType.LIST,
+                    BaseType.STRING,
+                    BaseType.BYTES,
+                    BaseType.BOOL,
+                ):
+                    # incorrectly assigned loop kwarg to something that is not a loop.
```
drdavella: I think this is the safest thing to do here.

tests/codemods/test_fix_task_instantiation.py
```diff
@@ -0,0 +1,195 @@
...
+                import asyncio
+                my_loop = asyncio.get_event_loop()
+                asyncio.Task(coro(1, 2), loop=my_loop)
+                """,
+                """
+                import asyncio
+                my_loop = asyncio.get_event_loop()
+                my_loop.create_task(coro(1, 2))
+                """,
+            ),
```
bdoyal: What is the significance of referencing the loop for this codemod in this test? To ensure an event loop is running? Asking since this test case uses `asyncio.get_event_loop()` and there is a slight difference with create_task in that it calls `asyncio.get_running_loop()` rather than `asyncio.get_event_loop()`.  AFAIK `asynio.get_event_loop()` will create a loop if none is running and `asyncio.get_running_loop()` will raise an exception if no loop is running.  I'm not entirely sure it significant here, but thought I should mention it since it is a little sneaky.  For reference:

https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop

bdoyal (commented):
@drdavella asked me to take a look at the aysnc inputs/outputs in one of the test files, so I'm just here to leave comments on that.

tests/codemods/test_fix_task_instantiation.py
clavedeluna: @andrecsilva could you double check my understanding of the docs, since it's a tiny bit confusing. Seems like eager_start is new so not a ton of use cases yet out there.

So if I understand correctly, if you want to create a `Task()` with eager_start=True, the code would be
```
loop = asyncio.get_running_loop()
task = asyncio.Task(my_coroutine(), loop=loop, eager_start=True)
```
(must pass in a loop otherwise it won't work)

Our replacement could either be
```
  loop = asyncio.get_running_loop()
  loop.set_task_factory(asyncio.eager_task_factory)
  task = loop.create_task(my_coroutine())
```
which is kinda what the docs suggest? but this also works:

```
  loop = asyncio.get_running_loop()
  task = asyncio.eager_task_factory(loop, my_coroutine())
```
though I"m not 100% sure they're totally equivalent.

I'm a tiny bit leaning to not support this right now if we aren't sure, but maybe I'm missing something today.

tests/codemods/test_fix_task_instantiation.py
clavedeluna: Very sneaky indeed! 
Using `get_event_loop` in these tests is actually only for demonstration purposes. For these tests, we don't care if the loop came from `get_event_loop` or just some loop that was passed in from some other random function. All we care about is that if a loop is passed in as a kwarg to `asyncio.Task`, then we use that very loop and call `create_task` on it, and not call `asyncio.create_task`.

So your explanation actually reinforces this unit test, since we wouldn't not want to call `get_running_loop` and raise an error.

LMK if that didn't make sense or if you feel a new unit test is needed.

tests/codemods/test_fix_task_instantiation.py
andrecsilva: That is also my understanding from reading the documentation.

src/core_codemods/fix_task_instantiation.py
clavedeluna: @drdavella I don't think we should use your exact summary wording bc the codemod may change to not only `asyncio.create_task`, but also to `some_loop.create_task` and now to `asyncio.eager_task_factory` which is why I just said high-level API.

clavedeluna:
latest commits add support for `eager_task`. this is a new feature in py3.12 and the docs are a bit confusing and there aren't a lot of examples online, so hopefully I got it right.

src/core_codemods/fix_task_instantiation.py
drdavella: How about `Use high-level `asyncio` APIs to instantiate tasks`?

Also I think our convention is to use title case for the summary.

drdavella: Nvm you basically did that already. I didn't even notice 🤦

src/core_codemods/fix_async_task_instantiation.py
```diff
@@ -0,0 +1,169 @@
...
+        name="fix-async-task-instantiation",
+        summary="Use High-Level `asyncio` API Functions to Create Tasks",
+        review_guidance=ReviewGuidance.MERGE_AFTER_CURSORY_REVIEW,
+        references=[
+            Reference(
+                url="https://docs.python.org/3/library/asyncio-task.html#asyncio.Task"
+            ),
+        ],
+    )
+    change_description = "Replace instantiation of `asyncio.Task` higher-level functions to create tasks."
```
drdavella:
```suggestion
    change_description = "Replace instantiation of `asyncio.Task` with higher-level functions to create tasks."
```

drdavella (approved):
Just asking for one more very minor change but otherwise looks 👍

andrecsilva (approved):

PR #256, opened 2024-02-13
change requests timeout codemod semgrep pattern

### LINKED ISSUE #244: Add requests timeout codemod is adding timeout kwarg to other methods
image

### Example of Issue
The following code is part of a pull request made by pixie to a setup script in one of my projects after a correctly triggering an [AddRequestsTimeouts](https://github.com/pixee/codemodder-python/blob/main/src/core_codemods/add_requests_timeouts.py) Codemod.

```python
versions = requests.get(versions_url, timeout=60).json(timeout=60)
```
This clearly would result in a `TypeError` since `"timeout"` is not a valid kwarg for `JSONDecoder.__init__`.

### Potential Cause
I assume that this issue is caused by the OpenAI function call outputs that are used to create code edits not being properly linted on the backend before the resulting code is included in a pull request.

### Potential Solution
A possible solution could be adding a valid function and argument parsing class to [file_parsers](https://github.com/pixee/codemodder-python/tree/main/src/codemodder/project_analysis/file_parsers) that gathers the valid function names, argument names and types by checking the args in function definitions, typeing info and calls to kwargs.get/pop in each function for each file in the repo. 

Then each function call in the AI written code could be linted for valid arguments+types and an error could be raised that triggers a subsequent OpenAI call requesting a revision and providing the valid args+types to the model to insure code is valid before including it a pull request.

As a backup in cases where multiple revisions produce invalid code, the valid args for functions in a given code snippet could be specified as an enum in the OpenAI function/tool definitions used to edit code.

Alternatively an Assistants API instance with the code_interpreter tool enabled could be prompted to test code snippets to prevent this issue but this would of course increase costs.

Good luck with this project absolutely awesome idea!

clavedeluna:
## Overview
*change semgrep rule for codemod to not match chained functions*

Closes #244 

More context:  the only other codemod that suffers from this is the `bad-lock-with-statement`. The issue there is the way we're using metavariable-pattern and focus-variable makes this fix not as easy.

### DIFF
--- a/src/core_codemods/add_requests_timeouts.py
+++ b/src/core_codemods/add_requests_timeouts.py
@@ -40,20 +40,20 @@ AddRequestsTimeouts = CoreCodemod(
             - pattern-inside: |
                 import requests
                 ...
-            - pattern: $CALL(...)
-            - pattern-not: $CALL(..., timeout=$TIMEOUT, ...)
+            - pattern: requests.$CALL(...)
+            - pattern-not: requests.$CALL(..., timeout=$TIMEOUT, ...)
             - metavariable-pattern:
                 metavariable: $CALL
                 patterns:
                   - pattern-either:
-                    - pattern: requests.get
-                    - pattern: requests.post
-                    - pattern: requests.put
-                    - pattern: requests.delete
-                    - pattern: requests.head
-                    - pattern: requests.options
-                    - pattern: requests.patch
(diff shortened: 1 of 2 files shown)


andrecsilva (approved):

drdavella (approved):
Nice fix! I'll backport a bugfix release once this is merged.

PR #257, opened 2024-02-13
Added check for setting Content-Type for flask-json-content-type

andrecsilva:
## Overview
Fixes a small issue with flask-json-response-type where the Content-Type was already set but the fix was still applied.


drdavella:
I think this looks great. I am closing and then reopening this PR just to test an integration.

drdavella (approved):
This is great, thanks.

PR #269, opened 2024-02-15
Filter sonar issues by status

### LINKED ISSUE #267: Ignore closed/fixed issues in Sonar security JSON results
It looks like Sonar security JSON files may possibly include closed/fixed issues in the results. We should filter the list of issues to make sure that we do not include closed/fixed issues when applying codemods.

The relevant issue fields appear to be `resolution` and `status`:
```json
"issues": [
    {
      // ...more fields here...
      "resolution": "FIXED",
      "status": "CLOSED",
      // ...more fields here...
    }
]
```

Let's make sure to ignore issues where `resolution == "FIXED"` or `status == "CLOSED"`. We may need to account for other values in the future.

andrecsilva:
## Overview
Sonar issues whose status are RESOLVED are now filtered from the results. Closes #267 .


drdavella (commented):
Is `"RESOLVED"` the only status we want to ignore? What about `"CLOSED"`? Probably there are others? Is it possible that we want to invert this and only consider issues that are `"OPEN"`?

clavedeluna:
> Is `"RESOLVED"` the only status we want to ignore? What about `"CLOSED"`? Probably there are others? Is it possible that we want to invert this and only consider issues that are `"OPEN"`?

and `resolution == "FIXED"`

andrecsilva:
> Is `"RESOLVED"` the only status we want to ignore? What about `"CLOSED"`? Probably there are others? Is it possible that we want to invert this and only consider issues that are `"OPEN"`?

@clavedeluna @drdavella 
You can only have `RESOLVED` or `OPEN`. If an issue is marked as `RESOLVED` it is accompanied by a `resolution` field. I don't remember the exact values for it but `WONTFIX` and `FIXED` are some of them.

This was tested in the latest SonarQube version, locally. It may be the case that other values are possible from previous version.

drdavella:
@andrecsilva the snippet in #267 was taken from actual API output and it appears to contain `"CLOSED"` as a possible value. It feels safer if we invert the condition and just check for `"OPEN"`.

It seems likely that the API might have other values than those produced by SonarQube locally since users can make decisions about how to triage certain issues.

clavedeluna (approved):

drdavella (approved):

drdavella:
Closes #267

PR #275, opened 2024-02-19
Fix CodemodCollection API

drdavella:
## Overview
*Remove unused `CodemodCollection` attributes*

## Description

* These should have been removed as part of the API refactoring in #213
* Also fixed an issue that was causing test errors when using `pytest-xdist` locally

### DIFF
--- a/tests/codemods/test_lazy_logging.py
+++ b/tests/codemods/test_lazy_logging.py
@@ -3,7 +3,7 @@ from tests.codemods.base_codemod_test import BaseSemgrepCodemodTest
 from core_codemods.lazy_logging import LazyLogging
 
 logging_funcs = {"debug", "info", "warning", "warn", "error", "critical"}
-each_func = pytest.mark.parametrize("func", logging_funcs)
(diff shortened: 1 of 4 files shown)


clavedeluna (approved):

andrecsilva (approved):

tests/codemods/test_lazy_logging.py
```diff
@@ -3,7 +3,7 @@
 from core_codemods.lazy_logging import LazyLogging
 
 logging_funcs = {"debug", "info", "warning", "warn", "error", "critical"}
-each_func = pytest.mark.parametrize("func", logging_funcs)
+each_func = pytest.mark.parametrize("func", sorted(logging_funcs))
```
drdavella, line 6: @clavedeluna this would be a good codemod, although we currently ignore test directories by default. I'd like to update that logic, though, so that the ignored directories are codemod-dependent.

clavedeluna: sorry which part exactly do you suggest as a codemod? sorting part?

PR #277, opened 2024-02-19
Replace pylint with ruff

### LINKED ISSUE #276: Use `ruff` instead of `pylint`
We should be using `ruff` since it's much faster and supports the same rules: https://github.com/astral-sh/ruff

clavedeluna:
## Overview
*`ruff` is a linter similar to pylint, but it's faster so we're going to favor using it*

## Description

* To do this switch, I removed everything pylint related and did some very small `noqa` changes to align with ruff since it warned us about some things that we actually had disabled in pylintrc
* FYII: inline ignore is done with `# noqa: id` while top-of-file is done with `# ruff: noqa: id`
* Now `ruff` will run locally in your pre-commit (yay!) as well as in CI under the pre-commit step, instead of as a separate linting step.
* I didn't do any deep dive to compare pylint and ruff so it's possible we may lose some interesting pylint checks that ruff doesn't have.

## Additional Details
* I can go a step further and actually replace`black` with `ruff format`.  Check out the difference by running `ruff format  src tests integration_tests --exclude tests/samples/`. It's minor differences. 

Closes #276


Makefile
```diff
@@ -11,9 +11,6 @@ integration-test:
 pygoat-test:
 	${PYTEST} -v ci_tests/test_pygoat_findings.py
 
-lint:
-	pylint -v codemodder core_codemods tests integration_tests
-
```
andrecsilva: Any reason for not replacing this with the `ruff` equivalent?

clavedeluna: Well... I figured since it runs in your pre-commit there isn't much of a point? Do you strongly want it?

drdavella: I use this command a lot, it would be good to keep around if possible.

andrecsilva (approved):

PR #302, opened 2024-02-25
Django models `__str__` codemod

### LINKED ISSUE #287: Codemod to add a `__str__` method to a Django model that didn't define it
Django users write lots of models and may forget that not defining a `__str__` will make the model really hard to read in admin tools.
It would be good to change this model
```
from django.db import models

class User(models.Model):
    name = models.CharField(max_length=100)

   .... 
```
To include:
```
    def __str__(self):
        return f'{self.name}'
```

The challenges I see to this is to decide what should `__str__` return? Which fields should we include? Is something like this acceptable?
```
    def __str__(self):
        return f'{field for field in all_defined_fields}'
```

clavedeluna:
## Overview
*Codemod that will add a cookie-cutter `__str__` method to any django model class that did not implement it*

## Description

* The implement

## Additional Details
* Found this django app plugin during my research https://pypi.org/project/django-dunder/#history
* There is a corresponding sonar rule for this but I wasn't able to get sonarcloud to fire so this will be investigated later on in a separate issue. Please ignore early sonarcloud failures as I was testing that part out.

Closes #287

### DIFF
--- a/src/core_codemods/django_model_without_dunder_str.py
+++ b/src/core_codemods/django_model_without_dunder_str.py
@@ -38,7 +38,7 @@ class DjangoModelWithoutDunderStrTransformer(
         self.report_change(original_node)
 
         dunder_str = cst.FunctionDef(
-            leading_lines=[cst.EmptyLine()],
+            leading_lines=[cst.EmptyLine(indent=False)],
             name=cst.Name("__str__"),
             params=cst.Parameters(params=[cst.Param(name=cst.Name("self"))]),
             body=cst.IndentedBlock(body=[cst.SimpleStatementLine(body=[cst.Pass()])]),
--- a/src/core_codemods/django_model_without_dunder_str.py
+++ b/src/core_codemods/django_model_without_dunder_str.py
@@ -36,8 +36,9 @@ class DjangoModelWithoutDunderStrTransformer(
             return updated_node
 
         self.report_change(original_node)
+
         dunder_str = cst.FunctionDef(
-            # leading_lines=[cst.EmptyLine()],
+            leading_lines=[cst.EmptyLine()],
             name=cst.Name("__str__"),
             params=cst.Parameters(params=[cst.Param(name=cst.Name("self"))]),
             body=cst.IndentedBlock(body=[cst.SimpleStatementLine(body=[cst.Pass()])]),
--- a/src/core_codemods/django_model_without_dunder_str.py
+++ b/src/core_codemods/django_model_without_dunder_str.py
@@ -43,9 +43,14 @@ class DjangoModelWithoutDunderStrTransformer(
         return updated_node.with_changes(body=new_body)
 
     def implements_dunder_str(self, original_node: cst.ClassDef) -> bool:
-        for node in original_node.body.body:
-            match node:
-                case cst.FunctionDef(name=cst.Name(value="__str__")):
+        """Check if a ClassDef or its bases implement `__str__`"""
+        if self.class_has_method(original_node, "__str__"):
+            return True
+
+        for base in original_node.bases:
+            if maybe_assignment := self.find_single_assignment(base.value):
+                classdef = maybe_assignment.node
+                if self.class_has_method(classdef, "__str__"):
                     return True
         return False
 
--- a/src/core_codemods/django_model_without_dunder_str.py
+++ b/src/core_codemods/django_model_without_dunder_str.py
@@ -76,7 +76,7 @@ def dunder_str_method() -> cst.FunctionDef:
 DjangoModelWithoutDunderStr = CoreCodemod(
     metadata=Metadata(
         name="django-model-without-dunder-str",
-        summary="Ensure Django Model Classes Implement A `__str__` Method",
+        summary="Ensure Django Model Classes Implement a `__str__` Method",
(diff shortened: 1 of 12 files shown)


src/core_codemods/django_model_without_dunder_str.py
```diff
@@ -0,0 +1,89 @@
...
+                    return True
+        return False
+
+
+def dunder_str_method() -> cst.FunctionDef:
+    self_body = cst.IndentedBlock(
+        body=[
+            cst.parse_statement("model_name = self.__class__.__name__"),
+            cst.parse_statement(
+                'fields_str = ", ".join([f"{field.name}={getattr(self, field.name)}" for field in self._meta.fields])'
```
andrecsilva: Change this from a list to a generator (`join([...])` -> `join((...))`).

src/core_codemods/docs/pixee_python_django-model-without-dunder-str.md
```diff
@@ -0,0 +1,30 @@
...
+```diff
+from django.db import models
+
+class Question(models.Model):
+    question_text = models.CharField(max_length=200)
+    pub_date = models.DateTimeField("date published")
++ 
++    def __str__(self):
++        model_name = self.__class__.__name__
++        fields_str = ", ".join([f"{field.name}={getattr(self, field.name)}" for field in self._meta.fields])
```
andrecsilva: Change this example to reflect the change in the other comment.

andrecsilva (approved):
Minor change

src/core_codemods/django_model_without_dunder_str.py
clavedeluna: doh! 🤦

src/core_codemods/docs/pixee_python_django-model-without-dunder-str.md
```diff
@@ -0,0 +1,30 @@
...
++        return f"{model_name}({fields_str})"
+```
+
+Without this change, the `Question` objects look like this in the interactive Django shell:
```
drdavella:
```suggestion
Without this change, the  string representation of `Question` objects look like this in the interactive Django shell:
```

src/core_codemods/docs/pixee_python_django-model-without-dunder-str.md
```diff
@@ -0,0 +1,30 @@
+If you've ever actively developed or debugged a Django application, you may have noticed Django models and their instances can sometimes be hard to read or distinguish one instance from another. Loading models in the interactive Django console or viewing them in the admin interface can be puzzling. This is because Django is trying to display your model objects as a plain strings.
```
drdavella:
```suggestion
If you've ever actively developed or debugged a Django application, you may have noticed that the string representations of Django models and their instances can sometimes be hard to read or to distinguish from one another. Loading models in the interactive Django console or viewing them in the admin interface can be puzzling. This is because the default string representation of Django models is fairly generic.
```

src/core_codemods/docs/pixee_python_django-model-without-dunder-str.md
```diff
@@ -0,0 +1,30 @@
...
+
+We've written this codemod to make your model objects human-readable. It will automatically detect all of your model's fields and display them as a nice string.
```
drdavella:
```suggestion
This codemod is intended to make the string representation of your model objects more human-readable. It will automatically detect all of your model's fields and display them as a descriptive string.
```

src/core_codemods/docs/pixee_python_django-model-without-dunder-str.md
```diff
@@ -0,0 +1,30 @@
...
+
+For example, the `Question` model from Django's popular Poll App tutorial will look like this:
```
drdavella:
```suggestion
For example, the default string representation of the `Question` model from Django's popular Poll App tutorial looks like this:
```

src/core_codemods/django_model_without_dunder_str.py
```diff
@@ -0,0 +1,89 @@
...
+        leading_lines=[cst.EmptyLine(indent=False)],
+        name=cst.Name("__str__"),
+        params=cst.Parameters(params=[cst.Param(name=cst.Name("self"))]),
+        body=self_body,
+    )
+
+
+DjangoModelWithoutDunderStr = CoreCodemod(
+    metadata=Metadata(
+        name="django-model-without-dunder-str",
```
drdavella, line 78: Not a huge deal but I feel like we're sometimes inconsistent as to whether a codemod is phrased in terms of a problem or a fix. I think we should prefer names in terms of fixes `add-django-model-dunder-str` but maybe we can just take it as a suggestion going forward.

src/core_codemods/django_model_without_dunder_str.py
```diff
@@ -0,0 +1,89 @@
...
+        summary="Ensure Django Model Classes Implement A `__str__` Method",
```
drdavella:
```suggestion
        summary="Ensure Django Model Classes Implement a `__str__` Method",
```

drdavella (changes requested):
Looks really nice overall. Just some requests for updating metadata and then we should be good.

drdavella (approved):

PR #307, opened 2024-02-27
Publish test utils

### LINKED ISSUE #289: Make test utilities part of the published package API
Currently classes such as `BaseCodemodTest` and `BaseIntegrationTest` are defined within our `tests` directory which is not part of the published package.

We want to enable external developers to create custom codemod plugins, which means that we should make these test utilities part of the published API.

This means moving these class definitions (and any dependencies) out of the `tests` directory (which should never be part of the package distribution) and into some kind of utilities module. My initial thought is that they should go somewhere like `src/codemodder/codemods/test_utilities/` although maybe there are better options.

This will enable external packages to import things like `BaseCodemodTest` and use them within their own packages. We should do our best to remove any hard-coded assumptions in these base classes about directory paths, etc.

clavedeluna:
## Overview
*Publish base test clases so now a developer can do `from codemodder.codemods.test import BaseCodemodTest`*

## Description

* I don't know why code coverage went down but it did, by 0.5.

## Additional Details
* We may want to do some more cleanup or go ahead and remove the `GitPython` dependency, but since I know we want this published asap I made this PR at this point.

Closes #289


andrecsilva (approved):

drdavella (approved):
This is great, thanks!

PR #315, opened 2024-02-28
Requirements parser handles trailing comments

### LINKED ISSUE #310: Requirements.txt parser needs to handle lines with trailing comments
It is valid for for a `requirements.txt` file to have lines with trailing comments:
```
foo>=1.2.3   # very informative comment
```

Right now we properly filter for lines that _begin_ with comments but we also need to strip any trailing comments before parsing the requirement. This logic occurs here: https://github.com/pixee/codemodder-python/blob/2cdc91c609bc36cf3c295c0f505da7a2ea9af12f/src/codemodder/project_analysis/file_parsers/requirements_txt_file_parser.py#L31

The logic above properly strips whitespace and filters lines that begin with comments but does not properly account for trailing comments.

We need to:
1. Add a good test case
2. Update the parser to handle these kinds of lines

clavedeluna:
## Overview

Requirements parser handles trailing comments. Trailing comments will be fully ignored from parsing. 

Closes #310

### DIFF
--- a/src/codemodder/project_analysis/file_parsers/requirements_txt_file_parser.py
+++ b/src/codemodder/project_analysis/file_parsers/requirements_txt_file_parser.py
@@ -28,7 +28,9 @@ class RequirementsTxtParser(BaseParser):
             logger.debug("Unknown encoding for file: %s", file)
             return None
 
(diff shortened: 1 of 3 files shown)


drdavella (approved):
👍 great, thanks!

PR #318, opened 2024-02-29
Add filtering logic to use-defusedxml transformation

drdavella:
## Overview
*Add filtering logic to use-defusedxml transformation*


andrecsilva (commented):
I don't think this inclusion is actually doing anything. Line includes/excludes filtering should already be in place with the `DefusedXmlModifier` (i.e. in `leave_Call` for `ImportedCallModifier`) and this is not a result-based codemod. I'm gonna guess this change is for future-proofing, but I don't see any reason to include it right now.

drdavella:
@andrecsilva I should have included this in the description but I am using this codemod as the basis for a closed-source SAST remediation codemod and the filtering logic does not work without this change.

andrecsilva (approved):

PR #326, opened 2024-03-02
jwt.decode sonar codemod

### LINKED ISSUE #298: Sonar: fix "JWT should be signed and verified" with `jwt-decode-verify`
We should be able to use the `jwt-decode-verify` codemod to remediate this sonar rule.

https://rules.sonarsource.com/python/RSPEC-5659/

clavedeluna:
## Overview
*Add codemod support for the sonar rule for jwt.decode*

## Description

* The challenge with this codemod is matching sonar results column to `decode` call node column. Sonar returns column start/end for `....verify=False` or `...."verify_signature": True` while the semgrep node we care about is `jwt.decode` which contains these verify keywords. To get the codemod to run, I had to implement a  codemod specific `match_location` that I called `fuzzy`. 

## Additional Details
* some minor cleanup. /documentation

Closes #298

### DIFF
--- a/integration_tests/test_sonar_exception_without_raise.py
+++ b/integration_tests/test_sonar_exception_without_raise.py
@@ -1,7 +1,5 @@
-from core_codemods.exception_without_raise import (
-    ExceptionWithoutRaise,
-    ExceptionWithoutRaiseTransformer,
-)
+from core_codemods.exception_without_raise import ExceptionWithoutRaiseTransformer
+from core_codemods.sonar.sonar_exception_without_raise import SonarExceptionWithoutRaise
 from codemodder.codemods.test import (
     BaseIntegrationTest,
     original_and_expected_from_code_path,
@@ -9,7 +7,7 @@ from codemodder.codemods.test import (
 
 
 class TestSonarExceptionWithoutRaise(BaseIntegrationTest):
-    codemod = ExceptionWithoutRaise
+    codemod = SonarExceptionWithoutRaise
     code_path = "tests/samples/exception_without_raise.py"
     original_code, expected_new_code = original_and_expected_from_code_path(
         code_path,
--- a/src/codemodder/codemods/base_visitor.py
+++ b/src/codemodder/codemods/base_visitor.py
@@ -23,7 +23,7 @@ class UtilsMixin(MetadataDependent):
     def filter_by_result(self, node):
         pos_to_match = self.node_position(node)
         if self.results is None:
-            return False
+            return True  # should be false??
         return any(result.match_location(pos_to_match, node) for result in self.results)
 
     def filter_by_path_includes_or_excludes(self, pos_to_match):
--- a/src/codemodder/codemods/base_visitor.py
+++ b/src/codemodder/codemods/base_visitor.py
@@ -23,7 +23,7 @@ class UtilsMixin(MetadataDependent):
     def filter_by_result(self, node):
         pos_to_match = self.node_position(node)
         if self.results is None:
-            return True
+            return False
         return any(result.match_location(pos_to_match, node) for result in self.results)
 
     def filter_by_path_includes_or_excludes(self, pos_to_match):
--- a/src/codemodder/codemods/base_visitor.py
+++ b/src/codemodder/codemods/base_visitor.py
@@ -23,7 +23,8 @@ class UtilsMixin(MetadataDependent):
     def filter_by_result(self, node):
         pos_to_match = self.node_position(node)
         if self.results is None:
-            return True  # should be false??
+            # Some codemods must run without results existing.
+            return True
         return any(result.match_location(pos_to_match, node) for result in self.results)
 
     def filter_by_path_includes_or_excludes(self, pos_to_match):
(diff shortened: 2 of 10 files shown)


integration_tests/test_sonar_exception_without_raise.py
```diff
@@ -1,15 +1,13 @@
...
+from core_codemods.sonar.sonar_exception_without_raise import SonarExceptionWithoutRaise
 from codemodder.codemods.test import (
     BaseIntegrationTest,
     original_and_expected_from_code_path,
 )
 
 
 class TestSonarExceptionWithoutRaise(BaseIntegrationTest):
-    codemod = ExceptionWithoutRaise
+    codemod = SonarExceptionWithoutRaise
```
clavedeluna, line 10: I realized this int test was importing the wrong codemod

src/codemodder/codemods/base_visitor.py
```diff
@@ -23,6 +23,7 @@ def __init__(
     def filter_by_result(self, node):
         pos_to_match = self.node_position(node)
         if self.results is None:
+            # Some codemods must run without results existing.
             return True
```
clavedeluna, line 28: I documented this because it was strange to return True here (do we filter by result or not filter by result?). Then I realized the logic here is that this method is used for all codemods - those with and those without results. We still want to analyze codemods even if there are no results. Maybe later on we can have a `filter_by_result` that's more relevant. The one I implemented for jwt sonar returns False here, since there should always be results if we want to run the codemod.

src/core_codemods/sonar/sonar_jwt_decode_verify.py
```diff
@@ -0,0 +1,43 @@
...
+                return any(
+                    self.match_location(pos_to_match, result) for result in self.results
+                )
+        return False
+
+    def match_location(self, pos, result):
+        return any(
+            same_line(pos, location) and fuzzy_column_match(pos, location)
+            for location in result.locations
+        )
```
andrecsilva, line 31: Did you encounter any examples where `fuzzy_column_match` is needed? From my observations most nodes reported by sonar would match libcst ones with one exception: `Tuple`s. If you look at the `match_location` from `SonarResult` you can see the exception treated there. We should discuss how to treat location match at some point.

andrecsilva (approved):

src/core_codemods/sonar/sonar_jwt_decode_verify.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class JwtDecodeVerifySonarTransformer(JwtDecodeVerifyTransformer):
+    def filter_by_result(self, node) -> bool:
+        """
+        Special case result-matching for this rule because the sonar
+        results returned have a start/end column for the verify keyword
+        within the `decode` call, not for the entire call like semgrep returns.
+        """
+        if self.results is None:
```
drdavella: I believe this will actually never occur in the Sonar case and you can remove this.

src/codemodder/codemods/base_visitor.py
drdavella: Yes I encountered this too. We're using `None` as a sentinel for codemods that do not have detectors, and we do not want to short-circuit those cases. If you don't mind, could you update the comment to say something along those lines?

src/core_codemods/sonar/sonar_jwt_decode_verify.py
```diff
@@ -0,0 +1,43 @@
...
+            return False
+        match node:
```
drdavella, line 18: This solution makes sense to me and it works for now. In the longer term, I think we should filter the Sonar/SAST results before they ever reach the transformer. This means we would add some logic to each detector that requires it. This would mean implementing some kind of visitor that validates each of the detected locations before passing the results to the transformer. There's a bit of a performance impact in this case since it effectively means another pass on the file, but it only applies to files where there are already results.

I'm saying this not because I think we need this change right now but because I've encountered similar issues with remediating another SAST tool and I think that updating the filtering logic here for each case is going to become cumbersome.

drdavella (approved):
Looks great! Just a few smallish comments.

src/core_codemods/sonar/sonar_jwt_decode_verify.py
clavedeluna: You're going to have to give me an example because adding `fuzzy_column_match`  is the entire point of this PR. As I stated in the PR description, sonar returns index for the keyword=value, while we need jwt.decode(...keyword=value). Unit tests have exact sonar results, too. So inspect those as well.

Or perhaps Im' not understanding your comment?

src/core_codemods/sonar/sonar_jwt_decode_verify.py
clavedeluna: yep very good point. I was trying to keep this method as similar to its base one, but you're right it's nonsensical in this case.

src/core_codemods/sonar/sonar_jwt_decode_verify.py
clavedeluna: Very interesting! The detector for this particular codemod is semgrep so that's also a little wrinkle.

src/core_codemods/sonar/sonar_jwt_decode_verify.py
andrecsilva: In the simplest terms, does anything break if you remove `fuzzy_column_match` and replace it by exact match?
If so, does it break consistently? (e.g. always with the same type of node)

clavedeluna: Yes, there are no location matches without fuzzy matching.

PR #329, opened 2024-03-05
New codemod to replace `hasattr(obj, "__call__")` with `callable`

### LINKED ISSUE #290: Codemod to replace use of `hasattr(obj, "__call__") with `callable`
Calling `hasattr(some_obj, "__call__")` can yield different results than `callable(obj)` in some cases, such as when `__getattr__` is defined.

Same behavior:
```
>>> class One:
...     pass
...
>>> one = One()
>>> callable(one)
False
>>> hasattr(one, "__call__")
False
```

Differing behavior:
```
>>> class Two:
...      def __getattr__(self, name):
...          pass
...
>>> two = Two()
>>> callable(two)
False
>>> hasattr(two, "__call__")
True
```

clavedeluna:
## Overview
*Using `callable` is the recommended and reliable way to check if something is a callable*

Closes #290

### DIFF
--- a/src/core_codemods/fix_hasattr_call.py
+++ b/src/core_codemods/fix_hasattr_call.py
@@ -1,19 +1,16 @@
 import libcst as cst
-from core_codemods.api import (
-    Metadata,
-    Reference,
-    ReviewGuidance,
-    SimpleCodemod,
-)
+
+from core_codemods.api import Metadata, Reference, ReviewGuidance, SimpleCodemod
 
 
 class TransformFixHasattrCall(SimpleCodemod):
     metadata = Metadata(
         name="fix-hasattr-call",
-        summary="todo",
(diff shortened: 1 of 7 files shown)


src/core_codemods/fix_hasattr_call.py
```diff
@@ -0,0 +1,33 @@
...
+    Reference,
+    ReviewGuidance,
+    SimpleCodemod,
+)
+
+
+class TransformFixHasattrCall(SimpleCodemod):
+    metadata = Metadata(
+        name="fix-hasattr-call",
+        summary="Replace `hasattr` Call with `callable`",
```
drdavella:
```suggestion
        summary="Use `callable` builtin to check for callables",
```

drdavella (dismissed):
Looks good, just a suggestion/request for changing the summary.

andrecsilva (approved):

PR #330, opened 2024-03-05
enable isort with black profile

### LINKED ISSUE #320: Add isort to project
It would be good to finally bite the bullet on `isort` and enable it. I think this involves the following steps:

* Add a pre-commit step
* Reformat the code using `isort`

clavedeluna:
## Overview
*Format codemodder imports with isort, black compatible*

## Description

* making it black compatible https://github.com/PyCQA/isort/issues/1518

Closes #320


drdavella (approved):
❤️ I think this is great. I'm approving but let's hold off on merging for a sec to make sure there aren't going to be major conflicts with other ongoing work.

clavedeluna:
adding to queue bc this is a good time to do so. the only other thing in the queue doesn't touch these files, it should fail anyway

andrecsilva (approved):

PR #336, opened 2024-03-06
Add test case for changed lined numbers in use-walrus-if

drdavella:
## Overview
*Add (failing) test case for computed line numbers in `use-walrus-if`*

## Details
* We need to make an update so that these change positions are computed properly

### DIFF
--- a/tests/codemods/test_walrus_if.py
+++ b/tests/codemods/test_walrus_if.py
@@ -183,7 +183,7 @@ if val is not None:
         self.run_and_assert(tmpdir, input_code, expected_output)
 
     @pytest.mark.xfail(reason="Need to recompute line numbers to reflect removed lines")
-    def test_multiple_warlus_changes(self, tmpdir):
+    def test_multiple_walrus_changes(self, tmpdir):
         input_code = """
         val = do_something()
         if val is not None:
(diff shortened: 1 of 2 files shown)


drdavella:
Bumped the unit test timeout since we keep hitting on this for the Python 3.12 tests. See #338.

tests/codemods/test_walrus_if.py
```diff
@@ -181,3 +181,35 @@ def func(y):
                 print("whatever", y)
             """
         self.run_and_assert(tmpdir, input_code, expected_output)
+
+    @pytest.mark.xfail(reason="Need to recompute line numbers to reflect removed lines")
+    def test_multiple_warlus_changes(self, tmpdir):
```
andrecsilva: warlus -> walrus

andrecsilva (commented):
I'm not sure what is the problem here. The current behavior seems to be correct as it reports the node positions of the `if`s in the original code.

drdavella:
@andrecsilva the line locations here affect the position of the comments in CodeTF. In general we frame our comments in terms of the fix rather than the problem and we want the comments to appear on the right side of the split diff, which means they need to be associated with new code positions.

clavedeluna (approved):

PR #350, opened 2024-03-10
New codemod to fix dataclass defaults

### LINKED ISSUE #250: Fix `dataclass` defaults
See this StackOverflow issue for a description of the problem: https://stackoverflow.com/questions/53632152/why-cant-dataclasses-have-mutable-defaults-in-their-class-attributes-declaratio

It would be useful to have a codemod that could fix this proactively. Even though it results in a runtime error, if it lives on a codepath that is not thoroughly tested, it might get missed.

clavedeluna:
## Overview
*This codemod will fix dataclass default values for list, dict, set, and replace it with the correct syntax*

## Description

* I left some todos and a test for future enhancement to also fix when the default values are populated. I think the right way is to use lambda? Haven't fully researched

Closes #250


andrecsilva (approved):

src/core_codemods/docs/pixee_python_fix-dataclass-defaults.md
```diff
@@ -0,0 +1,18 @@
+This codemod will fix instances of `dataclasses.dataclass` that define default lists, sets, or dicts which raise a runtime `ValueError`. The [dataclass documentation](https://docs.python.org/3/library/dataclasses.html#mutable-default-values) provides a clear explanation of why this code is disallowed and explains how to use `field(default_factory=...)` instead.
```
drdavella:
```suggestion
When defining a Python dataclass it is not safe to use mutable datatypes (such as `list`, `dict`, or `set`) as defaults for the attributes. This is because the defined attribute will be shared by all instances of the dataclass type. Using such a mutable default will ultimately result in a `ValueError` at runtime. This codemod updates attributes of `dataclasses.dataclass` with mutable defaults to use `dataclasses.field` instead. The [dataclass documentation](https://docs.python.org/3/library/dataclasses.html#mutable-default-values) providesmore details about why using `field(default_factory=...)` is the recommended pattern.
```

drdavella (dismissed):
Really nice codemod overall, thanks! Just requesting an update on the description.

PR #355, opened 2024-03-11
New codemod to add missing `self`/`cls`

### LINKED ISSUE #297: Add missing `self` or `cls` parameter to method
This is based on a Sonar rule: https://rules.sonarsource.com/python/RSPEC-5719/

We should implement this as a find and fix codemod but also as a Sonar codemod.

clavedeluna:
## Overview
*A codemod that checks if instance/class methods without any args need self/cls*

## Description

* This codemod follows [the sonar rule](https://rules.sonarsource.com/python/RSPEC-5719/) exactly, without adding any other features other than the edge cases for `__new__` and `__init_subclass__` sonar mentions
* However, while working on this I thought about additional steps we could take, either in this codemod or as a separate codemod:
1. As is, we're only checking methods/class methods with no args at all. However, what about those which have args but none are `self` / `cls`?
2.  As is, when a method has no args, we just add `self`. However, another valid change would be to make it a `@staticmethod`. For example:
```
def A(): 
	print("instance_method")
	# `self` is never used in the method, we could make this a staticmethod

def B(): 
	self.hi = 1
        # `self` is used, so this should have `self` as an arg
```

Right now we handle both cases the same: just add `self`. However, we could decide to handle for case `A` by checking if `self` is ever used. Could be a bit tricky. It is less risky to just add `self` than it is to incorrectly assume `self` is never used and making it a staticmethod.

Closes #297

### DIFF
--- a/src/core_codemods/docs/pixee_python_fix-missing-self-or-cls.md
+++ b/src/core_codemods/docs/pixee_python_fix-missing-self-or-cls.md
@@ -1,0 +1,15 @@
+Python instance methods must be defined with `self` as the first argument. Likewise, class methods must have `cls` as the first argument. This codemod will add these arguments the method/class method has no arguments defined.
+
+Our changes look something like this:
+
+```diff
+ class MyClass:
+-    def instance_method():
++    def instance_method(self):
+         print("instance_method")
+
+     @classmethod
+-    def class_method():
++    def class_method(cls):
+         print("class_method")
+```
--- a/src/core_codemods/docs/pixee_python_fix-missing-self-or-cls.md
+++ b/src/core_codemods/docs/pixee_python_fix-missing-self-or-cls.md
@@ -1,4 +1,4 @@
-Python instance methods must be defined with `self` as the first argument. Likewise, class methods must have `cls` as the first argument. This codemod will add these arguments the method/class method has no arguments defined.
+Python instance methods must be defined with `self` as the first argument. Likewise, class methods must have `cls` as the first argument. This codemod will add these arguments when the method/class method has no arguments defined.
 
 Our changes look something like this:
 
--- a/src/core_codemods/fix_missing_self_or_cls.py
+++ b/src/core_codemods/fix_missing_self_or_cls.py
@@ -4,51 +4,50 @@ from codemodder.codemods.libcst_transformer import (
     LibcstResultTransformer,
     LibcstTransformerPipeline,
 )
-from codemodder.codemods.utils_mixin import NameResolutionMixin
+from codemodder.codemods.utils_mixin import NameAndAncestorResolutionMixin
 from core_codemods.api import Metadata, ReviewGuidance
 from core_codemods.api.core_codemod import CoreCodemod
 
 
-class FixMissingSelfOrClsTransformer(LibcstResultTransformer, NameResolutionMixin):
+class FixMissingSelfOrClsTransformer(
+    LibcstResultTransformer, NameAndAncestorResolutionMixin
+):
     change_description = "Add `self` or `cls` parameter to instance or class method."
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.current_class_name = None
-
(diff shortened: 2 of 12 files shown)


tests/codemods/test_fix_missing_self_or_cls.py
```diff
@@ -0,0 +1,110 @@
...
+            def clsmethod(one, two):
+               pass
+        
+            @classmethod
+            def kls(**kwargs):
+               pass
+           """
+        self.run_and_assert(tmpdir, input_code, input_code)
+
+    # def test_exclude_line(self, tmpdir):
```
drdavella: Dead code? Can we just `xfail` if this is a TODO?

src/core_codemods/docs/pixee_python_fix-missing-self-or-cls.md
```diff
@@ -0,0 +1,15 @@
+Python instance methods must be defined with `self` as the first argument. Likewise, class methods must have `cls` as the first argument. This codemod will add these arguments the method/class method has no arguments defined.
```
drdavella:
```suggestion
Python instance methods must be defined with `self` as the first argument. Likewise, class methods must have `cls` as the first argument. This codemod will add these arguments when the method/class method has no arguments defined.
```

src/core_codemods/fix_missing_self_or_cls.py
```diff
@@ -0,0 +1,88 @@
...
+                node.params.posonly_params,
+            )
+        )
+
+
+FixMissingSelfOrCls = CoreCodemod(
+    metadata=Metadata(
+        name="fix-missing-self-or-cls",
+        review_guidance=ReviewGuidance.MERGE_WITHOUT_REVIEW,
+        summary="Fix Missing Positional Parameter for Instance and Class Methods",
```
drdavella:
```suggestion
        summary="Add Missing Positional Parameter for Instance and Class Methods",
```

src/core_codemods/fix_missing_self_or_cls.py
```diff
@@ -0,0 +1,88 @@
...
+                node.params.star_kwarg,
+                node.params.posonly_params,
+            )
+        )
+
+
+FixMissingSelfOrCls = CoreCodemod(
+    metadata=Metadata(
+        name="fix-missing-self-or-cls",
+        review_guidance=ReviewGuidance.MERGE_WITHOUT_REVIEW,
```
drdavella: Given the scenarios you mentioned above, maybe this should be cursory review?

drdavella (dismissed):
Looks really nice overall, just a few minor comments.

To address your points:
* I think it's okay to use the lack of `@staticmethod` to infer intent here, but obviously there might be a small set of cases where we'd be wrong
* Doing this when other arguments are present would be really cool but it does seem challenging. Unfortunately since `self` and `cls` are not actually keywords it becomes pretty hard to tell what was actually intended in those cases, although probably in 99% of cases our fix would be okay.

src/core_codemods/fix_missing_self_or_cls.py
```diff
@@ -0,0 +1,88 @@
...
+    def leave_ClassDef(
+        self, original_node: cst.ClassDef, updated_node: cst.ClassDef
+    ) -> cst.ClassDef:
+        self.current_class_name = None
+        return updated_node
+
+    def leave_FunctionDef(
+        self, original_node: cst.FunctionDef, updated_node: cst.FunctionDef
+    ) -> cst.FunctionDef:
+        if self.current_class_name:
```
andrecsilva: I think you need to check if this def is directly in the classdef body. The way it is right now, I think it would also change inner functions in nested declarations. For example, in the following:
```python
class A():
  def f(self):
    def g():
      pass
```
I think the codemod would change `g()`.
In any case, it warrants a test case.

andrecsilva (changes requested):

clavedeluna:
Reviewers: note the change in implementation to correctly handle nested functions.

andrecsilva (approved):

PR #356, opened 2024-03-11
chore: add ssh repo deploy key option to automated PRs

### LINKED ISSUE #322: pre-commit autoupdate PR should run tests/lint
[This PR](https://github.com/pixee/codemodder-python/pull/321) can't be directly merged because other GH workflows don't get run. This is actually a known thing.  [This issue](https://github.com/peter-evans/create-pull-request/issues/48) lists workarounds

Right now we're pulling the branch locally and force pushing to get the test/lint workflows to trigger.

maxArturo:
## Overview
Closes https://github.com/pixee/codemodder-python/issues/322

## Description

Per the [docs](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs) of `peter-evans/create-pull-request` GH action (and github in general) triggering an automated PR doesn't cascade down into other GH actions, which makes sense for sanity reasons. 

The proposed [workaround](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#push-using-ssh-deploy-keys) is to create a repo-scoped SSH [deploy key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys#deploy-keys) and use that in the GH action instead. 

This PR adds a proposed secret var, `SSH_REPO_DEPLOY_KEY`, which must be added in the settings of the repo *BEFORE* this PR is merged. 

## Additional Details
See warning above: the ssh deploy repo-scoped `SSH_REPO_DEPLOY_KEY` ⚠️ must ⚠️  be added before this is merged.

Additionally, note that with this workaround only the `on: push` workflows will be triggered. This should be good enough for the lint/testing workflows.

### DIFF
--- a/.github/workflows/pre-commit-autoupdate.yml
+++ b/.github/workflows/pre-commit-autoupdate.yml
@@ -12,11 +12,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          ssh-key: ${{ secrets.SSH_REPO_DEPLOY_KEY }}
       - uses: actions/setup-python@v5
       - uses: browniebroke/pre-commit-autoupdate-action@main
       - uses: peter-evans/create-pull-request@v6
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.SSH_REPO_DEPLOY_KEY }}
           branch: update/pre-commit-hooks
           title: Update pre-commit hooks
           commit-message: "chore: update pre-commit hooks"


drdavella:
Hi @maxArturo thank you for your contribution!

I think I understand the rationale for this change but I'm trying to understand the tradeoffs between the different options mentioned in the documentation that you linked.

I'm hung up on the restriction to `on: push`: doesn't this mean that the checks still won't run when the PR is _created_? But only when it's _merged_? Maybe that is indeed good enough in principle but it seems like the checks are currently preventing the PR from getting merged at all, and this solution would not fix that problem. But maybe I'm missing something.

maxArturo:
Hi @drdavella!

> I'm hung up on the restriction to on: push:

Yes, I did re-read and I see this 'action' is squarely aimed at dependency version updates. Your concern on the `push` trigger is valid and will likely not work. 

> I'm trying to understand the tradeoffs

Here are the token tradeoffs you need to consider so that a `on: pull-request` works as you want: 

- a [PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) seems heavy-handed IMHO since it is personally-linked and has a lot of drawbacks;
- a GH [app-generated token](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens) seems like the better alternative, but now you have to setup the app and keep docs/make sure everyone knows about this workflow.
- the machine version token alternative just seems like a PAT with extra steps 🤔 

I'd make a final observation: these complications seem to stem from going against the grain of what GH recommends, which is [dependabot-administered actions](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#about-dependabot-and-github-actions). You can schedule these on a daily basis and it works with `pip` dep updates.

There's a lot of documentation around it, and the one caveat I see for triggering [downstream workflows](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#responding-to-events) is: 

> For workflows initiated by Dependabot (github.actor == 'dependabot[bot]') using the pull_request, pull_request_review, pull_request_review_comment, push, create, deployment, and deployment_status events, the following restrictions apply:
>
> - GITHUB_TOKEN has read-only permissions by default.
> - Secrets are populated from Dependabot secrets. GitHub Actions secrets are not available.

So unless you're publishing to a registry or need secrets for some other workflow then it seems like a viable alternative. What do you think?

drdavella:
> these complications seem to stem from going against the grain of what GH recommends

This is a good observation. It does seem like this is potentially the correct solution to this problem.

> GITHUB_TOKEN has read-only permissions by default

This shouldn't be a problem since the action permissions can be adjusted.

> Secrets are populated from Dependabot secrets. GitHub Actions secrets are not available.

I don't think this is a problem either since we only rely on secrets for our PyPI deployment.

maxArturo:
@drdavella it shouldn't be hard to add a Dependabot PR + PoC. I'll close this PR in favor of the `dependabot` approach. Thanks!

PR #357, opened 2024-03-11
Use Pydantic to generate and validate CodeTF data models

drdavella:
## Overview
*Use `pydantic` to generate and validate CodeTF models*

## Description

* Using [pydantic](https://docs.pydantic.dev/latest/) is a good way to express data models using type hints and to enable validation
* This enables straightforward representation of the CodeTF data model and provides JSON generation for free
* These changes mean our data model is now precisely aligned with the CodeTF spec
* Now that we have a JSON Schema for CodeTF I added validation for our integration tests: https://github.com/pixee/codemodder-specs/blob/main/codetf.schema.json


andrecsilva (approved):

integration_tests/conftest.py
```diff
@@ -0,0 +1,11 @@
+import json
+
+import pytest
+import requests
+
+
+@pytest.fixture(scope="module")
```
clavedeluna, line 7: This new fixture means every time a test module runs we're making a request. I see this is also defined below, so now both unit tests and integration tests rely on internet connectivity. Just pointing that out.
Do we need it per module, or could this fixture run per test run session only?

src/codemodder/codetf.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class Result(BaseModel):
+    codemod: str
+    summary: str
+    description: str
+    references: Optional[list[Reference]] = None
+    properties: Optional[dict] = None
+    failedFiles: Optional[list[str]] = None
+    changeset: list[ChangeSet]
```
clavedeluna, line 82: should this have `= []`?

integration_tests/conftest.py
drdavella: @clavedeluna good catch. I did think about making it session scoped but I think we're running unit tests and integration tests in separate workflows so it would run twice anyway.

src/codemodder/codetf.py
drdavella: Hmm I don't think it's strictly necessary since this is a required field but it might be a good idea.

Issue #301, opened 2024-02-23
Handle SQL queries that are formatting using string format operator


drdavella:
We need to enhance the SQL parameterization codemod so that it can handle queries that are (improperly) formatted using the string format operator.

PR #361, opened 2024-03-12
Adds support for format operators in SQLQueryParameterization

andrecsilva:
## Overview
`SQLQueryParameterization` will now parameterize queries built with the format operator `%`

## Description

- `SQLQueryParameterization` will now correctly parameterize queries built with the format operator `%`;
- Added `LinearizeStringExpression`. It takes a string expression and finds all the pieces that composes that string. For example: `"1" + a + "2"` will return a list with the nodes that represents `"1"`, `a` and `"2"`;
- Added `RemoveUnusedVariables`. It removes local assignments that are not referenced anywhere else;
- Added several utilities for parsing printf style strings in `format_string_parser` module;

Tackles #301.

### DIFF
--- a/src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
+++ b/src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
@@ -27,8 +27,9 @@ class RemoveEmptyStringConcatenation(CSTTransformer):
     def leave_BinaryOperation(
         self, original_node: cst.BinaryOperation, updated_node: cst.BinaryOperation
     ) -> cst.BaseExpression:
-        if isinstance(original_node.operator, cst.Add):
-            return self.handle_node(updated_node)
+        match original_node.operator:
+            case cst.Add():
+                return self.handle_node(updated_node)
         return updated_node
 
     def leave_ConcatenatedString(
@@ -43,20 +44,21 @@ class RemoveEmptyStringConcatenation(CSTTransformer):
     ) -> cst.BaseExpression:
         left = updated_node.left
         right = updated_node.right
-        if self._is_empty_string_literal(left):
-            if self._is_empty_string_literal(right):
+        if _is_empty_string_literal(left):
+            if _is_empty_string_literal(right):
                 return cst.SimpleString(value='""')
             return right
-        if self._is_empty_string_literal(right):
-            if self._is_empty_string_literal(left):
+        if _is_empty_string_literal(right):
+            if _is_empty_string_literal(left):
                 return cst.SimpleString(value='""')
             return left
         return updated_node
 
-    def _is_empty_string_literal(self, node):
-        match node:
-            case cst.SimpleString() if node.raw_value == "":
-                return True
-            case cst.FormattedString() if not node.parts:
-                return True
-        return False
+
+def _is_empty_string_literal(node):
+    match node:
+        case cst.SimpleString() if node.raw_value == "":
+            return True
+        case cst.FormattedString() if not node.parts:
+            return True
+    return False
--- a/src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
+++ b/src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
@@ -5,6 +5,7 @@ from libcst import RemovalSentinel, SimpleString
 from libcst.codemod import ContextAwareTransformer
 
 from codemodder.codemods.utils_mixin import NameAndAncestorResolutionMixin
+from codemodder.utils.utils import is_empty_string_literal
 
 
 class RemoveEmptyStringConcatenation(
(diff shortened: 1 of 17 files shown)


src/codemodder/codemods/transformations/remove_empty_string_concatenation.py
```diff
@@ -41,20 +50,21 @@ def handle_node(
...
 
-    def _is_empty_string_literal(self, node):
-        match node:
-            case cst.SimpleString() if node.raw_value == "":
-                return True
-            case cst.FormattedString() if not node.parts:
-                return True
-        return False
+
+def _is_empty_string_literal(node):
```
clavedeluna: same function as the other file so move to a utility module and remove the leading `_` since it won't be private

src/codemodder/utils/linearize_string_expression.py
```diff
@@ -0,0 +1,210 @@
...
+        self.leaves.append(node)
+        return False
+
+    # recursive search
+    def visit_Name(self, node: cst.Name) -> Optional[bool]:
+        self.leaves.extend(self.recurse_Name(node))
+        return False
+
+    def visit_Attribute(self, node: cst.Attribute) -> Optional[bool]:
+        # TODO try to resolve this
```
clavedeluna: this todo doesn't make a lot of sense without more context, maybe remove it?

src/core_codemods/sql_parameterization.py
```diff
@@ -39,10 +56,57 @@
...
+                return parent.start.lower()
+            case PrintfStringText():
+                return self.extract_prefix(node.origin)
+        return ""
+
+    def _extract_prefix_raw_value(self, node: StringLiteralNodeType) -> tuple[str, str]:
+        raw_value = extract_raw_value(node)
+        if (prefix := self.extract_prefix(node)) is not None:
+            return prefix, raw_value
+        return prefix, raw_value
```
clavedeluna, line 81: maybe I'm missing something super subtle but can't you just
```suggestion
prefix = self.extract_prefix(node))
return prefix, raw_value
```
no if check, no double return

src/core_codemods/sql_parameterization.py
```diff
@@ -255,6 +348,19 @@ def _remove_literal_and_gather_extra(
                 self.changed_nodes[original_node] = updated_node.with_changes(
                     value=new_value
                 )
+            case PrintfStringText():
+                if extra_raw_value:
+                    extra = cst.SimpleString(
+                        value=("r" if "r" in prefix else "")
+                        + "'"
+                        + extra_raw_value
+                        + "'"
```
clavedeluna: probably cleaner to make it an f-str instead of using +

clavedeluna (changes requested):

src/codemodder/utils/format_string_parser.py
```diff
@@ -0,0 +1,187 @@
...
+    expr_dict: dict[cst.BaseExpression, cst.BaseExpression]
+) -> dict[str | cst.BaseExpression, cst.BaseExpression]:
+    return {
+        extract_raw_value(k): v
+        for k, v in expr_dict.items()
+        if isinstance(k, cst.SimpleString | cst.FormattedStringText)
+    }
+
+
+def parse_formatted_string(
```
drdavella, line 140: This is pretty heroic, nice work. I could have sworn that there was something in the standard library that helped to tokenize format strings but I'm having trouble digging it up right now.

src/core_codemods/sql_parameterization.py
```diff
@@ -98,35 +164,32 @@ def _build_param_element(self, prepend, middle, append, aliased_expr):
...
-        # (1) FindQueryCalls - Find and gather all the SQL query execution calls. The result is a dict of call nodes and their associated list of nodes composing the query (i.e. step (2)).
-        # (2) LinearizeQuery - For each call, it gather all the string literals and expressions that composes the query. The result is a list of nodes whose concatenation is the query.
-        # (3) ExtractParameters - Detects which expressions are part of SQL string literals in the query. The result is a list of triples (a,b,c) such that a is the node that contains the start of the string literal, b is a list of expressions that composes that literal, and c is the node containing the end of the string literal. At least one node in b must be "injectable" (see).
-        # (4) SQLQueryParameterization - Executes steps (1)-(3) and gather a list of injection triples. For each triple (a,b,c) it makes the associated changes to insert the query parameter token. All the expressions in b are then concatenated in an expression and passed as a sequence of parameters to the execute call.
-
-        # Steps (1) and (2)
+        # (1) FindQueryCalls -> (2) ExtractParameters -> (3) SQLQueryParameterization
```
drdavella: Just a nitpick but this should probably be a docstring instead of a comment.

drdavella (approved):
This looks great overall. It's a lot of very impressive work, thanks for the effort. 

I'm starting to see how some of these pieces could be taken and reused for other kinds of codemods (e.g. xpath injection).

clavedeluna (approved):

PR #365, opened 2024-03-13
Add pixeebot config; enable use-walrus-if

drdavella:
## Overview
*Add pixeebot config; enable `use-walrus-if`*


clavedeluna (approved):
sorry must be behind on the use for this, can you explain?

PR #366, opened 2024-03-13
Add detection tool metadata to CodeTF results

drdavella:
## Overview
*Add detection tool metadata to CodeTF results*

## Description

* The CodeTF spec introduced a new `detectionTool` field: https://github.com/pixee/codemodder-specs/pull/21
* This metadata will be used by downstream tooling for providing consistent presentation for tool remediation results
* We will populate the `findings` field later since that is not needed right now

### DIFF
--- a/src/codemodder/codemods/base_codemod.py
+++ b/src/codemodder/codemods/base_codemod.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import functools
 import importlib.resources
 from abc import ABCMeta, abstractmethod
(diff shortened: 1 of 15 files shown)


src/codemodder/codemods/base_codemod.py
```diff
@@ -31,6 +33,15 @@ class Metadata:
     review_guidance: ReviewGuidance
     references: list[Reference] = field(default_factory=list)
     description: str | None = None
+    tool: ToolMetadata | None = None
+
+
+@dataclass
+class ToolMetadata:
```
clavedeluna, line 40: since this is codetf related should it not be a pydantic class?

clavedeluna (approved):

src/codemodder/codemods/base_codemod.py
drdavella: Good question: this is sort of intermediate metadata that isn't directly used for CodeTF generation so I think `dataclass` is okay.

PR #380, opened 2024-03-14
`ToolMetadata.rule_url` should be optional; expose via API

drdavella:


drdavella:
Just FYI but I'm going to backport this onto 0.85.x

CRRogo (approved):

PR #381, opened 2024-03-14
Fix metadata key error

### LINKED ISSUE #378: Investigate Error When Executing RemoveUnnecessaryFStr after UnnecessaryFormatString
Currently executing the `RemoveUnnecessaryFStr` codemod after a `UnnecessaryFormatString` transformation will cause a `KeyError`. This may happen when running tests, as some codemods will use `UnnecessaryFormatString`.

Below is a snippet that reproduces this issue.

```python
code = dedent("""\
print('hello')
""")
tree = cst.parse_module(code)
file_context = FileContext(Path('.'),Path('.'),[],[],None)
context = CodemodContext()
result = UnnecessaryFormatString(context).transform_module(tree2)
ruf = RemoveUnnecessaryFStr()
ruf.transformer.transformers[0].transform(result,None,file_context)
```

drdavella:
## Overview
*Fix metadata key error with `RemoveUnnecessaryFStr codemod*

## Description

* Fixes #378 
* I believe this is due to an issue with multiple inheritance when multiple LibCST transforms are parents of the same codemod class
* LibCST is doing some ~funny~ clever stuff with caching the metadata dependencies, and multiple inheritance seems to break some assumptions
* I'm not sure why this only manifested recently but this change seems to resolve the issue

### DIFF
--- a/integration_tests/test_unnecessary_f_str.py
+++ b/integration_tests/test_unnecessary_f_str.py
@@ -1,5 +1,3 @@
-import pytest
-
 from codemodder.codemods.test import (
     BaseIntegrationTest,
     original_and_expected_from_code_path,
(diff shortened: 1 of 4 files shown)


andrecsilva (changes requested):
Good work on figuring this out. You should also re-enable the unit tests.

andrecsilva (approved):

clavedeluna (approved):

PR #384, opened 2024-03-16
Codemodder can handle non-existent requested codemods

### LINKED ISSUE #346: Requesting a non-existent codemod should be a warning and not a hard error
https://github.com/pixee/codemodder-specs/pull/18

clavedeluna:
## Overview
*Codemodder now does not stop with an error if you pass in an unknown codemod for include or exclude*

## Description

* Behavior up to this point in codmodder is that argparse would check if any of the codemod names you pass in are not know. If any were not known, it would just stop without running any codemods that are known.
* To correctly handle this new behavior of continuing on if you requested known codemods, I moved all the logic out of the argparse cli section and put it in a more appropriate place, in `match_codemods` which already does all the include/exclude check logic.
Now codemodder output will look like this
```
codemodder tests/samples/ --output here.txt --codemod-include=secured

[startup]
codemodder: python/0.84.1.dev25+g7f5d2ca.d20240313
command: codemodder tests/samples/ --output here.txt --codemod-include=secured
Requested codemod to include'secured' does not exist.
...
...
```

closes #346

### DIFF
--- a/src/codemodder/registry.py
+++ b/src/codemodder/registry.py
@@ -27,6 +27,20 @@ class CodemodCollection:
     codemods: list
 
 
+class TupledKeysDict(dict):
+    # todo: document
+    def get(self, key):
+        for key_tuple, value in self.items():
+            if key in key_tuple:
+                return value
+        raise KeyError
+
+    def pop(self, key):
+        for key_tuple in list(self.keys()):
+            if key in key_tuple:
+                del self[key_tuple]
+
+
 class CodemodRegistry:
     _codemods_by_name: dict[str, BaseCodemod]
     _codemods_by_id: dict[str, BaseCodemod]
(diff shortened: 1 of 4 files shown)


src/codemodder/registry.py
```diff
@@ -27,6 +27,24 @@ class CodemodCollection:
...
+            if key in key_tuple:
+                return value
+        raise KeyError
+
+    def pop(self, key):
+        for key_tuple in list(self.keys()):
+            if key in key_tuple:
+                del self[key_tuple]
+
+
```
andrecsilva: I understand that this is a quick solution for simple mapping, and we expect that mapping to be small, but both these operations are downgraded from (amortized) O(1) to O(n). 
One suggestion is to store the keys within the value itself, so `d[codemod_name] = d[codemod_id] = (codemod.id, codemod.name, codemod)`. Deleting or `pop` would be a matter of reading the stored tuple and deleting each entry.

andrecsilva (approved):

src/codemodder/registry.py
clavedeluna: Good point here, I was meant to come back to think thru complexity but forgot to do so.

While your solution is more efficient, I find it really confusing, and confusing leads to harder to maintain and more bugs. My solution was based on the fact that the original code here was already iterating, and iterating TWICE. See  [here](https://github.com/pixee/codemodder-python/blob/main/src/codemodder/registry.py#L66) and [here](https://github.com/pixee/codemodder-python/blob/main/src/codemodder/registry.py#L74). So I wanted to reduce the iteration to just once over all codemods, and just once over the exclude names list.

Your suggestion would be more efficient, but would look like this:
```
codemods = {}
codemods["secure_random"] = codemods["secure_random_id"] = ("secure_random_id", "secure_random", "secure random codemod")

# dict is now
{'secure_random': ('secure_random_id', 'secure_random', 'secure random codemod'), 'secure_random_id': ('secure_random_id', 'secure_random', 'secure random codemod')}

exclude = "secure_random_id"
value = codemods.pop(exclude)
# assuming value is not None

codemods.pop(value[0]) # need to catch keyerror in case value[0] was exclude
codemods.pop(value[1]) # need to catch keyerror in case value[1] was exclude
```

So yes it gets the job done and I could just make it a separate function and heavily document it to explain what's going on, but the dict solution was more pythonic (to me) and easier to read.

Maybe Dan has a tie-breaker vote here.

drdavella: As written I think this is currently equivalent to the following:
```python
base_codemods = dict()
for codemod in self.codemods:
    base_codemods[codemod.id] = codemod
    base_codemods[codemod.name] = codemod
```

This would enable O(1) lookup both by name and by ID. In either case we have a different problem now which is that two codemods could have the same `name` even if they have different IDs. It's nice to be able to look up _core_ codemods by name but I'm going to propose that all other codemods should only be keyed by ID.

tl;dr: I think we can just use a dict.

clavedeluna: Yeah doing with a dict is ok, the code would look like this:

```
        base_codemods = dict()
        for codemod in self.codemods:
            if (sast_only and codemod.origin != "pixee") or (not sast_only and codemod.origin == "pixee"):
                base_codemods[codemod.id] = codemod
                base_codemods[codemod.name] = codemod

        if codemod_exclude and not codemod_include:
            for name_or_id in codemod_exclude:
                try:
                    codemod = base_codemods[name_or_id]
                except KeyError:
                    logger.warning(
                        f"Requested codemod to exclude'{name_or_id}' does not exist."
                    )
                    continue
                # have to do both bc I don't know which it was in `name_or_id`
                base_codemods.pop(codemod.name, None)
                base_codemods.pop(codemod.id, None)
            return list(set(base_codemods.values()))
```

> but I'm going to propose that all other codemods should only be keyed by ID.

@drdavella ok to do this as a separate issue?

drdavella: @clavedeluna yes good for me as long as we create an issue for it.

clavedeluna: https://github.com/pixee/codemodder-python/issues/388

PR #386, opened 2024-03-18
chore: add pre-commit downstream workflows

maxArturo:
## Overview
- If there's a `pre-commit` update:
    - "require" and run the [`lint`](./github/workflows/lint.yml) and [`test`](./github/workflows/test.yml) workflows, avoiding any direct recursive calls to GH actions

## Description

Based on our [previous conversation](https://github.com/pixee/codemodder-python/pull/356#issuecomment-1995675603) I looked at how to get this running with dependabot, but I quickly realized it wouldn't work as expected. 

It would only run when a `pip` dependency changed, so effectively it wouldn't run the specific code that the current GH action [runs](https://github.com/browniebroke/pre-commit-autoupdate-action/blob/1e32b32e531346c631e93007c766971301367ec9/action.yml#L11-L14) with a daily frequency.

Also, using `dependabot` with `pre-commit` is very low on the GH team's [priority](https://github.com/dependabot/dependabot-core/issues/1524). 

So I decided to "unfurl" the needed downstream actions as direct dependencies on the "pre-commit" job, by "requiring" them and conditionally running linting and testing if there's a `pre-commit` update. This works because we can see if the pr was [created or updated](https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-behaviour) in the outputs of the create-pr action. 

### PROOF:
You can see the [successful run](https://github.com/maxArturo/codemodder-python/actions/runs/8326875669) and then the lint/test actions being triggered afterwards.

## Additional Details
The one thing to keep in mind is if you want something else to conditionally run based on `pre-commit update`, you have to add it here as a direct downstream dependency. That's it!

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,6 +10,7 @@ on:
       - main
       - develop
   merge_group:
+  workflow_call:
 
 concurrency:
   group: (${{ github.workflow }}-${{ github.event.inputs.branch || github.event.pull_request.head.ref }})
--- a/.github/workflows/pre-commit-autoupdate.yml
+++ b/.github/workflows/pre-commit-autoupdate.yml
@@ -6,6 +6,7 @@ on:
     - cron: "0 0 * * *"
   # on demand
   workflow_dispatch:
+  push:
 
 jobs:
   auto-update:
--- a/.github/workflows/pre-commit-autoupdate.yml
+++ b/.github/workflows/pre-commit-autoupdate.yml
@@ -35,11 +35,12 @@ jobs:
           branch: update/pre-commit-hooks
           title: Update pre-commit hooks
           commit-message: "chore: update pre-commit hooks"
-          body: Update versions of pre-commit hooks to latest version.
+          body:
+            Update versions of pre-commit hooks to latest version.
             # output is called pull-request-operation
       - name: Set outputs
         id: check-diff
-        if: | 
+        if: |
           steps.pr-create.outputs.pull-request-number &&
           (steps.pr-create.outputs.pull-request-operation == 'created' ||
           steps.pr-create.outputs.pull-request-operation == 'updated')
@@ -47,15 +48,12 @@ jobs:
 
   lint-workflow:
     name: Run lint
-    uses: ./.github/workflows/lint.yml
-    if: ${{ needs.auto-update.outputs.diff }}
-    runs-on: ubuntu-latest
     needs: auto-update
-
+    if: ${{ needs.auto-update.outputs.diff }}
+    uses: ./.github/workflows/lint.yml
 
   test-workflow:
     name: run tests on diff
-    uses: ./.github/workflows/test.yml
-    if: ${{ needs.auto-update.outputs.diff }}
-    runs-on: ubuntu-latest
     needs: auto-update
+    if: ${{ needs.auto-update.outputs.diff }}
+    uses: ./.github/workflows/test.yml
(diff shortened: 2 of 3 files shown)


clavedeluna:
> You can see the [successful run](https://github.com/maxArturo/codemodder-python/actions/runs/8326875669) and then the lint/test actions being triggered afterwards.

But the unit tests [did not run](https://github.com/maxArturo/codemodder-python/actions/runs/8326875669/job/22783436787) ?

maxArturo:
I cancelled the tests because I didn’t  want to wait for them to finish, but otherwise it should work as expected. You can see that in the logs. https://productionresultssa12.blob.core.windows.net/actions-results/1b5b6d3c-d7b0-4923-873d-84837550d205/workflow-job-run-bed211fd-7dd0-5134-e289-e2d60b68d8e4/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-03-18T14%3A27%3A30Z&sig=NY%2FOB170%2BMjtxbE0KxgD3WU4KXsTnGCRnnzRJvCzWzY%3D&sp=r&spr=https&sr=b&st=2024-03-18T14%3A17%3A25Z&sv=2021-12-02MaxOn Mar 18, 2024, at 10:27 AM, Dani Alcala ***@***.***> wrote:﻿

You can see the successful run and then the lint/test actions being triggered afterwards.

But the unit tests did not run ?

clavedeluna:
IFF integration tests fail it's because of timeout, fixing it here https://github.com/pixee/codemodder-python/pull/387

clavedeluna (approved):
This is a reasonable approach to solve for the minor problem of having to manually pull and force push that auto-update commit.

Do want @drdavella  to give this approval before merge though.

drdavella:
@maxArturo thanks again for the contribution. I think this is pretty cool and it looks like it works. 

I recently discovered an action for checking diffs and I wonder whether it would help simplify this action at all? It's not a blocker but I'm curious looking ahead: https://github.com/technote-space/get-diff-action

I also wonder whether we can use a similar approach to make the PyPI release conditional on test execution. It's hard to do that right now because you can't filter a `workflow_run` action on a specific tag (only on a branch). Anyway, just an idle thought, maybe it doesn't make much sense.

drdavella (approved):

maxArturo:
Hi @drdavella my pleasure!

> I recently discovered an action for checking diffs and I wonder whether it would help simplify this action at all? It's not a blocker but I'm curious looking ahead: https://github.com/technote-space/get-diff-action

Oh that's neat! Looks like it's very full featured. I do worry about the fact that it's archived/frozen, and it might be heavy-handed for this use case. I think the `git diff` work is already being done by `peter-evans/create-pull-request`: if there's no changes to the repo I believe it won't create/update any PR's, which is why this is the conditional:

```yaml
       if: |
          (steps.pr-create.outputs.pull-request-operation == 'created' ||
          steps.pr-create.outputs.pull-request-operation == 'updated') &&
          steps.pr-create.outputs.pull-request-number
```
If we ever need a more advanced filter criteria (e.g a subset of files/folders) to conditionally execute, I'd recommend running a "straight up" custom `git diff` command and avoid pulling in an entire node/NPM project that will do almost the same! 😄 

> I also wonder whether we can use a similar approach to make the PyPI release conditional on test execution. It's hard to do that right now because you can't filter a `workflow_run` action on a specific tag (only on a branch). Anyway, just an idle thought, maybe it doesn't make much sense.

Could be! Looks like the releases are handled manually, so you could just call a `uses` from [this line](https://github.com/pixee/codemodder-python/blob/main/.github/workflows/deploy_to_pypi.yml#L7) and include whatever workflows you'd like to run _before_ the pypi release happens. 

I haven't had to do this before, but there's a good opportunity keep things DRY and group your `code sanity` GH workflows in an "omnibus" `yml` file. Then, you'd `uses` that in all the spots where you want a gate (at this point, here and on release it seems). There seems to be a balance between autotriggers (`on: push`, etc) and "manual" calling, but only the team members know best which fits where.

PR #393, opened 2024-03-19
Sonar tmpfile codemod

### LINKED ISSUE #371: Sonar: fix insecure temporary file creation with `secure-tempfile`
https://rules.sonarsource.com/python/type/Vulnerability/RSPEC-5445/

clavedeluna:
## Overview
*Added a sonar codemod for `secure-tempfile`*

## Description

* I also added an enforcement mechanism via assertions to ensure sonar_issues.json is updated for every new integration test. This will fail until #423 is merged since that PR fixes one codemod that had not updated this file. This demonstrates how it works.
* This particular codemod already had an entry in sonar_issues.json (I had forgotten to delete it when removing most of the files under tests/samples) so I just used that.

Closes #371


andrecsilva (approved):
Pending the test failures, this looks good to me.

src/codemodder/scripts/generate_docs.py
```diff
@@ -314,6 +314,11 @@ class DocMetadata:
         guidance_explained=CORE_METADATA["fix-missing-self-or-cls"].guidance_explained,
         need_sarif="Yes (Sonar)",
     ),
+    "secure-tempfile-S5445": DocMetadata(
```
drdavella, line 317: Not necessarily for this PR but we should be able to rewrite this whole section in terms of a dict comprehension and avoid a lot of duplication.

drdavella (approved):

PR #397, opened 2024-03-19
Fix `enable-jinja2-autoescape` when kwargs can't be known

drdavella:
## Overview
*Fix `enable-jinja2-autoescape` when kwargs can't be known*

## Description

* Handle case where contents of `**kwargs` can't be known
* If we add `autoescape` in this case and it's already present in `kwargs` then we are going to break the code

### DIFF
--- a/tests/codemods/test_enable_jinja2_autoescape.py
+++ b/tests/codemods/test_enable_jinja2_autoescape.py
@@ -117,6 +117,42 @@ class TestEnableJinja2Autoescape(BaseSemgrepCodemodTest):
     def test_autoescape_callable(self, tmpdir, code):
         self.run_and_assert(tmpdir, code, code)
 
+    def test_kwargs_unpacked(self, tmpdir):
+        input_code = (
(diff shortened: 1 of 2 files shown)


tests/codemods/test_enable_jinja2_autoescape.py
```diff
@@ -117,6 +117,42 @@ def test_autoescape_enabled(self, tmpdir):
     def test_autoescape_callable(self, tmpdir, code):
         self.run_and_assert(tmpdir, code, code)
 
+    def test_kwargs_unpacked(self, tmpdir):
+        input_code = (
+            expexted_output
+        ) = """
+        import jinja2
+        env = jinja2.Environment(**kwargs)
```
clavedeluna, line 125: I would've thought a more telling test case would be
`env = jinja2.Environment(**{"autoescape": True})` but I think it will trigger in all cases.

clavedeluna (approved):

PR #399, opened 2024-03-20
Use pip cache to speed up workflow runs

drdavella:
Not sure whether this will work, but it's worth a try 🤞


clavedeluna (approved):
> Not sure whether this will work

are you able to manually re-run all actions and see if there was a difference in time?

drdavella:
@clavedeluna that would be a good idea but I think we'll just YOLO it on the merge and see what happens.

PR #409, opened 2024-03-25
Revert "chore: add pre-commit downstream workflows"

clavedeluna:
Reverts pixee/codemodder-python#386

Unfortunately this didn't have the effect we wanted :/


drdavella (approved):

drdavella:
@maxArturo we appreciate the effort, thank you. This just seems much more complicated/annoying than any of us expected.

PR #410, opened 2024-03-25
Add some more debugging for SAST results

drdavella:
## Overview
*Add preemptive debug logs for SAST results*


clavedeluna (approved):
The test failures may be related if they didn't expect short-circuit or if they assert on log lines

clavedeluna:
@drdavella there isn't a reason why `fix-missing-self-or-cls` would be failing in this particular PR but not in main. Does it fail for you locally too? If anything, the current state of `fix-missing-self-or-cls` (missing the results filter) would make it pass when it should not.

clavedeluna:
rebasing to see if hopefully the failures are gone

PR #411, opened 2024-03-25
Parallelize integration tests

clavedeluna:
## Overview
*Be able to run all integration-tests in parallel for speed*

## Description

* As we add more codemods, running integration tests in parallel is really important as these take a long time one after the other.
* To do this, we had to move away from files in disk to testing with temp files /dirs. I was able to do this for all integration tests **except sonar tests**, which still require files in tests/samples because those align with sonar_issues.json. Later on hopefullly we can revisit the sonar issue here.
* So for non-sonar integration tests, all relevant files (code, output file, dependency file) are temp files. For sonar the output file is a tempfile too.
* As part of this work I also cleaned up quite a bit. One of my favorite things was making the `replacement_lines` be 1-index as this was confusing to see compared with `expected_line_change`. It didn't make sense to have these lines be 0-index when everything else for line changes is 1-indexed.
* **results**: locally integration tests ran for me in 30s. In actions it takes <5min bc only 4 workers are used.

## Additional Details
* As a result of this work I discovered that one of the most recently added sonar integration tests `test_sonar_fix_missing_self_or_cls` is missing the `is_node_selected` check, it passes when it shouldn't. This is because I hadn't updated sonar_issues.json because nothing reminded me to do so. I'll ticket this and thoughts around how to prevent this issue.
* We also want to figure out a better way to handle updating sonar_issues.json
* For some reason unit test coverage lowered by 0.05 or so only in github actions (locally it's still > 94) so I just lowered ithe cutoff slightly.


integration_tests/test_dependency_manager.py
```diff
@@ -1,20 +1,27 @@
...
 import pytest
 
-from codemodder.codemods.test.integration_utils import SAMPLES_DIR, CleanRepoMixin
+from codemodder.codemods.test.integration_utils import SAMPLES_DIR
 
 
-class TestDependencyManager(CleanRepoMixin):
-    output_path = "test-codetf.txt"
+class TestDependencyManager:
+    output_path = tempfile.mkstemp()[1]
```
drdavella: Would it make more sense to do this within `setup_class`?

src/codemodder/codemods/test/integration_utils.py
```diff
@@ -62,6 +46,43 @@ class BaseIntegrationTest(DependencyTestMixin, CleanRepoMixin):
...
+        cls.code_dir = tempfile.mkdtemp()
+
+        if not hasattr(cls, "code_filename"):
+            # Only a few codemods require the analyzed file to have a specific filename
+            # All others can just be `code.py`
+            cls.code_filename = "code.py"
+
+        cls.code_path = os.path.join(cls.code_dir, cls.code_filename)
+
+        if cls.code_filename == "settings.py" and "Django" in str(cls):
```
drdavella, line 59: It feels like maybe this should just happen in a Django-specific subclass.

drdavella (approved):

src/codemodder/codemods/test/integration_utils.py
clavedeluna: very very true. Honestly I was just trying to make as few changes to specific integration test files as possible. Just for now I'm going to leave it as is.

PR #419, opened 2024-03-27
Update `fickling` version; removes Torch dependency

### LINKED ISSUE #341: Bump to latest release of `fickling` to remove PyTorch dependency
This change has been merged: https://github.com/trailofbits/fickling/pull/95

Once the release is out we want to update our pinned version of `fickling` so that users do not have a dependency on PyTorch.

drdavella:
## Overview
*Update `fickling` dependency*

## Details
* This update removes the Torch dependency, see https://github.com/trailofbits/fickling/pull/95#issuecomment-2021303776
* I expect this to make our CI tests faster since Torch no longer needs to be installed

Closes #341

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,7 @@ test = [
     "django>=4,<6",
     "numpy~=1.26.0",
     "flask_wtf~=1.2.0",
-    "fickling~=0.1.0",
+    "fickling~=0.1.0,>=0.1.3",
 ]
 complexity = [
     "radon==6.0.*",
--- a/src/codemodder/dependency.py
+++ b/src/codemodder/dependency.py
@@ -92,10 +92,10 @@ Security = Dependency(
 )
 
 Fickling = Dependency(
-    Requirement("fickling~=0.1.0"),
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -67,7 +67,7 @@ test = [
     "django>=4,<6",
     "numpy~=1.26.0",
     "flask_wtf~=1.2.0",
-    "fickling~=0.1.0",
+    "fickling~=0.1.0,>=0.1.3",
```
clavedeluna, line 70: any reason not to just make it `>=0.1.3"`?

clavedeluna (approved):
could you update codemod_dependencies.txt? I'm probably the only one who remembers the existance of that file but. that's ok for now

pyproject.toml
drdavella: I think it's safer to keep it explicitly pinned to a particular minor version for now.

drdavella:
@clavedeluna I added the update. In the slightly longer term I think there might be a way to configure renovate to make the modification directly to our `dependency.py` file, but I need to do some investigation.

PR #420, opened 2024-03-27
Add safeguards for constant subprocess commands

### LINKED ISSUE #263: Do not apply `sandbox-process-creation` when arguments are literals
## Description
We may encounter cases where all of the command arguments to a `subprocess` call are hard-coded literals. In these cases hardening does not really make any sense and should be skipped.

A simple implementation would be to detect whether the immediate argument is a literal list that contains only literal strings. However a more sophisticated implementation would involve some constant propagation. As part of this issue we should explore the level of effort involved in the latter approach. If it's too difficult, implement the simpler approach for now and open another ticket.

This issue was initially described in https://github.com/pixee/python-security/issues/10

drdavella:
## Overview
*Prevent false positives when adding hardening for subprocess commands*

## Description

* We need to avoid hardening subprocess calls when the commands are hardcoded literals
* Fixes #263

### DIFF
--- a/integration_tests/test_process_sandbox.py
+++ b/integration_tests/test_process_sandbox.py
@@ -8,25 +8,29 @@ class TestProcessSandbox(BaseIntegrationTest):
     original_code = """
     import subprocess
 
-    subprocess.run("echo 'hi'", shell=True)
-    subprocess.run(["ls", "-l"])
+    cmd = " ".join(["ls"])
+
+    subprocess.run(cmd, shell=True)
+    subprocess.run([cmd, "-l"])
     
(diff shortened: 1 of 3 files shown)


clavedeluna (changes requested):
1. It's  a bit weird that the existing unit tests now have this "cmd" which isn't defined in the code snippet of the test. Do you think it makes sense to define it with an appropriate cmd value (argument passed from a func?)
2. If cmd as a func arg isn't a good simple case for ^, could you add a test case for `def func(cmd):  subprocess.run(cmd....`

clavedeluna (approved):

PR #423, opened 2024-03-29
Correct node position for FuncDef nodes and fix sonar codemods

### LINKED ISSUE #413: `fix-missing-self-or-cls` is missing `self.node_is_selected` check
All codemods should have logic similar to `self.node_is_selected` to check if results exist for this codemod. When writing this codemod we forgot to update sonar_issues.json which made the integration test for this codemod pass. However, adding the `self.node_is_selected` would've alerted us to having forgotten to add a corresponding sonar issue.

clavedeluna:
## Overview
*Add custom node_position for libcst.FuncDef nodes to support codemods that change code at function level*

## Description

* By default, libcst calculates all node start/end position. For simple nodes that span one line, the start and end line are the same. However, function def nodes have an end line and col position that belongs to the last line of the function, so:
```
# pos starts at `def` below
def func(arg):
     print(something) # pos ends here
```
However, for many codemods we really just care about matching the first line of the function def. So I've implemented this instead
```
# pos starts at `def` below
def func(arg):# pos ends here
     print(something) 
```
so the end position is the text length of the `def ...` line but index 0.

This now allows us to add include/exclude and results filtering to these codemods

## Additional Details
* I also decided to json-ify sonar_issues.json. 
*  was able to dedupe code
* future work includes handling other nodes like class def 

Closes #413


src/codemodder/codemods/base_visitor.py
```diff
@@ -48,7 +50,18 @@ def node_is_selected(self, node) -> bool:
...
+        match node:
+            case cst.FunctionDef():
+                # By default a function's position includes the entire
+                # function definition. Instead, we will only use the first line
+                # of the function definition.
+                params_end = self.get_metadata(PositionProvider, node.params).end
+                return CodeRange(
+                    start=self.get_metadata(PositionProvider, node).start,
+                    end=CodePosition(params_end.line, params_end.column + 1),
+                )
```
andrecsilva, line 62: I'd rater have a general solution instead making one for every node type. I suggest changing either includes/excludes match or `node_is_selected` to match match any node that begins on a given line.

clavedeluna: I'm not sure I agree with a more general solution like that. In that case, why do we care about start/end columns at all right now? Is there not a reason why we do column matching in `match_location` nearby code? If there is, then that more general solution of just matching start line doesn't make much sense.

andrecsilva (approved):
Since discussions about matching nodes are pending, I think this is ok for a temporary solution.

PR #433, opened 2024-04-02
Support sonar hotspots and secure-random sonar hotspot rule

### LINKED ISSUE #376: Sonar: fix using PNRGs with `secure-random`
https://rules.sonarsource.com/python/type/Security%20Hotspot/RSPEC-2245/

clavedeluna:
## Overview
*Add support for sonar hotspots and the first sonar hotspot, secure-random*

## Description

* Secure random is the first sonar hotspots (a category separate from issues) that we support. While we have a CLI arg separately for a file for hotspots, internally these can be processed pretty much the same way with just some minor tweaking to existing code

Closes #376

### DIFF
--- a/src/codemodder/codemods/sonar.py
+++ b/src/codemodder/codemods/sonar.py
@@ -54,12 +54,6 @@ class SonarCodemod(SASTCodemod):
 class SonarDetector(BaseDetector):
     _lazy_cache = None
 
-    def _process_sonar_findings(self, sonar_json_files: list[str]) -> SonarResultSet:
-        combined_result_set = SonarResultSet()
-        for file in sonar_json_files or []:
-            combined_result_set |= SonarResultSet.from_json(file)
-        return combined_result_set
-
     def apply(
         self,
         codemod_id: str,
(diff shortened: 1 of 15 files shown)


src/codemodder/codemods/sonar.py
```diff
@@ -54,20 +54,21 @@ def from_core_codemod(
...
     ) -> ResultSet:
         if not self._lazy_cache:
-            self._lazy_cache = self._process_sonar_findings(
+            self._lazy_cache = process_sonar_findings(
                 context.tool_result_files_map.get("sonar", [])
             )
         return self._lazy_cache
+
+
+def process_sonar_findings(sonar_json_files: list[str]) -> SonarResultSet:
```
drdavella, line 70: We could probably just use `@functools.cache` here and get rid of the `_lazy_cache` on the class?

drdavella (approved):
Nice, looks great!

src/codemodder/codemods/sonar.py
clavedeluna: I swear I recall there was a reason this was written as is, I think @andrecsilva wrote it originally? because we also posed using @functools.cache at that time

clavedeluna: going to merge so I can use this PR to add more sonar codemods, but can come back to this if needed

PR #438, opened 2024-04-03
Implement prototype codemods for DefectDojo remediation

drdavella:
## Overview
*Add new codemods to remediate (Semgrep) findings via DefectDojo*

## Description

* [DefectDojo](https://www.defectdojo.org/) is an open-source vulnerability management platform
  * It aggregates findings from multiple rule providers and exposes them via a unified API
* This PR adds the ability to process DefectDojo findings JSON files
* It also implements two codemods that remediate Semgrep rules represented as DefectDojo findings:
  * [django-secure-set-cookie](https://semgrep.dev/playground/r/python.django.security.audit.secure-cookies.django-secure-set-cookie?editorMode=advanced)
  * [avoid-insecure-deserialization](https://semgrep.dev/playground/r/python.django.security.audit.avoid-insecure-deserialization.avoid-insecure-deserialization?editorMode=advanced)
  * Eventually we can also adapt these codemods to remediate semgrep findings directly
* I did some refactoring to enable code reuse. Hopefully it's not too much of a mess
* Since this is primarily for demo purposes I skipped the integration tests for now. They can be added later

### DIFF
--- a/src/codemodder/codemodder.py
+++ b/src/codemodder/codemodder.py
@@ -156,22 +156,13 @@ def run(original_args) -> int:
     logger.info("codemodder: python/%s", __version__)
     logger.info("command: %s %s", Path(sys.argv[0]).name, " ".join(original_args))
 
-    # TODO: sonar files should be _parsed_ here as well
     # TODO: this should be dict[str, list[Path]]
-
     tool_result_files_map: DefaultDict[str, list[str]] = detect_sarif_tools(
         [Path(name) for name in argv.sarif or []]
     )
-    (
-        tool_result_files_map["sonar"].extend(argv.sonar_issues_json)
-        if argv.sonar_issues_json
-        else None
-    )
-    (
-        tool_result_files_map["sonar"].extend(argv.sonar_hotspots_json)
-        if argv.sonar_hotspots_json
-        else None
-    )
+    tool_result_files_map["sonar"].extend(argv.sonar_issues_json or [])
+    tool_result_files_map["sonar"].extend(argv.sonar_hotspots_json or [])
+    tool_result_files_map["defectdojo"] = argv.defectdojo_findings_json
 
     repo_manager = PythonRepoManager(Path(argv.directory))
     context = CodemodExecutionContext(
--- a/src/codemodder/codemodder.py
+++ b/src/codemodder/codemodder.py
@@ -162,7 +162,7 @@ def run(original_args) -> int:
     )
     tool_result_files_map["sonar"].extend(argv.sonar_issues_json or [])
     tool_result_files_map["sonar"].extend(argv.sonar_hotspots_json or [])
-    tool_result_files_map["defectdojo"] = argv.defectdojo_findings_json
+    tool_result_files_map["defectdojo"] = argv.defectdojo_findings_json or []
 
     repo_manager = PythonRepoManager(Path(argv.directory))
     context = CodemodExecutionContext(
--- a/src/core_codemods/defectdojo/api.py
+++ b/src/core_codemods/defectdojo/api.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from functools import cache
 from pathlib import Path
 from typing import cast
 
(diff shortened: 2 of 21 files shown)


src/codemodder/codemodder.py
```diff
@@ -156,22 +156,13 @@ def run(original_args) -> int:
...
-        else None
-    )
-    (
-        tool_result_files_map["sonar"].extend(argv.sonar_hotspots_json)
-        if argv.sonar_hotspots_json
-        else None
-    )
+    tool_result_files_map["sonar"].extend(argv.sonar_issues_json or [])
+    tool_result_files_map["sonar"].extend(argv.sonar_hotspots_json or [])
+    tool_result_files_map["defectdojo"] = argv.defectdojo_findings_json or []
```
clavedeluna, line 165: I changed this logic in [this PR](https://github.com/pixee/codemodder-python/pull/433/files) in favor of the if/else None because as you have it now, `tool_result_files_map` will have a `sonar` key even if we don't have any sonar findings, I don't think that's what we want? with if/else None, nothing is added if the argv. is empty

src/core_codemods/defectdojo/api.py
```diff
@@ -0,0 +1,78 @@
...
+
+class DefectDojoDetector(BaseDetector):
+    def apply(
+        self,
+        codemod_id: str,
+        context: CodemodExecutionContext,
+        files_to_analyze: list[Path],
+    ) -> ResultSet:
+        result_set = DefectDojoResultSet()
+        for filename in context.tool_result_files_map.get("defectdojo", []):
```
clavedeluna: I think you may also want a cache version here just like sonar has in `process_sonar_findings`

src/core_codemods/defectdojo/results.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Match location for DefectDojo results
+
+        Since DefectDojo does not provide column information, we can only match based on line number.
+        We check whether the start line of the result is within the range of the node.
+        """
+        del node
+        return any(
+            pos.start.line <= location.start.line <= pos.end.line
+            for location in self.locations
```
clavedeluna, line 43: could you reuse the existing `fuzzy_column_match` function here?

src/core_codemods/defectdojo/results.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+
+
+class DefectDojoResultSet(ResultSet):
+    @classmethod
+    @cache
```
clavedeluna, line 49: nice catch that this should be cached. Could you also add it to `SonarResultSet.from_json`?

src/core_codemods/defectdojo/semgrep/avoid_insecure_deserialization.py
```diff
@@ -0,0 +1,50 @@
...
+    LibcstResultTransformer,
+    NameResolutionMixin,
+    HardenPyyamlCallMixin,
+    CodemodProtocol,
+):
+    def leave_Call(
+        self, original_node: cst.Call, updated_node: cst.Call
+    ) -> cst.Call | None:
+        if self.node_is_selected(original_node):
+            if self.find_base_name(original_node) == "yaml.load":
```
clavedeluna: seems worth combining the two if statements into one

clavedeluna (changes requested):

src/codemodder/codemodder.py
drdavella: Each of the tool detectors does something like this, so in either case the default gets treated as an empty array.
```python
context.tool_result_files_map.get("sonar", [])
```

Also using `None` defeats the purpose of having a `defaultdict` here. I think it's cleaner just to set it to an empty array.

src/core_codemods/defectdojo/results.py
drdavella: I'm looking at that function and it just applies to columns, but now I'm questioning whether the logic there is correct. In either case, DefectDojo doesn't give us column information so that function can't help us.

src/codemodder/codemodder.py
clavedeluna: The if/else None isn't doing this:
`defaultdict(<class 'list'>, {'sonar': None})`
it's doing this:
`defaultdict(<class 'list'>, {})` # totally empty

which is what I think we want if sonar json isn't used. There is no reason to have the "sonar" key at all in that dict if it isn't used.

src/core_codemods/defectdojo/results.py
clavedeluna: @andrecsilva recently commented something similar that leads to the general question of: why do we use column info in the first place? Perhaps there are strong reasons in some cases with lines that have multiple nodes. we should explore this later on

clavedeluna (approved):

src/codemodder/codemodder.py
drdavella: I see. I don't think it's a problem to have the sonar key since in either case the detector fills the result with an empty list.

src/core_codemods/defectdojo/results.py
drdavella: We can use column information to differentiate ambiguous cases. For example, imagine we get a finding that says there's a SQL injection vulnerability on this line:
```python
result = cursor.execute(query).something_else(whatever)
```

Without column information, we don't know which of those call sites potentially need to get modified and if we're not careful we would end up modifying both of them.

PR #450, opened 2024-04-06
New find-and-fix and sonar codemod for floating point equality

### LINKED ISSUE #374: Sonar codemod: fix "Floating point numbers should not be tested for equality"
https://rules.sonarsource.com/python/type/Bug/RSPEC-1244/

clavedeluna:
## Overview
*New codemod(s) to replace floating point equality comparison with `math.isclose`*

## Description

* This codemod will apply if at least one of the variables of the operation is a floating point number.
* In cases when we cannot determine what a constant would be (function args), and none o the related variables are floats, this codemod will not apply
* would've been nice to implement the find part of this with semgrep but it doesn't have typed metavariable matching for python (makes sense)

Closes #374

### DIFF
--- a/src/core_codemods/docs/pixee_python_fix-float-equality.md
+++ b/src/core_codemods/docs/pixee_python_fix-float-equality.md
@@ -1,6 +1,6 @@
 In most programming languages, floating point arithmetic is imprecise due to the way floating point numbers are stored as binary representations. Moreover, the result of calculations with floats can vary based on when rounding happens. Using equality or inequality to compare floats or their operations will almost always be imprecise and lead to bugs.
 
-For these reasons, this codemod changes any operations involving equality or inequality with floats to the recommended `math.isclose` function. We have explicitly defined the default parameters `rel_tol=1e-09` and  `abs_tol=0.0` as a starting point for you to consider depending on your calculation needs.
+For these reasons, this codemod changes any operations involving equality or inequality with floats to the recommended `math.isclose` function. This codemod uses the default parameter values `rel_tol=1e-09` and  `abs_tol=0.0` but makes them explicit as a starting point for you to consider depending on your calculation needs.
 
 Our changes look like the following:
 ```diff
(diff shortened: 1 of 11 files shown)


src/core_codemods/fix_float_equality.py
```diff
@@ -0,0 +1,98 @@
...
+                return self.at_least_one_float(left_type.left, left_type.right)
+            case (_, cst.BinaryOperation()):
+                return self.at_least_one_float(right_type.left, right_type.right)
+        return False
+
+
+FixFloatEquality = CoreCodemod(
+    metadata=Metadata(
+        name="fix-float-equality",
+        summary="Replace `==` or `!=` with `math.isclose` Call for Floats or Arithmetic Operations with Floats",
```
drdavella:
```suggestion
        summary="Use `math.isclose` Instead of Direct Equality for Floats",
```

src/core_codemods/docs/pixee_python_fix-float-equality.md
```diff
@@ -0,0 +1,12 @@
+In most programming languages, floating point arithmetic is imprecise due to the way floating point numbers are stored as binary representations. Moreover, the result of calculations with floats can vary based on when rounding happens. Using equality or inequality to compare floats or their operations will almost always be imprecise and lead to bugs.
+
+For these reasons, this codemod changes any operations involving equality or inequality with floats to the recommended `math.isclose` function. We have explicitly defined the default parameters `rel_tol=1e-09` and  `abs_tol=0.0` as a starting point for you to consider depending on your calculation needs.
```
drdavella:
```suggestion
For these reasons, this codemod changes any operations involving equality or inequality with floats to the recommended `math.isclose` function. This codemod uses the default parameter values `rel_tol=1e-09` and  `abs_tol=0.0` but makes them explicit as a starting point for you to consider depending on your calculation needs.
```

drdavella (changes requested):
Looks great overall, thanks for letting me nitpick the metadata.

src/core_codemods/fix_float_equality.py
```diff
@@ -0,0 +1,98 @@
...
+)
+from codemodder.codemods.utils_mixin import NameAndAncestorResolutionMixin
+from core_codemods.api import Metadata, Reference, ReviewGuidance
+from core_codemods.api.core_codemod import CoreCodemod
+
+
+class FixFloatEqualityTransformer(
+    LibcstResultTransformer, NameAndAncestorResolutionMixin
+):
+    change_description = "Replace `==` or `!=` with `math.isclose` callz"
```
drdavella: Sorry I missed this one on the original review.
```suggestion
    change_description = "Replace `==` or `!=` with `math.isclose`"
```

drdavella (changes requested):

drdavella (approved):

Issue #388, opened 2024-03-18
Only core pixee codemods can be requested by name


clavedeluna:
Inception of this idea is [here](https://github.com/pixee/codemodder-python/pull/384#discussion_r1528571022)

Basically as we add more and more codemods, and users add their own, the chance that two codemods have the same name increases, even with different IDs. We currently allow `--codemod-include` and `--codemod-exclude` (and likely other CLI features?) to allow either codemod name or codemod id. However, it would be smart to change this to only allow core pixee codemods to be asked by name. Everything  by ID

PR #458, opened 2024-04-10
Support simple patterns for codemod include/exclude

### LINKED ISSUE #316: Codemodder include and exclude flags should understand patterns
We need `--codemod-include` and `--codemod-exclude` to be able to include/exclude all codemods from a particular origin. For example, it would be useful to give an option like this:
```
--codemod-include=sonar:*
```
which could ensure all Sonar codemods are included.

Conversely, it would be useful to give an option like this:
```
---codemod-exclude=pixee:*
```
which would exclude all core Pixee codemods.

I think the pattern matching only needs to be supported at the level of origin. We don't need to support globs/patterns for specific names like `pixee:python/xml-*`.

drdavella:
## Overview
*Support simple patterns for codemod include/exclude CLI flags*

## Description

* The spec was updated to enable simple pattern matching for codemod include/exclude
* In order to support this I simplified the exclusion logic a bit
* I took the liberty of removing the warning for non-existing codemods given to exclude; I'm not sure it's necessary
* I'm still a little unsure about the need for `sast_only` but I left it as-is for this PR

Closes #316

### DIFF
--- a/src/codemodder/registry.py
+++ b/src/codemodder/registry.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 from dataclasses import dataclass
 from importlib.metadata import entry_points
 from typing import TYPE_CHECKING, Optional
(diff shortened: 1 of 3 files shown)


src/codemodder/registry.py
```diff
@@ -64,36 +65,41 @@ def match_codemods(
...
                     continue
 
-                # remove both by name and id since we don't know which `name_or_id` represented
-                base_codemods.pop(codemod.name, None)
-                base_codemods.pop(codemod.id, None)
+                if bool(sast_only) != bool(codemod.origin == "pixee"):
+                    base_codemods[codemod.id] = codemod
+
             # Remove duplicates and preserve order
             return list(dict.fromkeys(base_codemods.values()))
```
clavedeluna: The logic is so subtly changed so I'm going off passing unit tests but I _think_ you can remove this list(dict.from_keys...) since that was necessary when base_codemods did this:
```
                    base_codemods[codemod.id] = codemod
                    base_codemods[codemod.name] = codemod
```
but since that was only necessary to log an excluded codemod and that's been removed, you can remove this duplication removal.

src/codemodder/registry.py
```diff
@@ -64,36 +65,41 @@ def match_codemods(
...
+                for exclude in codemod_exclude
+                if "*" in exclude
+            ]
+            names = set(name for name in codemod_exclude if "*" not in name)
             for codemod in self.codemods:
-                if (sast_only and codemod.origin != "pixee") or (
-                    not sast_only and codemod.origin == "pixee"
+                if (
+                    codemod.id in names
+                    or (codemod.origin == "pixee" and codemod.name in names)
```
clavedeluna, line 77: does this line accomplish [this issue](https://github.com/pixee/codemodder-python/issues/388) for excluded?

clavedeluna (approved):
FYI the request to log bad excluded was yours and you see the lengths I had to go to get it there so it's kinda nice that this code is simpler without it. I do actually think the logging was useful but up to you.

Approving but I think small change is needed.

src/codemodder/registry.py
drdavella: Yes I think it does for the exclude case but I think we'd want to add more tests so we can keep it as a separate issue.

drdavella:
> FYI the request to log bad excluded was yours and you see the lengths I had to go to get it there so it's kinda nice that this code is simpler without it. I do actually think the logging was useful but up to you.

Yes sorry I see that this was complicated and I'm just not sure it's necessary so I was able to simplify it.

PR #461, opened 2024-04-11
Change default behavior for Semgrep and ResultSet

drdavella:
## Overview
*Change default path and rule ID normalization for Semgrep and `ResultSet`*

## Description

* Essentially this PR inverts the default behavior of path normalization for SAST results
* Now all SAST detectors compute paths relative to the given base directory _except_ for our internal Semgrep runs
* It turns out that the way we do it for internal Semgrep runs is a little weird, and most/all other SAST tools need to handle it the other way
* The same is true for the way that we handle Semgrep rule IDs: our internal usage is a special case

### DIFF
--- a/src/codemodder/semgrep.py
+++ b/src/codemodder/semgrep.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Iterable, Optional
 
-from typing_extensions import Self
+from typing_extensions import Self, override
 
 from codemodder.context import CodemodExecutionContext
 from codemodder.logging import logger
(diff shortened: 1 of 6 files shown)


src/codemodder/semgrep.py
```diff
@@ -56,8 +42,27 @@ def from_sarif(cls, sarif_location) -> Self:
...
+            ]
+
+        return None
+
+    @classmethod
+    def from_sarif(
+        cls, sarif_result, sarif_run, truncate_rule_id: bool = False
+    ) -> Self:
+        rule_id = cls.extract_rule_id(sarif_result, sarif_run, truncate_rule_id)
         if not rule_id:
```
clavedeluna, line 66:
```suggestion
        if not (rule_id := cls.extract_rule_id(sarif_result, sarif_run, truncate_rule_id)):
```

src/codemodder/semgrep.py
```diff
@@ -98,7 +116,6 @@ def run(
         command = [
             "semgrep",
             "scan",
-            "--legacy",
```
clavedeluna, line 101: <3

clavedeluna (approved):
tiniest of change

PR #467, opened 2024-04-12
Add LLM client to context and optional dependency on OpenAI

drdavella:
## Overview
*Centrally manage LLM client for codemods that may require it*

### DIFF
--- a/src/codemodder/context.py
+++ b/src/codemodder/context.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import itertools
 import logging
+import os
 from pathlib import Path
 from textwrap import indent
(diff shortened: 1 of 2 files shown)


andrecsilva (approved):

src/codemodder/context.py
```diff
@@ -20,7 +21,15 @@
...
 from codemodder.utils.timer import Timer
 
+try:
+    from openai import Client
+except ImportError:
+    Client = None
+
+
 if TYPE_CHECKING:
+    from openai import Client
```
clavedeluna, line 31: would this not suffer from `importError` like above?

PR #475, opened 2024-04-16
Codemod to fix `math.isclose` comparison to `0`

### LINKED ISSUE #457: New find+fix and sonar codemod for `math.isclose` without `abs_tol` kwarg if comparing to `0`
https://rules.sonarsource.com/python/RSPEC-6727/

very similar to `fix-float-equality` codemod

clavedeluna:
## Overview
*`math.isclose` with a `0` arg should have a  `abs_tol` kwarg defined*

To accomplish this, I added some useful functionality that checks if a cst.Node is or evaluates to `0`, int or float, or some other edge cases. I'm sure we will build on this later on as well.

Closes #457

### DIFF
--- a/src/codemodder/codemods/utils.py
+++ b/src/codemodder/codemods/utils.py
@@ -201,6 +201,18 @@ def is_zero(node: cst.CSTNode) -> bool:
                 return float(node.value) == 0
             except (ValueError, TypeError):
                 return False
-        # todo: add float(), int() etc
-        case _:
-            return False
+        case cst.Call(func=cst.Name("int"), args=[]) | cst.Call(
+            func=cst.Name("float"), args=[]
+        ):
+            # int() or float() == 0
+            return True
+        case cst.Call(
+            func=cst.Name("int"), args=[cst.Arg(value=cst.Name(value="False"))]
+        ) | cst.Call(
+            func=cst.Name("float"), args=[cst.Arg(value=cst.Name(value="False"))]
+        ):
+            # int(False) or float(False)
+            return True
+        case cst.Call(func=cst.Name("int")) | cst.Call(func=cst.Name("float")):
+            return is_zero(node.args[0].value)
+    return False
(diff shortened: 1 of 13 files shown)


src/codemodder/codemods/utils.py
```diff
@@ -192,3 +192,27 @@ def is_assigned_to_True(original_node: cst.Assign):
         isinstance(original_node.value, cst.Name)
         and original_node.value.value == "True"
     )
+
+
+def is_zero(node: cst.CSTNode) -> bool:
```
drdavella, line 197: This is pretty cool! I do wonder whether `ast.literal_eval` would work just as well for these purposes, but I'm okay with this.

src/core_codemods/docs/pixee_python_fix-math-isclose.md
```diff
@@ -0,0 +1,12 @@
+The default value for the `abs_tol` argument to a `math.isclose` call is `0`. Using this default when comparing a value against `0`, such as in `math.isclose(a, 0)` is equivalent to a strict equality check to `0`, which is not the intended use of the `math.isclose` function.
+
+This codemod adds `abs_tol=1e-09` to any call to `math.isclose` with one of of the first arguments evaluating to `0` if `abs_tol` is not already specified. `1e-09` is a starting point for you to consider depending on your calculation needs.
```
drdavella, line 3:
```suggestion
This codemod adds `abs_tol=1e-09` to any call to `math.isclose` if one of of the first arguments evaluates to `0` and `abs_tol` is not already specified. `1e-09` is a suggestion which you may need to adjust depending on your calculation needs.
```

drdavella (approved):

src/codemodder/codemods/utils.py
clavedeluna: I tested it out but it only works well for the `case cst.Integer() | cst.Float():`, everything else still requires special casing manually so yeah gonna leave it as is for now. Good call though!

PR #477, opened 2024-04-16
use-walrus-if codemod can handle unused variables

### LINKED ISSUE #264: Handle unused assignments in `use-walrus-if` codemod
## Details
In some cases the `use-walrus-if` codemod makes a change such that the involved variable is no longer used at all. In other words, the original conditional is the only place where the result of the assignment was used.

We can use the scope provider and parent metadata in libCST to determine which such cases occur. In these cases we should not rewrite with the walrus operator at all but simply use the RHS of the original assign expression as the conditional value without any assignment at all.

We can actually see an example where the current behavior causes our own pipelines to fail in #258.

clavedeluna:
## Overview
*`use-walrus-if` codemod should not use a walrus if variable is not used*

Closes #264

### DIFF
--- a/src/core_codemods/use_walrus_if.py
+++ b/src/core_codemods/use_walrus_if.py
@@ -77,7 +77,7 @@ class UseWalrusIf(SimpleCodemod, NameResolutionMixin):
                 return test
         return None
 
-    def _single_access(self, original_node: cst.CSTNode) -> bool:
+    def _single_access(self, original_node: cst.IfExp) -> bool:
         match original_node.test:
             case cst.Name():
                 access = self.find_accesses(original_node.test)
@@ -98,7 +98,6 @@ class UseWalrusIf(SimpleCodemod, NameResolutionMixin):
                 continue
 
             assign, target, value = found_assign
-            # breakpoint()
             match if_test:
                 # If test can be a comparison expression
                 case cst.Comparison(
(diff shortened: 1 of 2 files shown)


src/core_codemods/use_walrus_if.py
```diff
@@ -87,7 +98,7 @@ def on_visit(self, node: cst.CSTNode) -> Optional[bool]:
                 continue
 
             assign, target, value = found_assign
-
+            # breakpoint()
```
andrecsilva: Debugging leftover, I'm guessing.

src/core_codemods/use_walrus_if.py
```diff
@@ -130,22 +141,31 @@ def leave_If(self, original_node, updated_node):
...
                     return updated_node.with_changes(
-                        test=updated_node.test.with_changes(left=named_expr)
+                        test=updated_node.test.with_changes(left=new_expression)
                     )
 
         return original_node
 
     def leave_Assign(self, original_node: cst.Assign, updated_node: cst.Assign):
         del updated_node
+        # breakpoint()
```
andrecsilva: Same as previous one.

src/core_codemods/use_walrus_if.py
```diff
@@ -76,6 +77,16 @@ def _filter_if(self, node: cst.CSTNode) -> cst.BaseExpression | None:
                 return test
         return None
 
+    def _single_access(self, original_node: cst.CSTNode) -> bool:
```
andrecsilva: Change this type hint of `original_node` to `IfExp`.

andrecsilva (approved):
Some small nitpicks, looks ok otherwise.

PR #479, opened 2024-04-17
fix-dataclass-defaults can handle populated defaults

### LINKED ISSUE #354: `fix-dataclass-defaults` codemod should fix populated mutable defaults
This codemod should handle this case:
```
from dataclasses import dataclass

@dataclass
class Test:
    name: str = ""
    phones: list = [1, 2, 3]
    friends: dict = {"one": 1}
    family: set = set((1, 2))
```
I believe we can use `lambda` with `field` somehow

clavedeluna:
Closes #354

### DIFF
--- a/src/codemodder/codemods/base_visitor.py
+++ b/src/codemodder/codemods/base_visitor.py
@@ -66,6 +66,15 @@ class UtilsMixin(MetadataDependent):
     def lineno_for_node(self, node):
         return self.node_position(node).start.line
 
+    def code(self, node: cst.CSTNode) -> str:
+        """
+        Only a cst.Module node has a `code` attribute which converts the node
+        back to the original code as a str. To get the code for any node,
+        the suggested approach is to wrap this node in a `cst.Module` node.
+        """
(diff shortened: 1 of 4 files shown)


src/codemodder/codemods/base_visitor.py
```diff
@@ -66,6 +66,15 @@ def node_position(self, node):
     def lineno_for_node(self, node):
         return self.node_position(node).start.line
 
+    def code(self, node: cst.CSTNode) -> str:
```
drdavella, line 69: Just curious but I don't see where this is used in the changes below. Is this just to enable testing?

drdavella (approved):

src/codemodder/codemods/base_visitor.py
clavedeluna: it is used :) it's kinda buried in there isn't it? https://github.com/pixee/codemodder-python/pull/479/files#diff-f5b72205f33eb7f46fdbe920b68b37edc816051d59deec75147453b533e567e8R74

PR #482, opened 2024-04-18
add language metadata field

clavedeluna:
## Overview
*Adding a language metadata field with a `python` default so we can create codemods dedicated for other languages*


drdavella (approved):
Not sure whether this belongs to `Metadata` or whether it's solely a property of the codemod itself but we can revisit that later.

PR #483, opened 2024-04-18
Short circuit faster when no tool results are available

drdavella:
## Overview
*We need to short-circuit much faster when no tool results are available for a particular codemod*


dunningdan (approved):

src/codemodder/codemods/base_codemod.py
```diff
@@ -162,6 +162,10 @@ def _apply(
             else None
         )
 
+        if results is not None and not results:
```
clavedeluna, line 165: maybe I'm missing something super subtle but can't you just do `if not results`? since results could be None or an empty dict?

PR #488, opened 2024-04-19
add detector name and update test utils

clavedeluna:
## Overview
*Update testing for extensions to test other languages and base detectors need a name to know which type of detector it is*


drdavella (approved):
Approving to allow progress but we might need to consolidate this with `ToolMetadata` at some point.

clavedeluna:
removing name thing, just updating tests

Issue #382, opened 2024-03-15
Setting `shell=False` requires the command to be an list of strings


drdavella:
Right now our codemod does not account for the fact that in most cases for `shell=False` to work the command needs to be a list of strings rather than a single string.

For example, the following is valid:
```python
subprocess.run("echo foo", shell=True)
```

Whereas the following is not:
```python
subprocess.run("echo foo", shell=False)
FileNotFoundError: [Errno 2] No such file or directory: 'echo foo'
```

Instead, it needs to be the following:
```python
subprocess.run(["echo", "foo"], shell=False)
```

We have a few options here but the most correct is probably to do the following:
1. Detect whether the first argument to the subprocess call is a string or array
2. In cases where this is not possible, do not make a modification
3. If the argument is an array, make the change
4. If the argument is a string, split the string on empty spaces and convert to an array

clavedeluna:
@drdavella  just to be clear since you did not specify the codemod, it's `subprocess-shell-false` you're talking about, not `sandbox-process-creation`

clavedeluna:
This is a good refinement, so I'm bringing this test case that I thought about to see how we want to handle it.

Say we want to run `ls -l | grep '.txt'` with subprocess. This has a pipe which is a shell feature. The way you'd write the code for this would be
`subprocess.run("ls -l | grep '.txt'", shell=True)` which would correctly `ls` any .txt files. If we follow the suggestions in the issue, we would end up converting this to ` subprocess.run(["ls", "-l", "|", "grep", "'.txt'"], shell=False)` which breaks entirely.

There are more complicated ways to handle shell features like a pipe, we could use Popen and subprocess.PIPE, but that deviates from the purpose of this codemod I feel like.

So what do we want to do? Do we want to check if strings like "|" or other shell feature strings are present and not change it?

drdavella:
@clavedeluna that's a really good question. There are definitely valid cases for using `shell=True`. Maybe the safest heuristic is to simply ignore entirely any cases that have string literals instead of arrays.

clavedeluna:
Would be a similar logic to #263 
So perhaps the most refined heuristic would be to check if:
1. if argument is str literal, `subprocess.run("echo foo", shell=True)`, don't do anything
1a. similarly, if arg is a var pointing to str literal
```
cmd = "echo foo"
subprocess.run(cmd, shell=True)
```
don't do anything

2. if argument is a list, check if all list elements resolve to a str literal, if so, don't do anything
[...]

PR #489, opened 2024-04-19
Fix FileNotFoundError caused by SubprocessShellFalse, Improve CombineStartswithEndswith, Add CombineIsinstanceIssubclass

LucasFaudman:
### `SubprocessShellFalse` 
- Problem: Current `SubprocessShellFalse` codemod will cause a `FileNotFoundError` when the first arg is a string.
- Example: Consider `subprocess.run('ls -l', shell=True)`. The codemod would break this code since with `shell=False` subprocess will try to find a non-existant file name 'ls -l'.
- Solution: Only run codeMod if the first arg is not a `m.SimpleString() | m.ConcatenatedString() | m.FormattedString()`. (This does not solve the case where the first arg is a cst.Name() )

### `CombineStartswithEndswith`
- Problem: Current `CombineStartswithEndswith` only supports `m.SimpleString()` and does not support chaining more than 2 `startswith` or `endswith` calls.
- Solution: 
    - Allow arg to be of type `m.Tuple()| m.SimpleString() | m.ConcatenatedString() | m.FormattedString() | m.Name()`
    - Join all the `startswith` and `endswith` calls into a single call and remove duplicate args with the same `evaluated_value`
```python
# Input
s.startswith('a') or s.startswith('b') or s.startswith('c')
s.startswith(f'{a}') or s.startswith(('con' 'cat', someVar)) or s.startswith('simple')
# CodeMod result before changes
s.startswith(('a', 'b')) or s.startswith('c')
s.startswith(f'{a}') or s.startswith(('con' 'cat', someVar)) or s.startswith('simple')    
# After 
s.startswith(('a', 'b', 'c'))
s.startswith((f'{a}', 'con' 'cat', someVar, 'simple'))
```

### `CombineIsinstanceIssubclass`
- New CodeMod very similar to `CombineStartswithEndswith` but for `isinstance` and `issubclass` calls.
```python
# Input
isinstance(a, int) or isinstance(a, float) or isinstance(a, str)
issubclass(a, str) or issubclass(a, (bytes, bytearray)) or issubclass(a, memoryview)

# Output
isinstance(a, (int, float, str))
issubclass(a, (str, bytes, bytearray, memoryview))
```

### DIFF
--- a/src/codemodder/codemods/utils.py
+++ b/src/codemodder/codemods/utils.py
@@ -1,6 +1,6 @@
 from enum import Enum
 from pathlib import Path
-from typing import Any, Optional, TypeAlias
+from typing import Any, Iterator, Optional, Type, TypeAlias, TypeVar
 
 import libcst as cst
 from libcst import MetadataDependent, matchers
@@ -216,3 +216,23 @@ def is_zero(node: cst.CSTNode) -> bool:
         case cst.Call(func=cst.Name("int")) | cst.Call(func=cst.Name("float")):
             return is_zero(node.args[0].value)
     return False
+
+
+_CSTNode = TypeVar("_CSTNode", bound=cst.CSTNode)
+
+
+def extract_boolean_operands(
+    node: cst.BooleanOperation, ensure_type: Type[_CSTNode] = cst.CSTNode
+) -> Iterator[_CSTNode]:
+    """
+    Recursively extract operands from a cst.BooleanOperation node from left to right as an iterator of nodes.
+    """
+    if isinstance(node.left, cst.BooleanOperation):
+        # Recursively yield operands from the boolean operation on the left
+        yield from extract_boolean_operands(node.left, ensure_type)
+    else:
+        # Yield left operand
+        yield cst.ensure_type(node.left, ensure_type)
+
+    # Yield right operand
+    yield cst.ensure_type(node.right, ensure_type)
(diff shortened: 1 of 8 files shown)


drdavella:
Hi @LucasFaudman! Nice to see your contribution here 😄 

We will take a look at this PR. At a glance it seems like it potentially solves #382.

Also, in general it's a bit easier if individual fixes are separated into individual PRs. This makes it easier to review and generally speeds things up since comments that only affect one fix don't have to block other fixes.

If there's any chance you can break this into three separate PRs, that would be appreciated. If not, we'll do our best as-is.

LucasFaudman:
> Hi @LucasFaudman! Nice to see your contribution here 😄
> 
> We will take a look at this PR. At a glance it seems like it potentially solves #382.
> 
> Also, in general it's a bit easier if individual fixes are separated into individual PRs. This makes it easier to review and generally speeds things up since comments that only affect one fix don't have to block other fixes.
> 
> If there's any chance you can break this into three separate PRs, that would be appreciated. If not, we'll do our best as-is.

Sure, I know you mentioned that before, sorry about that. I have my local git workflows setup to BFG my branch history on merging into public repos to prevent to making security mistakes which is why I've made the large PRs in the past. 

This afternoon i'll re-branch this into 3 PRs and leave a comment on #382 clarifying the issue.

PR #490, opened 2024-04-19
Update codetf bindings

drdavella:
## Overview
*Update CodeTF bindings to conform to new spec*

## Description

* The CodeTF spec was updated in https://github.com/pixee/codemodder-specs/pull/25
* This PR makes our bindings trivially conform with the new spec
* Actually _populating_ those new fields will come in a future PR


clavedeluna (approved):
Well if things aren't getting more complicated!

Approving but you'll need to update integration tests.

PR #492, opened 2024-04-20
prevent SubprocessShellFalse from causing FileNotFoundError when 1st arg is string

LucasFaudman:
### `SubprocessShellFalse` 
- Problem: Current `SubprocessShellFalse` codemod will cause a `FileNotFoundError` when the first arg is a string.
- Example: Consider `subprocess.run('ls -l', shell=True)`. The codemod would break this code since with `shell=False` subprocess will try to find a non-existant file name 'ls -l'.
- Solution: Only run codeMod if the first arg is not a `m.SimpleString() | m.ConcatenatedString() | m.FormattedString()`. 

Note: This only catches m.SimpleString() | m.ConcatenatedString() | m.FormattedString(). If the first arg is a cst.Name() or cst.Call() which returns a string and the type cannot be derived then this issue could still occur.

### DIFF
--- a/src/core_codemods/subprocess_shell_false.py
+++ b/src/core_codemods/subprocess_shell_false.py
@@ -1,5 +1,5 @@
 import libcst as cst
-from libcst import matchers
+from libcst import matchers as m
 from libcst.metadata import ParentNodeProvider
 
 from codemodder.codemods.check_annotations import is_disabled_by_annotations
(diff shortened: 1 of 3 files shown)


src/core_codemods/subprocess_shell_false.py
```diff
@@ -35,17 +35,22 @@ class SubprocessShellFalse(SimpleCodemod, NameResolutionMixin):
...
             return updated_node
 
-        if self.find_base_name(original_node.func) in self.SUBPROCESS_FUNCS:
+        if (
+            self.find_base_name(original_node.func) in self.SUBPROCESS_FUNCS
+        ) and not m.matches(
+            original_node.args[0],
+            m.Arg(
+                value=m.SimpleString() | m.ConcatenatedString() | m.FormattedString()
+            ),  # First argument to subprocess.<func> cannot be a string or setting shell=False will cause a FileNotFoundError
```
clavedeluna: instead of a comment could you make this match check a separate func, so we can do `if self.find_base ... and not self._is_first_arg.....`

clavedeluna (changes requested):

src/core_codemods/subprocess_shell_false.py
LucasFaudman: Sure, done.

drdavella (commented):
@LucasFaudman this is very nice work, thank you!

I just want to mention two things which I think are complimentary with the work you've done here:
* We could also update the Semgrep pattern in order to add dataflow tracking which would let us detect cases where the string isn't defined inline at the call
* For cases where we can determine conclusively that the first argument _is_ a string, we could update the call to use an array literal (although that's only going to work for the very simplest cases)

I don't think either of those things needs to be part of this particular PR, but it's worth considering going forward.

clavedeluna:
@LucasFaudman  let us know when you're ready for re-review as we're excited to get these fixes in!

LucasFaudman:
> @LucasFaudman let us know when you're ready for re-review as we're excited to get these fixes in!

@clavedeluna This PR and #493 are ready for re-review. Writing the additional docs for #494 now. Thanks!

clavedeluna (approved):

clavedeluna:
@LucasFaudman you'll need to revisit the integration test failure here

LucasFaudman:
> @LucasFaudman you'll need to revisit the integration test failure here

Will review today

LucasFaudman:
@clavedeluna changed arg from string to list, integration tests now pass

clavedeluna:
@LucasFaudman if there was a corresponding issue for this PR, could you close it and link it to this PR?

PR #493, opened 2024-04-20
support >2 or's in CombineStartswithEndswith, args joined to tuple no duplicates

### LINKED ISSUE #309: Handle more cases for `combine-startswith-endswith`
It was recently reported that the `combine-startswith-endswith` codemod would be more effective if it handled some additional cases, including:
* Cases where more than two calls are involved:
```python
if x.startswith("foo") or x.startswith("bar") or x.startswith("baz"): ...
```
* Cases where one call already uses a tuple:
if x.startswith(("foo", "bar")) or x.startswith("baz"): ...
```
* Cases where more than one call uses a tuple:
if x.startswith(("foo", "bar")) or x.startswith(("baz", "biz")): ...
```

We should add test cases and handle these cases where possible.

LucasFaudman:
### `CombineStartswithEndswith`
- Problem: Current `CombineStartswithEndswith` only supports `m.SimpleString()` and does not support chaining more than 2 `startswith` or `endswith` calls.
- Solution: 
    - Allow arg to be of type `m.Tuple()| m.SimpleString() | m.ConcatenatedString() | m.FormattedString() | m.Name()`
    - Join all the `startswith` and `endswith` calls into a single call and remove duplicate args with the same `evaluated_value`
```python
# Input
s.startswith('a') or s.startswith('b') or s.startswith('c')
s.startswith(f'{a}') or s.startswith(('con' 'cat', someVar)) or s.startswith('simple')
# CodeMod result before changes
s.startswith(('a', 'b')) or s.startswith('c')
s.startswith(f'{a}') or s.startswith(('con' 'cat', someVar)) or s.startswith('simple')    
# After 
s.startswith(('a', 'b', 'c'))
s.startswith((f'{a}', 'con' 'cat', someVar, 'simple'))
```

Closes #309

### DIFF
--- a/src/codemodder/codemods/utils.py
+++ b/src/codemodder/codemods/utils.py
@@ -228,11 +228,8 @@ def extract_boolean_operands(
     Recursively extract operands from a cst.BooleanOperation node from left to right as an iterator of nodes.
     """
     if isinstance(node.left, cst.BooleanOperation):
-        # Recursively yield operands from the boolean operation on the left
         yield from extract_boolean_operands(node.left, ensure_type)
     else:
-        # Yield left operand
         yield cst.ensure_type(node.left, ensure_type)
 
-    # Yield right operand
     yield cst.ensure_type(node.right, ensure_type)
--- a/src/codemodder/codemods/utils.py
+++ b/src/codemodder/codemods/utils.py
@@ -1,6 +1,6 @@
 from enum import Enum
 from pathlib import Path
-from typing import Any, Optional, TypeAlias
+from typing import Any, Iterator, Optional, Type, TypeAlias, TypeVar
 
 import libcst as cst
 from libcst import MetadataDependent, matchers
@@ -216,3 +216,23 @@ def is_zero(node: cst.CSTNode) -> bool:
         case cst.Call(func=cst.Name("int")) | cst.Call(func=cst.Name("float")):
             return is_zero(node.args[0].value)
     return False
+
+
+_CSTNode = TypeVar("_CSTNode", bound=cst.CSTNode)
+
+
+def extract_boolean_operands(
+    node: cst.BooleanOperation, ensure_type: Type[_CSTNode] = cst.CSTNode
+) -> Iterator[_CSTNode]:
+    """
+    Recursively extract operands from a cst.BooleanOperation node from left to right as an iterator of nodes.
+    """
+    if isinstance(node.left, cst.BooleanOperation):
+        # Recursively yield operands from the boolean operation on the left
+        yield from extract_boolean_operands(node.left, ensure_type)
+    else:
+        # Yield left operand
+        yield cst.ensure_type(node.left, ensure_type)
+
+    # Yield right operand
+    yield cst.ensure_type(node.right, ensure_type)
--- a/src/core_codemods/combine_startswith_endswith.py
+++ b/src/core_codemods/combine_startswith_endswith.py
@@ -18,37 +18,14 @@ class CombineStartswithEndswith(SimpleCodemod, NameResolutionMixin):
     def leave_BooleanOperation(
         self, original_node: cst.BooleanOperation, updated_node: cst.BooleanOperation
     ) -> cst.CSTNode:
+        if not self.filter_by_path_includes_or_excludes(
+            self.node_position(original_node)
+        ):
+            return updated_node
+
         if self.matches_startswith_endswith_or_pattern(original_node):
-            if not self.filter_by_path_includes_or_excludes(
-                self.node_position(original_node)
-            ):
-                return updated_node
-
             self.report_change(original_node)
-
-            elements = []
-            seen_evaluated_values = set()
-            for call in extract_boolean_operands(updated_node, ensure_type=cst.Call):
-                arg_value = call.args[0].value
-                if isinstance(arg_value, cst.Tuple):
-                    arg_elements = arg_value.elements
-                else:
-                    arg_elements = (cst.Element(value=arg_value),)
-
-                for element in arg_elements:
-                    if (
-                        evaluated_value := getattr(
-                            element.value, "evaluated_value", None
-                        )
-                    ) in seen_evaluated_values:
-                        # If an element has a non-None evaluated value that has already been seen, continue to avoid duplicates
-                        continue
-                    if evaluated_value is not None:
-                        seen_evaluated_values.add(evaluated_value)
-                    elements.append(element)
-
-            new_arg = cst.Arg(value=cst.Tuple(elements=elements))
-            return cst.Call(func=call.func, args=[new_arg])
+            return self.make_new_call_from_boolean_operation(updated_node)
 
         return updated_node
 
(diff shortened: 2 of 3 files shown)


clavedeluna:
I think you can add `Closes #..issue` to your PR description for the issue you're fixing here :) Great work!

src/codemodder/codemods/utils.py
```diff
@@ -216,3 +216,23 @@ def is_zero(node: cst.CSTNode) -> bool:
...
+
+
+def extract_boolean_operands(
+    node: cst.BooleanOperation, ensure_type: Type[_CSTNode] = cst.CSTNode
+) -> Iterator[_CSTNode]:
+    """
+    Recursively extract operands from a cst.BooleanOperation node from left to right as an iterator of nodes.
+    """
+    if isinstance(node.left, cst.BooleanOperation):
+        # Recursively yield operands from the boolean operation on the left
```
clavedeluna: Your code is nice and clean and easy to interpret so no need for these comments here or a bit below!

src/core_codemods/combine_startswith_endswith.py
```diff
@@ -17,36 +18,54 @@ class CombineStartswithEndswith(SimpleCodemod, NameResolutionMixin):
...
     ) -> cst.CSTNode:
-        if not self.filter_by_path_includes_or_excludes(
-            self.node_position(original_node)
-        ):
-            return updated_node
-
         if self.matches_startswith_endswith_or_pattern(original_node):
-            left_call = cst.ensure_type(updated_node.left, cst.Call)
-            right_call = cst.ensure_type(updated_node.right, cst.Call)
+            if not self.filter_by_path_includes_or_excludes(
```
clavedeluna: any reason you moved this down? It makes more sense to stay where it was so no work happens if this node is excluded. Also you could combine the if statements if you like.

src/core_codemods/combine_startswith_endswith.py
```diff
@@ -17,36 +18,54 @@ class CombineStartswithEndswith(SimpleCodemod, NameResolutionMixin):
...
 
-            new_arg = cst.Arg(
-                value=cst.Tuple(
-                    elements=[
-                        cst.Element(value=left_call.args[0].value),
-                        cst.Element(value=right_call.args[0].value),
-                    ]
-                )
-            )
+            elements = []
```
clavedeluna: could you move ln29-ln49 into its own method so it would be nice and clean
` new_arg = cst.Arg(value=cst.Tuple(elements=self._make_elements()))`

src/core_codemods/combine_startswith_endswith.py
```diff
@@ -17,36 +18,54 @@ class CombineStartswithEndswith(SimpleCodemod, NameResolutionMixin):
...
+            seen_evaluated_values = set()
+            for call in extract_boolean_operands(updated_node, ensure_type=cst.Call):
+                arg_value = call.args[0].value
+                if isinstance(arg_value, cst.Tuple):
+                    arg_elements = arg_value.elements
+                else:
+                    arg_elements = (cst.Element(value=arg_value),)
```
clavedeluna:
```suggestion
                    arg_elements = arg_elements = arg_value.elements if isinstance(arg_value, cst.Tuple) else (cst.Element(value=arg_value),)
```

src/core_codemods/combine_startswith_endswith.py
```diff
@@ -60,7 +79,24 @@ def matches_startswith_endswith_or_pattern(
...
+            operator=m.Or(),
+            right=startswith | endswith,
         )
+        if m.matches(node, chained_or):
+            return all(
+                call.func.value.value == node.right.func.value.value  # Same function
+                for call in extract_boolean_operands(node, ensure_type=cst.Call)
+            )
+
+        # No match
```
clavedeluna: no need for comment

tests/codemods/test_combine_startswith_endswith.py
```diff
@@ -52,3 +53,170 @@ def test_exclude_line(self, tmpdir):
...
+                "x.{func}((y, 'foo')) or x.{func}((z, 'bar'))",
+                "x.{func}((y, 'foo', z, 'bar'))",
+            ),
+        ],
+    )
+    def test_name_args(self, tmpdir, func, input_code, expected):
+        self._format_func_run_test(tmpdir, func, input_code, expected)
+
+    _100_of_each_type = [
+        (f"'{i}'", f"'{i}con' 'cat{i}'", f"f'fmt{i}'", f"name{i}") for i in range(100)
```
clavedeluna: I gotta ask, any reason we need 100? Seems like if you test with 3 or 4 that should prove the point that we can support Nth types.

clavedeluna (changes requested):
Small asks (I hope?). Lovely job here! You're a great contributor <3

src/core_codemods/combine_startswith_endswith.py
LucasFaudman: You're right, this was a mistake. I originally did this with the decorator below and just copied the code into the wrong spot when refactoring to try to make minimal changes. 

However I noticed that a lot of the codemods begin with a statement to check filters by includes/excludes and check that a node is not selected so I think a decorators could be cleaner and reduce repetition for these cases.

See [utils_decorators](https://github.com/LucasFaudman/codemodder-python/blob/main/src/codemodder/codemods/utils_decorators.py) in my fork.

```python
def check_filter_by_path_includes_or_excludes(
    func: Callable[[_BaseVisitorOrTransformer, _CSTNode, _CSTNode], cst.CSTNode]
) -> Callable[[_BaseVisitorOrTransformer, _CSTNode, _CSTNode], cst.CSTNode]:
    """
    Decorator to filter nodes based on path-includes or path-excludes flags.

    Calls the decorated func only if the original node is not filtered.
    otherwise, returns the updated node as is.

    ```
    @check_filter_by_path_includes_or_excludes
    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call) -> cst.Call
        # rest of the method
    ```

    Is equivalent to:

    ```
    def leave_Call(self, original_node: cst.Call, updated_node: cst.Call) -> cst.Call
        # Added by the decorator
        if not self.filter_by_path_includes_or_excludes(
            self.node_position(original_node)
            ):
            return updated_node

        # rest of the method
    ```
    """

    @wraps(func)
    def wrapper(
        instance: _BaseVisitorOrTransformer,
        original_node: _CSTNode,
        updated_node: _CSTNode,
    ) -> cst.CSTNode:
        if not instance.filter_by_path_includes_or_excludes(
            instance.node_position(original_node)
        ):
            return updated_node
        return func(instance, original_node, updated_node)

    return wrapper
```

clavedeluna: I like the decorator pattern! Some codemods don't need this excludes check which is why we didn't generalize like this, but there are enough that do that the pattern looks nice. @drdavella can give his opinion too

tests/codemods/test_combine_startswith_endswith.py
LucasFaudman: No, you're right this isn't strictly necessary was just getting familiar with the framework when I wrote this so I wanted to make sure I wasn't missing something by testing with a large number.

src/core_codemods/combine_startswith_endswith.py
```diff
@@ -60,7 +56,49 @@ def matches_startswith_endswith_or_pattern(
...
+        ):
+            return True
+
+        # Check for chained case: x.startswith("...") or x.startswith("...") or x.startswith("...") or ...
+        chained_or = m.BooleanOperation(
+            left=m.BooleanOperation(operator=m.Or()),
+            operator=m.Or(),
+            right=startswith | endswith,
         )
+        if m.matches(node, chained_or):
```
clavedeluna:
```suggestion
        if m.matches(node, m.BooleanOperation(
            left=m.BooleanOperation(operator=m.Or()),
            operator=m.Or(),
            right=startswith | endswith,
        )):
```

clavedeluna (approved):

PR #501, opened 2024-04-23
Add fixed/unfixed finding metadata support; populate for defectdojo

drdavella:
## Overview
*Add support for fixed and unfixed finding metadata and populate for defectdojo codemods*

## Details
* This is an important part of our storytelling and it was added to the CodeTF spec recently
* I expect there to be some iteration on the internal representation of these data models but this should get us started for now
* We'll need to go back and populate the same for our Sonar codemods as well


clavedeluna (approved):
Is an UnfixedFinding a finding that has not been resolved? or something else?

drdavella:
@clavedeluna yes, it's a finding that is theoretically addressed by this codemod but is unable to be fixed for one reason or another.

PR #510, opened 2024-04-24
Gracefully handle errors that occur during transformation

drdavella:


src/codemodder/codemods/libcst_transformer.py
```diff
@@ -269,9 +269,15 @@ def apply(
...
 
         tree = source_tree
-        with file_context.timer.measure("transform"):
-            for transformer in self.transformers:
-                tree = transformer.transform(tree, results, file_context)
+
+        try:
+            with file_context.timer.measure("transform"):
+                for transformer in self.transformers:
+                    tree = transformer.transform(tree, results, file_context)
```
clavedeluna, line 276: what about try/catching each transformation instead of all at the same time? that one if one fails another one can run

andrecsilva (approved):

src/codemodder/codemods/libcst_transformer.py
andrecsilva: This is about all the transforms in the pipeline and not codemodder. Each is meant to be essential for the overall transformation. If one fails we should not run the others.

PR #527, opened 2024-04-29
Adds a codemod that removes break or continue statements out of loop

### LINKED ISSUE #283: Codemod to remove `break` or `continue` outside of a loop
`break` or `continue` outside of a loop will result in a `SyntaxError` (but it's still parsable by libcst). Many linters have a rule to check for this. In most cases the correct thing is to remove these keywords, but there may be some cases where the user intended to put the keyword elsewhere. 

#### examples
- https://stackoverflow.com/questions/2462566/python-break-outside-loop
```
narg=len(sys.argv)
print "@length arg= ", narg
if narg == 1:
        print "@Usage: input_filename nelements nintervals"
        break
```
author probably meant to use something like `sys.exit` but removing the `break` is the right thing here       

- https://pylint.readthedocs.io/en/stable/user_guide/messages/error/not-in-loop.html
```
def print_even_numbers():
    for i in range(100):
        if i % 2 == 0:
            print(i)
    else:
        continue  # [not-in-loop]
```
intent is really unclear here, did the author mean to put `continue` under the if statement? In this case, if we just remove the `continue`, it will break the code. We should not do anything or remove the entire `else` clause

andrecsilva:
Also includes sonar version.

Closes #283

### DIFF
--- a/src/core_codemods/break_or_continue_out_of_loop.py
+++ b/src/core_codemods/break_or_continue_out_of_loop.py
@@ -39,8 +39,14 @@ class BreakOrContinueOutOfLoopTransformer(
         ancestors = self.path_to_root(original_node)
 
         # is it inside a for or while ?
-        if any(filter(lambda n: isinstance(n, cst.For | cst.While), ancestors)):
-            return updated_node
+        maybe_loop_ancestor = next(
+            filter(lambda n: isinstance(n, cst.For | cst.While), ancestors), None
+        )
+        match maybe_loop_ancestor:
+            # ensure it is not insde the else block of a while/for
+            case cst.For() | cst.While() as loop:
+                if loop.orelse not in ancestors:
+                    return updated_node
 
         self.report_change(original_node)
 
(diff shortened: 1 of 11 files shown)


tests/codemods/test_break_or_continue_out_of_loop.py
```diff
@@ -0,0 +1,58 @@
...
+from core_codemods.break_or_continue_out_of_loop import BreakOrContinueOutOfLoop
+
+
+class TestBreakOrContinueOutOfLoop(BaseCodemodTest):
+    codemod = BreakOrContinueOutOfLoop
+
+    def test_name(self):
+        assert self.codemod.name == "break-or-continue-out-of-loop"
+
+    def test_continue(self, tmpdir):
```
clavedeluna, line 11: an important test case is within the issue writeup, could you add it?

```
def print_even_numbers():
    for i in range(100):
        if i % 2 == 0:
            print(i)
    else:
        continue  # [not-in-loop]
```
removing the entire else clause would be the best thing here, but doing nothing is also an option.

clavedeluna (changes requested):

src/core_codemods/break_or_continue_out_of_loop.py
```diff
@@ -0,0 +1,63 @@
...
+    change_description = "Removed break or continue statement out of loop."
+
+    def _handle_break_or_continue(
+        self,
+        original_node: cst.Break | cst.Continue,
+        updated_node: cst.Break | cst.Continue,
+    ):
+        ancestors = self.path_to_root(original_node)
+        # is it inside a for or while ?
+        if any(filter(lambda n: isinstance(n, cst.For | cst.While), ancestors)):
```
clavedeluna: I would just remove the commend and move this check to a `_is_inside_loop` method, for clarity

tests/codemods/test_break_or_continue_out_of_loop.py
andrecsilva: ~~This case is covered in other tests, but I will add it nonetheless. It will replace the `continue` with `pass`.~~

andrecsilva: I was mistaken, I didn't realize loops could have an `else` clause. I went ahead and also added logic to remove the else entirely if empty.

clavedeluna (approved):
Looks awesome!

PR #494, opened 2024-04-20
Add new CodeMod CombineIsinstanceIssubclass. isistance(x, str) or isnstance(x, (bytes, list)) -> isinstance(x, (str, bytes, list))

LucasFaudman:
### `CombineIsinstanceIssubclass`
- New CodeMod very similar to `CombineStartswithEndswith` but for `isinstance` and `issubclass` calls.
```python
# Input
isinstance(a, int) or isinstance(a, float) or isinstance(a, str)
issubclass(a, str) or issubclass(a, (bytes, bytearray)) or issubclass(a, memoryview)

# Output
isinstance(a, (int, float, str))
issubclass(a, (str, bytes, bytearray, memoryview))
```

PR #537, opened 2024-05-01
Add a mechanism to enforce testing files for new codemods

### LINKED ISSUE #507: Enforce integration tests and all necessary metadata for a new codemod
See https://github.com/pixee/codemodder-python/pull/494#issuecomment-2073363786

The original PR added the codemod and unit tests and all pipeline steps pass. We really need to make sure to add an enforcement mechanism so the codemod is registered, an integration test is added, etc. I think by just enforcing adding an integration test that will force pretty much everything else, as an integration test won't run without it being registered. Then once the codemod is registered, if docs weren't added there's another test that will fail.

clavedeluna:
## Overview
*Use `coverage-threshold` to detect if a new codemod is missing unit or integration test*

## Description
It's common for contributors of new core_codemods to forget to add unit tests, and even more common, integration tests. [Here's](https://github.com/pixee/codemodder-python/pull/494/commits/207647f94cfb9a0fd37550750f706dfd7f8a1ae5) an example when a contributor got all the way to PR but forgot to add an integration test because it wasn't enforced anywhere. Without an integration test, we also don't enforce registering the new codemod or adding docs. While there may be rare cases when we want to just add the codemod, 99% of the time we want the testing to be there.

To remind us to add the right testing for a new codemod, I decided to use [coverage-threshold](https://github.com/DeanWay/coverage-threshold) after reading [this issue](https://github.com/nedbat/coveragepy/issues/691).  I also considered something in pre-commit or a pytest fixture to look at all core_codemods and find matching test files.

But adding this mechanism for per-file test coverage is clean and does the job. Setting `file_line_coverage_min` at 50 for any file in core_codemods correctly detects if a new codemod is untested or undertested, in both unit and integration tests. I moved our existing overall test coverage to `line_coverage_min` but allowed integration tests to have a lower configured value to pass.

Overall, I think this will reduce our mental load to remember to add the correct testing for a new codemod and will make it easier for new contributors. 🚀 

Closes #507

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,12 @@
 PYTEST = pytest -v
-COV_FLAGS = --cov-fail-under=93.5 --cov=codemodder --cov=core_codemods
+COV_FLAGS = --cov=codemodder --cov=core_codemods
 XDIST_FLAGS = --numprocesses auto
 
 test:
-	COVERAGE_CORE=sysmon ${PYTEST} ${COV_FLAGS} tests ${XDIST_FLAGS}
+	COVERAGE_CORE=sysmon ${PYTEST} ${COV_FLAGS} tests ${XDIST_FLAGS} && coverage json && coverage-threshold
 
 integration-test:
-	${PYTEST} integration_tests ${XDIST_FLAGS}
+	COVERAGE_CORE=sysmon ${PYTEST} integration_tests --cov=core_codemods ${XDIST_FLAGS}  && coverage json && coverage-threshold
 
 pygoat-test:
 	${PYTEST} -v ci_tests/test_pygoat_findings.py
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ test:
 	COVERAGE_CORE=sysmon ${PYTEST} ${COV_FLAGS} tests ${XDIST_FLAGS} && coverage json && coverage-threshold
 
 integration-test:
-	COVERAGE_CORE=sysmon ${PYTEST} integration_tests --cov=core_codemods ${XDIST_FLAGS}  && coverage json && coverage-threshold
+	COVERAGE_CORE=sysmon ${PYTEST} integration_tests --cov=core_codemods ${XDIST_FLAGS}  && coverage json && coverage-threshold --line-coverage-min 80
 
 pygoat-test:
 	${PYTEST} -v ci_tests/test_pygoat_findings.py
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,7 @@ get-hashes = 'codemodder.scripts.get_hashes:main'
 [project.optional-dependencies]
 test = [
     "coverage>=7.3,<7.6",
+    "coverage-threshold~=0.4",
     "Flask<4",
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -1,12 +1,12 @@
...
+COV_FLAGS = --cov=codemodder --cov=core_codemods
 XDIST_FLAGS = --numprocesses auto
 
 test:
-	COVERAGE_CORE=sysmon ${PYTEST} ${COV_FLAGS} tests ${XDIST_FLAGS}
+	COVERAGE_CORE=sysmon ${PYTEST} ${COV_FLAGS} tests ${XDIST_FLAGS} && coverage json && coverage-threshold
 
 integration-test:
-	${PYTEST} integration_tests ${XDIST_FLAGS}
+	COVERAGE_CORE=sysmon ${PYTEST} integration_tests --cov=core_codemods ${XDIST_FLAGS}  && coverage json && coverage-threshold --line-coverage-min 80
```
andrecsilva, line 9: My suggestion is to do away with overall `line-coverage` here (and just keep the per-file one). 

Looking at the failed run in the previous commit, the line coverage for integration tests varies wildly and this will have to be tweaked a lot. Also, since we are already enforcing this for unit tests, it feels redundant here.

clavedeluna: The overall line coverage for integration tests doesn't vary, it's always above 80% which is much lower than unit tests. And that of course makes sense since integration tests don't intend to cover all paths.

That being said, I agree total line coverage doesn't make much sense for integration tests, but given how `coverage-threshold` is configured, I have to pass in `--line-coverage-min #` because it first uses whatever I configured in pyproject.toml, which is the 93%+ for unit tests. Passing it as a CLI flag overrides that. Maybe it makes sense to just set it to `--line-coverage-min 0`? But that at point, 80 is just as well, all integration test runs I see are 83%+

andrecsilva: See here:
https://github.com/pixee/codemodder-python/actions/runs/8910019509/job/24468409979
Coverage can go as low as 50% and is achieved by at least one file:
```
 /opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/site-packages/core_codemods/defectdojo/semgrep/avoid_insecure_deserialization.py      20     10    50%
```
That's what I meant by varying wildly.

clavedeluna: I think we're ok with that.

pyproject.toml
```diff
@@ -110,3 +111,9 @@ extend-exclude = '''
   src/codemodder/_version.py
 )/
 '''
+
+[coverage-threshold]
+line_coverage_min = 93.5
+    [coverage-threshold.modules."src/core_codemods/"]
```
drdavella: I think this can be at the same level of indentation as above.

drdavella (approved):

pyproject.toml
clavedeluna: you're right! I copied the format [exactly](https://github.com/DeanWay/coverage-threshold?tab=readme-ov-file#config) as shown but they must have an indent mistake

PR #549, opened 2024-05-08
Updating importance level

whodes:
## Overview
Some of the codemods are marked as MEDIUM importance, when they should be marked as LOW. Updating the code to reflect that.


drdavella (approved):
Thank you @whodes!

PR #557, opened 2024-05-10
Report unfixed when parsing/transforming errors

### LINKED ISSUE #514: Parser and transformation failures should also update unfixed findings
For codemods that fix tool results, if we fail to parse or transform a particular file, we should also update the unfixed findings in CodeTF to reflect the fact that we won't be able to fix findings for that codemod in that file.

clavedeluna:
## Overview
*In a codemodder run that fixes tool results, if there is a file parsing or transforming error, we will report that as an unfixed result for tools that currently support `Finding`*

Defectdojo is the only tool we currently support adding `Finding`s for so that's what I used for testing here.

Closes #514

### DIFF
--- a/src/codemodder/codemods/libcst_transformer.py
+++ b/src/codemodder/codemods/libcst_transformer.py
@@ -264,17 +264,20 @@ class LibcstTransformerPipeline(BaseTransformerPipeline):
                     source_tree = cst.parse_module(f.read())
         except Exception:
             file_context.add_failure(file_path)
+            if context.fixing_tool_results:
+                file_context.report_unfixed_for_file(reason="parsing file")
             logger.exception("error parsing file %s", file_path)
             return None
 
         tree = source_tree
-
         try:
             with file_context.timer.measure("transform"):
                 for transformer in self.transformers:
                     tree = transformer.transform(tree, results, file_context)
         except Exception:
             file_context.add_failure(file_path)
+            if context.fixing_tool_results:
+                file_context.report_unfixed_for_file(reason="transforming file")
             logger.exception("error transforming file %s", file_path)
             return None
 
--- a/src/codemodder/codemods/libcst_transformer.py
+++ b/src/codemodder/codemods/libcst_transformer.py
@@ -263,10 +263,8 @@ class LibcstTransformerPipeline(BaseTransformerPipeline):
                 with open(file_path, "r", encoding="utf-8") as f:
                     source_tree = cst.parse_module(f.read())
         except Exception:
-            file_context.add_failure(file_path)
-            if context.fixing_tool_results:
-                file_context.report_unfixed_for_file(reason="parsing file")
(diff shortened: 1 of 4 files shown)


src/codemodder/context.py
```diff
@@ -77,6 +77,7 @@ def __init__(
         self.path_exclude = path_exclude
         self.max_workers = max_workers
         self.tool_result_files_map = tool_result_files_map or {}
+        self.fixing_tool_results = bool(self.tool_result_files_map)
```
clavedeluna: this is the only clever way I figured to distinguish between a run for find-n-fix codemods versus tool fixing codemods

andrecsilva: Will this check correctly distinguish find-n-fix codemods that uses Semgrep?

src/codemodder/codemods/libcst_transformer.py
```diff
@@ -265,17 +265,20 @@ def apply(
                     source_tree = cst.parse_module(f.read())
         except Exception:
             file_context.add_failure(file_path)
+            if context.fixing_tool_results:
+                file_context.report_unfixed_for_file(reason="parsing file")
```
drdavella:
```suggestion
                file_context.report_unfixed_for_file(reason="Failed to parse file")
```

src/codemodder/codemods/libcst_transformer.py
```diff
@@ -265,17 +265,20 @@ def apply(
...
         tree = source_tree
-
         try:
             with file_context.timer.measure("transform"):
                 for transformer in self.transformers:
                     tree = transformer.transform(tree, results, file_context)
         except Exception:
             file_context.add_failure(file_path)
+            if context.fixing_tool_results:
+                file_context.report_unfixed_for_file(reason="transforming file")
```
drdavella:
```suggestion
                file_context.report_unfixed_for_file(reason="Failed to transform file")
```

drdavella (changes requested):
This is on the right track but I think the right way to do this is by updating the `add_failure` method of `FileContext`. `FileContext` already knows about all findings associated with that particular file, so it simply becomes a matter of building unfixed findings with the given reason.

src/codemodder/context.py
drdavella: I don't think this will be necessary if we handle this within `FileContext`.

src/codemodder/codemods/libcst_transformer.py
```diff
@@ -263,19 +263,18 @@ def apply(
                 with open(file_path, "r", encoding="utf-8") as f:
                     source_tree = cst.parse_module(f.read())
         except Exception:
-            file_context.add_failure(file_path)
-            logger.exception("error parsing file %s", file_path)
+            file_context.add_failure(file_path, reason := "Failed to parse file")
```
drdavella, line 266: Neat ❤️

drdavella (approved):

PR #561, opened 2024-05-13
Adds extraction of code flows and related locations

andrecsilva:
## Overview
Adds extraction of code flows and related locations from sarif/json files.

## Description

For some SAST results, the fix to the problem may involve changes somewhere in the reported codeflows. This PR extracts and those codeflows and related locations and make those available to use for codemods.

### DIFF
--- a/src/codemodder/codeql.py
+++ b/src/codemodder/codeql.py
@@ -3,7 +3,7 @@ from pathlib import Path
 
 from typing_extensions import Self
 
-from codemodder.result import LineInfo, Location, Result, ResultSet
+from codemodder.result import LineInfo, Location, LocationWithMessage, Result, ResultSet
 from codemodder.sarifs import AbstractSarifToolDetector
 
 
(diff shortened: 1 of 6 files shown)


src/codemodder/semgrep.py
```diff
@@ -71,7 +71,29 @@ def from_sarif(
...
+        all_flows: list[list[Location]] = [
+            [
+                SemgrepLocation.from_sarif(locations.get("location"))
+                for locations in threadflow.get("locations", {})
+            ]
+            for codeflow in sarif_result.get("codeFlows", {})
+            for threadflow in codeflow.get("threadFlows", {})
+        ]
+        related_locations: list[LocationWithMessage] = []
+        if "relatedLocations" in sarif_result.keys():
```
drdavella: Small thing but this is redundant:
```suggestion
        if "relatedLocations" in sarif_result:
```

src/codemodder/codeql.py
```diff
@@ -52,7 +52,29 @@ def from_sarif(
...
+        all_flows: list[list[Location]] = [
+            [
+                CodeQLLocation.from_sarif(locations.get("location"))
+                for locations in threadflow.get("locations", {})
+            ]
+            for codeflow in sarif_result.get("codeFlows", {})
+            for threadflow in codeflow.get("threadFlows", {})
+        ]
+        related_locations: list[LocationWithMessage] = []
+        if "relatedLocations" in sarif_result.keys():
```
drdavella:
```suggestion
        if "relatedLocations" in sarif_result:
```

src/codemodder/semgrep.py
```diff
@@ -71,7 +71,29 @@ def from_sarif(
         for location in sarif_result["locations"]:
             artifact_location = SemgrepLocation.from_sarif(location)
             locations.append(artifact_location)
-        return cls(rule_id=rule_id, locations=locations)
+        all_flows: list[list[Location]] = [
```
drdavella, line 74: Is the only difference between this and CodeQL the location class? Seems like a good opportunity for some kind of generics pattern but it's not that important.

drdavella (approved):

src/codemodder/semgrep.py
andrecsilva: For now yes, at least this particular part (codeflows) are the same. They both follow the spec:
https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790990

I think we may be able to generalize all of them with a single class parameterized by location objects. I'll write a ticket for it.

PR #567, opened 2024-05-14
Populate Sonar findings

### LINKED ISSUE #503: Populate fixed and unfixed finding metadata for Sonar codemods
See #501. We need to add the same kinds of changes for Sonar codemods in order to provide better storytelling about what is actually being fixed.

clavedeluna:
## Overview
*Sonar results should now populate `finding=...` and report these as either fixed or unfixed*

## Description

* I renamed some methods for consistency. We could enforce a protocol later on.
* I'm not 100% sure the values I set for sonar's `Finding(...)` are correct?
* We don't have a sonar codemod that calls `report_unfixed` nor could I think of one at the top of my head, so. I just tested this as a transformer error to ensure a Finding is correctly populated as unfixed.

Closes #503

### DIFF
--- a/src/core_codemods/sonar/results.py
+++ b/src/core_codemods/sonar/results.py
@@ -6,8 +6,9 @@ from pathlib import Path
 import libcst as cst
 from typing_extensions import Self
 
+from codemodder.codetf import Finding, Rule
 from codemodder.logging import logger
-from codemodder.result import LineInfo, Location, Result, ResultSet
+from codemodder.result import LineInfo, Location, ResultSet, SASTResult
 
 
 class SonarLocation(Location):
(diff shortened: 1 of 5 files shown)


andrecsilva (approved):

src/core_codemods/sonar/results.py
```diff
@@ -36,7 +37,21 @@ def from_result(cls, result) -> Self:
...
+            finding_id=rule_id,
+            rule_id=rule_id,
+            locations=locations,
+            codeflows=all_flows,
+            finding=Finding(
+                id=rule_id,
+                rule=Rule(
+                    id=rule_id,
+                    name=rule_id,
+                    url=None,
```
drdavella, line 51: I think we _could_ do the URL here but I'm fine to leave it as a separate issue. It would be something like
```python
url="https:/r/ules.sonarsource.com/python/RSPEC-<rule-id-here>/"
```

drdavella (approved):

src/core_codemods/sonar/results.py
clavedeluna: unfortunately I don't think we've yet figured out why sonar urls are inconsistent. grep `rules.sonarsource.com/` in our code and you see stuff like
https://rules.sonarsource.com/python/type/Vulnerability/RSPEC-5144/
https://rules.sonarsource.com/python/type/Bug/RSPEC-6725/

so either there is a generalized url for all rules and we just copied the wrong thing or there is no generalized rule url.

quick url check says we have urls that can be generalized

PR #586, opened 2024-05-22
mkdir codemod fixes

clavedeluna:
#577 wasn't    quite ready to be merged, but no worries, the fixes are here!

- correctly pass leading lines
- unit test to show codemod works in function scope
- report unfixed for cases we don't yet handle
- add missing custom node is selected logic and test

### DIFF
--- a/src/core_codemods/tempfile_mktemp.py
+++ b/src/core_codemods/tempfile_mktemp.py
@@ -4,7 +4,6 @@ from typing import Optional
 import libcst as cst
 from libcst import matchers
 from libcst.codemod import CodemodContext
-from libcst.metadata import ScopeProvider
 
 from codemodder.codemods.libcst_transformer import (
     LibcstResultTransformer,
@@ -34,7 +33,8 @@ class TempfileMktempTransformer(
         super().__init__(context, results, file_context, _transformer)
 
     def visit_Call(self, node: cst.Call) -> None:
-        self.mktemp_calls.add(node) if self._is_mktemp_call(node) else None
+        if self._is_mktemp_call(node):
+            self.mktemp_calls.add(node)
 
     def leave_SimpleStatementLine(
         self,
@@ -157,21 +157,6 @@ class TempfileMktempTransformer(
                             wrapper_func = call.func
                             return (wrapper_func, maybe_value)
         return None
-
-
-class GatherMkTempCallsVisitor(cst.CSTVisitor, NameAndAncestorResolutionMixin):
-    METADATA_DEPENDENCIES = (ScopeProvider,)
-
-    def __init__(self):
-        super().__init__()
-        self.calls = []
-
-    def visit_Call(self, node: cst.Call) -> None:
-        (
-            self.calls.append(node)
-            if self.find_base_name(node.func) == "tempfile.mktemp"
-            else None
-        )
 
 
 TempfileMktemp = CoreCodemod(
--- a/src/core_codemods/tempfile_mktemp.py
+++ b/src/core_codemods/tempfile_mktemp.py
@@ -48,11 +48,16 @@ class TempfileMktempTransformer(
             case cst.SimpleStatementLine(body=[bsstmt]):
                 if maybe_tuple := self._is_assigned_to_mktemp(bsstmt):
                     assign_name, call = maybe_tuple
-                    return self.report_and_change(call, assign_name)
+                    return self.report_and_change(
+                        call, assign_name, original_node.leading_lines
+                    )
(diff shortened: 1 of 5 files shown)


src/core_codemods/tempfile_mktemp.py
```diff
@@ -115,6 +159,21 @@ def _mktemp_is_sink(
...
+        self.calls = []
+
+    def visit_Call(self, node: cst.Call) -> None:
+        (
+            self.calls.append(node)
+            if self.find_base_name(node.func) == "tempfile.mktemp"
+            else None
+        )
+
+
```
andrecsilva: Is this used anywhere?

src/core_codemods/tempfile_mktemp.py
```diff
@@ -19,25 +23,64 @@ class TempfileMktempTransformer(
...
+        context: CodemodContext,
+        results: list[Result] | None,
+        file_context: FileContext,
+        _transformer: bool = False,
+    ):
+        self.mktemp_calls: set[cst.Call] = set()
+        super().__init__(context, results, file_context, _transformer)
+
+    def visit_Call(self, node: cst.Call) -> None:
+        self.mktemp_calls.add(node) if self._is_mktemp_call(node) else None
```
andrecsilva: The `else` is useless here.

src/core_codemods/tempfile_mktemp.py
```diff
@@ -19,25 +23,64 @@ class TempfileMktempTransformer(
...
+                        wrapper_func_name,
+                        original_node.leading_lines,
+                        assignment=False,
+                    )
+
+        # If we get here it's because there is a mktemp call but we haven't fixed it yet.
+        for unfixed in self.mktemp_calls:
+            self.report_unfixed(unfixed, reason="Pixee does not yet support this fix.")
+        self.mktemp_calls.clear()
         return updated_node
```
andrecsilva, line 67: This logic feels weird here.

Say a mktemp call `mcall` is contained in the statement, `leave_Call` will be called first and `mcall` will be added to `self.mktemp_calls`. 

Afterwards, `leave_simpleStatementLine` will run. If `mcall` is fixed, it will be reported as changed, but since its removal from `self.mktemp_calls` only happens at the end, it will also be reported as unfixed.

In short, can you add a test to show that no call is being reported as fixed and unfixed at the same time?

andrecsilva (changes requested):
Some small concerns.

src/core_codemods/tempfile_mktemp.py
clavedeluna: you can't write a ternary expression in Python without an else. Just wanted to make it one line instead of
```
if self._is_mktemp_call(node):
    self.mktemp_calls.add(node)
```

src/core_codemods/tempfile_mktemp.py
clavedeluna: I checked thought I checked this. The key is the `return` statements above. If the code path hits the fixed cases we have handled, it returns, never reaching this code. Easy enough to add an assert statement to demonstrate, so will do.

src/core_codemods/tempfile_mktemp.py
clavedeluna: woops thanks for the good eye, old code

src/core_codemods/tempfile_mktemp.py
clavedeluna: see https://github.com/pixee/codemodder-python/pull/586/files#diff-d452b243674c71654239be30328a960ed2c1e0aea06bc5b57c1ab7c8169c3cb9R43

andrecsilva: > I checked thought I checked this. The key is the `return` statements above. If the code path hits the fixed cases we have handled, it returns, never reaching this code. Easy enough to add an assert statement to demonstrate, so will do.

The problem is when you have two or more mktemp calls. If the first one is fixed but the second one is not. When you report the second one, the first should still be included in `self.mktemp_calls`. This results in it being reported as unfixed.

clavedeluna: Example please. Because what you're describing sounds like a super edge case of two calls in the same line and we don't really handle those cases or test that for any codemod right now.

andrecsilva: Just include `self.mktemp_calls.clear()` before the returns in `leave_SimpleStatementLine` and I'll be satisfied.

andrecsilva (approved):

PR #592, opened 2024-05-23
Enable use of Azure OpenAI client

drdavella:
## Overview
*Enable use of Azure OpenAI client as alternative to OpenAI*

## Details
* https://github.com/pixee/codemodder-specs/pull/30

### DIFF
--- a/src/codemodder/context.py
+++ b/src/codemodder/context.py
@@ -23,15 +23,20 @@ from codemodder.registry import CodemodRegistry
 from codemodder.utils.timer import Timer
 
 try:
-    from openai import Client
+    from openai import AzureOpenAI, OpenAI
 except ImportError:
-    Client = None
+    OpenAI = None
+    AzureOpenAI = None
 
 
 if TYPE_CHECKING:
-    from openai import Client
+    from openai import OpenAI
 
     from codemodder.codemods.base_codemod import BaseCodemod
+
+
+class MisconfiguredAIClient(ValueError):
+    pass
 
 
 class CodemodExecutionContext:
@@ -49,7 +54,7 @@ class CodemodExecutionContext:
     path_exclude: list[str]
     max_workers: int = 1
     tool_result_files_map: dict[str, list[str]]
-    llm_client: Client | None = None
+    llm_client: OpenAI | None = None
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


src/codemodder/context.py
```diff
@@ -80,16 +89,39 @@ def __init__(
...
-        if not Client:
-            logger.debug("OpenAI API client not available")
+    def _setup_llm_client(self) -> OpenAI | None:
+        if not AzureOpenAI:
+            logger.info("Azure OpenAI API client not available")
+            return None
+
+        azure_openapi_key = os.getenv("CODEMODDER_AZURE_OPENAI_API_KEY")
+        azure_openapi_endpoint = os.getenv("CODEMODDER_AZURE_OPENAI_ENDPOINT")
+        if bool(azure_openapi_key) ^ bool(azure_openapi_endpoint):
```
clavedeluna, line 99: could use walrus here for 2 lines above

src/codemodder/context.py
```diff
@@ -80,16 +89,39 @@ def __init__(
         self.tool_result_files_map = tool_result_files_map or {}
         self.llm_client = self._setup_llm_client()
 
-    def _setup_llm_client(self) -> Client | None:
-        if not Client:
-            logger.debug("OpenAI API client not available")
+    def _setup_llm_client(self) -> OpenAI | None:
+        if not AzureOpenAI:
```
clavedeluna, line 93: the PR description is not very descriptive so I'm guessing here but if the goal is to use either OpenAI or AzureOpenAI (or nothing), shouldn't the `if not OpenAI:` check also be up here and not short circuit ` if not AzureOpenAI:``

tests/test_context.py
```diff
@@ -77,3 +82,120 @@ def test_failed_dependency_description(self, mocker):
 ```"""
             in description
         )
+
+    def test_setup_llm_client_no_env_vars(self, mocker):
```
clavedeluna, line 86: could add some `caplog` assertions for the logging, maybe that would clear up my misunderstanding above

clavedeluna (changes requested):

src/codemodder/context.py
drdavella: The order of precedence is to check for Azure OpenAI first and then OpenAI as per the spec linked above. However there's not really going to be a case where `AzureOpenAI` is defined but `OpenAI` is not. That means the check below for `OpenAI` is theoretically redundant although it does help satisfy the type checker.

src/codemodder/context.py
drdavella: That's true although I think in this case it might make it a little harder to read 😅

src/codemodder/context.py
clavedeluna: ah thanks for clarifying. if there isn't such case you could `if not AzureOpenAI or OpenAI` but it's all good

clavedeluna (approved):

src/codemodder/context.py
drdavella: That's true, that's probably a bit clearer.

PR #598, opened 2024-05-27
Xml transformer

### LINKED ISSUE #597: Update XML parser to handle adding new and nested elements
We should cover these cases:

1. add a new element (nested inside the root element)
```
<root>
    <child1>Content 1</child1>
</root>
```
becomes
```
<root>
    <child1>Content 1</child1>
    <child2>Content 2</child2>
</root>
```
2. add a new element nested anywhere
```
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
  <system.web>
  </system.web>
  <system.webServer>
    <httpProtocol>
    </httpProtocol>
  </system.webServer>
</configuration>
```
becomes
```
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
  <system.web>
  </system.web>
  <system.webServer>
    <httpProtocol>
      <customHeaders>
        <add name="X-Frame-Options" value="SAMEORIGIN" />
      </customHeaders>
    </httpProtocol>
  </system.webServer>
</configuration>
```

3. add a new nested element with new tags
```
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
  <system.web>
  </system.web>
  <system.webServer>
    <httpProtocol>
      <customHeaders>
        <add name="X-Frame-Options" value="SAMEORIGIN" />
      </customHeaders>
    </httpProtocol>
  </system.webServer>
</configuration>
```
becomes
```
<?xml version="1.0" encoding="utf-8" ?>
<configuration>
  <system.web>
  </system.web>
  <system.webServer>
    <httpProtocol>
      <customHeaders>
        <add name="X-Frame-Options" value="SAMEORIGIN" />
        <add name="Content-Security-Policy" value="default-src 'self';" />
      </customHeaders>
    </httpProtocol>
  </system.webServer>
</configuration>
```

clavedeluna:
## Overview
*XML Transformer can now add new and new nested elements*

Follows the original structure but adds a new XML class to handle these cases. One concern we may want to handle add some point is formatting. While the transformer adds the new elements at the right location (nested inside the right place), the formatting doesn't necessarily follow the original file.

Closes #597

### DIFF
--- a/src/codemodder/codemods/xml_transformer.py
+++ b/src/codemodder/codemods/xml_transformer.py
@@ -1,4 +1,5 @@
 import mmap
+from dataclasses import dataclass, field
 from tempfile import TemporaryFile
 from xml.sax import SAXParseException, handler
 from xml.sax.handler import LexicalHandler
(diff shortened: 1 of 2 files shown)


andrecsilva (approved):
It may be worth looking into something like [Beautiful soup](https://beautiful-soup-4.readthedocs.io/en/latest/) to format the xml file before writing it.

clavedeluna:
> It may be worth looking into something like [Beautiful soup](https://beautiful-soup-4.readthedocs.io/en/latest/) to format the xml file before writing it.

Yes IIRC beautifusoup does a decent job keeping formatting and newlines. In fact we could've used that library instead of xml XMLGenerator for parsing and updating the xml. Did you look into that initially? If so, were there downsides, other than performance for large files?

andrecsilva:
> > It may be worth looking into something like [Beautiful soup](https://beautiful-soup-4.readthedocs.io/en/latest/) to format the xml file before writing it.
> 
> Yes IIRC beautifusoup does a decent job keeping formatting and newlines. In fact we could've used that library instead of xml XMLGenerator for parsing and updating the xml. Did you look into that initially? If so, were there downsides, other than performance for large files?

The problem is the location. Beautifulsoup uses `lxml` as a backend to parse xml files and `lxml` will only report the line of an element but not the column.

PR #599, opened 2024-05-27
Update XML writing

### LINKED ISSUE #591: Bug: SystemError with XML codemod on MacOS
When running the unit test suite on my Macbook I get the following error:
```python
            # don't write anything if no changes were issued
            # avoids simply formatting the file
            if changes and not context.dry_run:
                with open(file_context.file_path, "w+b") as original:
                    # mmap can't map empty files, write something first
                    original.write(b"a")
                    # copy contents of result into original file
                    # the snippet below preserves the original file metadata and accounts for large files.
                    output_file.seek(0)
                    output_mmap = mmap.mmap(output_file.fileno(), 0)

                    original.truncate()
                    original_mmap = mmap.mmap(original.fileno(), 0)
>                   original_mmap.resize(len(output_mmap))
E                   SystemError: mmap: resizing not available--no mremap()
```

clavedeluna:
## Overview
Use normal file read/write operations instead of `mmap`. There was no benefit to use it (and an occasional error in Mac came up) because we need the entire file for diff.
Also added some missing error handling and testing
Closes #591

### DIFF
--- a/src/codemodder/codemods/xml_transformer.py
+++ b/src/codemodder/codemods/xml_transformer.py
@@ -1,7 +1,6 @@
-import mmap
 from dataclasses import dataclass, field
 from tempfile import TemporaryFile
-from xml.sax import SAXParseException, handler
+from xml.sax import handler
 from xml.sax.handler import LexicalHandler
 from xml.sax.saxutils import XMLGenerator
 from xml.sax.xmlreader import AttributesImpl, Locator
@@ -13,6 +12,7 @@ from codemodder.codetf import Change, ChangeSet
 from codemodder.context import CodemodExecutionContext
 from codemodder.diff import create_diff
 from codemodder.file_context import FileContext
+from codemodder.logging import logger
 from codemodder.result import Result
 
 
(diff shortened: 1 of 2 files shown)


src/codemodder/codemods/xml_transformer.py
```diff
@@ -190,36 +191,27 @@ def apply(
...
                 changes = transformer_instance.changes
                 output_file.seek(0)
+            except Exception:
+                file_context.add_failure(
+                    file_path, reason := "Failed to parse XML file"
+                )
+                logger.exception("%s %s", reason, file_path)
+                return None
 
-            except SAXParseException:
```
clavedeluna, line 194: specific is better than general but I chose to switch to the general exception to more closely follow libcst transformer

andrecsilva (approved):

src/codemodder/codemods/xml_transformer.py
```diff
@@ -171,12 +171,13 @@ def apply(
         file_context: FileContext,
         results: list[Result] | None,
     ) -> ChangeSet | None:
-        if file_context.file_path.suffix.lower() not in (".config", ".xml"):
+        if (file_path := file_context.file_path).suffix.lower() not in (
```
drdavella: Actually isn't this already handled by `default_extensions` on the codemod itself?

drdavella (approved):
Looks good, passes on my machine 👍

src/codemodder/codemods/xml_transformer.py
clavedeluna: Looks like it is, perhaps @andrecsilva added it to prevent any other files from continuing on which is pretty reasonable. Want me to remove it?

drdavella: I think we should trust the codemod to do this filtering since there may be other extensions we don't anticipate quite yet.

PR #600, opened 2024-05-28
generalize sonar urls

### LINKED ISSUE #569: Generalize sonar rule urls
See https://github.com/pixee/codemodder-python/pull/567#discussion_r1600062936

1. Look over all the sonar rule urls and determine if they can be replaced with `url="https:/r/ules.sonarsource.com/python/RSPEC-<rule-id-here>/"` for a more general look
2. If so, update them
3. also update [src/core_codemods/sonar/results.py](https://github.com/pixee/codemodder-python/pull/567/files/4d98ed4a673804a16356e37f6a7a0021420d5224#diff-0e27daff2e9cf8c4c921028f8daa0e6fa713b3153952e0e4f63d0a53a3a83792) `Finding(...url="https:/r/ules.sonarsource.com/python/RSPEC-<rule-id-here>/"`

clavedeluna:
## Overview
*Sonar urls can be constructed from rule id*

I ran a quick script on the results and they're all valid URLs.

Closes #569

### DIFF
--- a/src/core_codemods/sonar/api.py
+++ b/src/core_codemods/sonar/api.py
@@ -7,7 +7,7 @@ from codemodder.codemods.base_transformer import BaseTransformerPipeline
 from codemodder.context import CodemodExecutionContext
 from codemodder.result import ResultSet
 from core_codemods.api.core_codemod import CoreCodemod, SASTCodemod
-from core_codemods.sonar.results import SonarResultSet
+from core_codemods.sonar.results import SonarResultSet, sonar_url_from_id
 
 
 class SonarCodemod(SASTCodemod):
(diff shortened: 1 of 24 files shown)


drdavella (approved):
This is great thanks!

> I ran a quick script on the results and they're all valid URLs.

Any chance you could commit this so we could make it part of CI/CD? Or just use it as a sanity check?

clavedeluna:
> This is great thanks!
> 
> > I ran a quick script on the results and they're all valid URLs.
> 
> Any chance you could commit this so we could make it part of CI/CD? Or just use it as a sanity check?

it's very much a script that involved some copy paste so it isn't really for adding to CI directly. However, thinking a bit more deeply, if you're concerned about our codemods having accessible and valid urls, I could do something like: for any codemod in the registry visit the url and check it's a 200. WE could add that as an integration test since it may take a while. If that sounds good, I can ticket that as a nice "when we have time" test addition.

Issue #607, opened 2024-05-30
BUG: SQL parameterization codemod fails to remediate Sonar finding


drdavella:
See the Sonar findings attached to this PR: https://github.com/drdavella/pygoat-sonar/pull/5

Codemodder fails to remediate the Sonar issue for the SQL parameterization.

Search for this issue number to find an xfailing test case.

andrecsilva:
@drdavella The test added in #608 doesn't match the location file (the sonar result points to line 20 and the file has 19 lines), are you sure you've added the right one?

In any case, the transformation works, but it is failing to match the location.

drdavella:
@andrecsilva it's possible that the unit test isn't quite right, but this result file was taken directly from an actual Sonar scan. You can try checking out this branch on my fork of `pygoat`: https://github.com/drdavella/pygoat-sonar/pull/5/files

And then reproduce with this codemodder command:
```
codemodder $HOME/pygoat-sonar --path-include="introduction/new_view.py" --codemod-include="sonar:python/sql-parameterization-S3649" --dry-run --verbose --sonar-issues-json $HOME/codemodder-python/tests/samples/sonar/sql_parameterization.json
```

The source file matches the results but no fix is produced.

PR #608, opened 2024-05-30
Add (failing) test case for Sonar SQL finding

drdavella:
Adds a failing test case for the bug described in #607


tests/samples/sonar/issues.json
```diff
@@ -0,0 +1 @@
...nent":"drdavella_pygoat-sonar2:introduction/new_view.py","textRange":{"startLine":17,"endLine":17,"startOffset":16,"endOffset":41},"msg":"Source: a user can craft an HTTP request with malicious content"}]}],"status":"OPEN","message":"Change this code to not construct SQL queries directly from user-controlled data.","effort":"30min","debt":"30min","tags":["cwe","sql"],"creationDate":"2024-05-30T18:34:05+0200","updateDate":"2024-05-30T18:35:24+0200","type":"VULNERABILITY","organization":"drdavella","pullRequest":"5","cleanCodeAttribute":"COMPLETE","cleanCodeAttributeCategory":"INTENTIONAL","impacts":[{"softwareQuality":"SECURITY","severity":"HIGH"}]}],"components":[{"organization":"drdavella","key":"drdavella_pygoat-sonar2:introduction/new_view.py","uuid":"AY_KW3gxkpLwWuMztk5a","enabled":true,"qualifier":"FIL","name":"new_view.py","longName":"introduction/new_view.py","path":"introduction/new_view.py","pullRequest":"5"},{"organization":"drdavella","key":"drdavella_pygoat-sonar2","uuid":"AY_KWyTh-nOrCSyUsFgG","enabled":true,"qualifier":"TRK","name":"pygoat-sonar","longName":"pygoat-sonar","pullRequest":"5"}],"organizations":[{"key":"drdavella","name":"Dan D\u0027Avella"}],"facets":[]}
```
clavedeluna: this file name will not get confused with tests/samples/sonar_issues.json :sweat_smile:

clavedeluna: in all seriousness, do we need both files?

tests/codemods/sonar/test_sonar_sql_parameterization.py
```diff
@@ -58,3 +63,51 @@ def f():
...
+                conn = sqlite3.connect("example")
+                result = conn.cursor().execute(sql, (user, ))
+
+                json_response = json.dumps({"user": result.fetchone()[0]})
+                return HttpResponse(json_response.encode("utf-8"))
+            else:
+                return redirect("/")
+        """.lstrip()
+
+        issues = json.dumps(SAMPLE_FILE_PATH.read_text())
```
clavedeluna: if the issue is still reproduced by just having issues = {...} as all other unit tests that would be a bit cleaner and easier to read which issue it is. Right now the test fails bc no results are matched, ` logger.debug("No results for %s", self.id)`

drdavella: Ah good catch, thank you! I want to represent this finding as accurately as possible and I think it might be a little unwieldy to include the whole thing in the source file, but I'll see what I can do. In any case I need to fix this problem.

tests/samples/sonar/issues.json
drdavella: Uhhh let me see about that. I might need to make this name more specific to the particular test that uses it.

clavedeluna (approved):

PR #615, opened 2024-06-03
Fixed a bug, adjusted tests

### LINKED ISSUE #607: BUG: SQL parameterization codemod fails to remediate Sonar finding

andrecsilva:
## Overview
Closes #607

### DIFF
--- a/tests/codemods/sonar/test_sonar_sql_parameterization.py
+++ b/tests/codemods/sonar/test_sonar_sql_parameterization.py
@@ -4,12 +4,10 @@ from pathlib import Path
 from codemodder.codemods.test import BaseSASTCodemodTest
 from core_codemods.sonar.sonar_sql_parameterization import SonarSQLParameterization
 
-SAMPLE_FILE_PATH = (
-    Path(__file__).parents[2] / "samples" / "sonar" / "sql_parameterization.json"
-)
-SAMPLE_FILE_PATH_2 = (
-    Path(__file__).parents[2] / "samples" / "sonar" / "sql_parameterization2.json"
-)
+SAMPLE_FILE_PATH = [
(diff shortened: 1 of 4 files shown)


clavedeluna (approved):
LG on my end but I think @drdavella had a reason for keeping the test as is so since you're changing it you might want to provide an explanation as to why otherwise it isn't clear if the bug is fixed or if the test change circumvents it

drdavella:
@clavedeluna this was my fault, I pointed at the wrong example PR in #607. I'll confirm on my end whether this fixes the problem.

EDIT: I was using this as the test case: https://github.com/drdavella/pygoat-sonar/pull/1

tests/codemods/sonar/test_sonar_sql_parameterization.py
```diff
@@ -1,14 +1,15 @@
...
 
-import pytest
-
 from codemodder.codemods.test import BaseSASTCodemodTest
 from core_codemods.sonar.sonar_sql_parameterization import SonarSQLParameterization
 
 SAMPLE_FILE_PATH = (
     Path(__file__).parents[2] / "samples" / "sonar" / "sql_parameterization.json"
 )
+SAMPLE_FILE_PATH_2 = (
```
drdavella: Just a tiny thing but I think I prefer the pattern of making `SAMPLE_FILE_PATHS = [ ... ]` instead of encoding indices in the name.

drdavella (approved):

PR #623, opened 2024-06-04
LLM and test refactor

clavedeluna:
## Overview
*This PR moves LLM initialization into its own module and updates tests for easier api*

## Description

* We want to be able to initialize an LLM client in testing, outside of codemodder context, so moving it out to its own module so it can be used
* updated tests to create a DiffError class so we can differentiate (and catch) code errors versus codemodder diff output errors. This will help LLM codemods mostly.


clavedeluna:
@drdavella this will need to be released

andrecsilva (approved):

drdavella (approved):

drdavella:
@clavedeluna thanks! I'll create a release once this and #622 are merged.

PR #624, opened 2024-06-04
Update `openai` and test codemodder without it

clavedeluna:
~I noticed we had openai declared both as a project dependency ( but that was not defined in  `all = ...` and as a test dependency. I think just making it a project dependency this was is the correct way, but perhaps there was a different intention here?~

Fixes a bug in codemodder if openai is not installed and test in CI to prevent it. Also updated the dependency.

Can currently open dependabot PR if this is merged.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,6 +38,24 @@ jobs:
         run: python -m build .
       - name: Twine Check
         run: twine check dist/*
+  test-minimal:
+    # Test that a codemodder run doesn't require any optional dependencies.
+    name: Run codemodder list
+    runs-on: ubuntu-22.04
+    timeout-minutes: 5
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Set Up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+          cache: 'pip'
+      - name: Install Codemodder Package
+        # Only install what most users would, not optional dependencies
+        run: pip install .
+      - name: Run codemodder
+        run: codemodder --list
   test:
     name: Run pytest
     runs-on: ubuntu-22.04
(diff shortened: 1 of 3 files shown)


drdavella (commented):
@clavedeluna up til now we have included `openai` as an optional dependency so that users of OSS codemodder do not require the client if they're not using it. That's the reason for including it both in optional dependencies and in test dependencies since we do have unit tests that rely on it.

Until we have LLM codemods in this repo I might suggest keeping it that way.

clavedeluna:
@drdavella 

> @clavedeluna up til now we have included `openai` as an optional dependency so that users of OSS codemodder do not require the client if they're not using it. That's the reason for including it both in optional dependencies and in test dependencies since we do have unit tests that rely on it.
> 
> Until we have LLM codemods in this repo I might suggest keeping it that way.

I didn't say this outright in the PR description but the problem I see is that without openai getting installed as a project dependency, running codemodder fails

```
  File "/Users/dani/codemodder-python/src/codemodder/llm.py", line 47, in <module>
    def setup_llm_client() -> OpenAI | None:
                              ~~~~~~~^~~~~~
TypeError: unsupported operand type(s) for |: 'NoneType' and 'NoneType'
```

you can see this by not having openai in your environment and just running codemodder any way you like. Thankfully the issue is just a typing thing, but we weren't even aware of this because none of our environments test without openai installed. I can fix that typing issue but we'd have to test codemodder in a non-openai environment. But.... maybe just have it as a requirement?

drdavella:
@clavedeluna I still think we shouldn't require OpenAI as a dependency for this project right now. You are correct that there is an issue and thank you for identifying it. I believe the right fix is to add this at the top of the file:
```python
from __future__ import annotations
```

And maybe it's a good idea to add a separate pipeline test step that just installs the codemodder package and runs `codemodder --list` to ensure that we don't have any other optional dependencies lurking.

drdavella (approved):

PR #633, opened 2024-06-10
Implement provider registry

drdavella:
## Overview
*Add provider registry for managing access to external resources*

## Description

The provider registry is designed to manage access to external resources that may be required for particular codemods. A prototypical use case would be codemods that shell out to another codemod provider (e.g. `jscodeshift`). The provider registry enables external packages to register extensions declaring required providers. The framework is then able to manage codemod execution depending on which providers are declared by a particular codemod.

### DIFF
--- a/src/codemodder/codemods/base_codemod.py
+++ b/src/codemodder/codemods/base_codemod.py
@@ -162,7 +162,7 @@ class BaseCodemod(metaclass=ABCMeta):
         files_to_analyze: list[Path],
         rules: list[str],
     ) -> None:
-        if self.provider is not None and (
+        if self.provider and (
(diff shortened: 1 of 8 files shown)


src/codemodder/codemods/base_codemod.py
```diff
@@ -159,6 +162,15 @@ def _apply(
         files_to_analyze: list[Path],
         rules: list[str],
     ) -> None:
+        if self.provider is not None and (
```
clavedeluna:
```suggestion
        if self.provider and (
```
seems more clear

clavedeluna (approved):

PR #646, opened 2024-06-17
Support parsing/writing pyproject.toml with `poetry`

### LINKED ISSUE #634: Add dependencies to projects that are using `poetry`
For codemods that add dependencies, we should add support for `poetry`. It looks like dependencies are generally expressed in `pyproject.toml` under `[tool.poetry.dependencies]` although it also appears possible to include subgroups: https://python-poetry.org/docs/managing-dependencies/

It should be fairly simple to extend our existing dependency manager for `pyproject.toml` to also consider `poetry`.

We can limit scope here by saying that we will _not_ update lock files. It will be incumbent upon the user to do this themselves, although we anticipate providing a GitHub action that can do it automatically.

clavedeluna:
## Overview
*codemodder can now parse projects with poetry*

## Description

* poetry is a pretty common python dependency management tool
* it's declared within pyproject.toml, so I've updated that parser and writer to support poetry
* i've also handled some edge cases, such as rare case when dependencies are declared in multiple places.
* one interesting thing I had to account for is the caret `^` version declaration that's common in poetry but not parsable by `packaging.requirements`. We don't have to do much right now other than make sure we can at least parse these dependencies and store them. Our new dependencies won't use this format.

We don't currently have integration level testing for this kind of stuff so I tested this with a poetry pyproject.toml file in pygoat and it all worked as expected. `defusexml` was correctly added in the expected place.

Closes #634

### DIFF
--- a/tests/dependency_management/test_pyproject_writer.py
+++ b/tests/dependency_management/test_pyproject_writer.py
@@ -36,7 +36,7 @@ def test_update_pyproject_dependencies(tmpdir, dry_run):
     pyproject_toml.write(dedent(orig_pyproject))
 
     store = PackageStore(
-        type=FileType.REQ_TXT,
+        type=FileType.TOML,
         file=pyproject_toml,
         dependencies=set(),
         py_versions=[">=3.10.0"],
@@ -127,7 +127,7 @@ def test_add_same_dependency_only_once(tmpdir):
     pyproject_toml.write(dedent(orig_pyproject))
 
     store = PackageStore(
-        type=FileType.REQ_TXT,
+        type=FileType.TOML,
         file=pyproject_toml,
         dependencies=set(),
         py_versions=[">=3.10.0"],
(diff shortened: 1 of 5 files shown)


andrecsilva (approved):

tests/dependency_management/test_pyproject_writer.py
```diff
@@ -219,3 +219,191 @@ def test_pyproject_no_dependencies(tmpdir):
...
+    updated_pyproject = f"""\
+        [tool.poetry]
+        name = "example-project"
+        version = "0.1.0"
+        description = "An example project to demonstrate Poetry configuration."
+        authors = ["Your Name <your.email@example.com>"]
+
+        [tool.poetry.dependencies]
+        {Security.requirement.name} = "{str(Security.requirement.specifier)}"
+        {DefusedXML.requirement.name} = "{str(DefusedXML.requirement.specifier)}"
```
drdavella, line 353: Just a minor thing but sorta looks like we need an additional newline here.

drdavella (approved):
Looks really nice. Thanks for all of the tests; they really help me understand what's going on.

tests/dependency_management/test_pyproject_writer.py
clavedeluna: I saw this but it's all under the responsibility of whatever `tomlkit.dump(pyproject, f)` decides to do when it writes the file back. I'll take a quick look to see if we can do something about it

clavedeluna: a added a tiny update (last commit) and was able to add this newline!

PR #651, opened 2024-06-18
dependency type stubs

### LINKED ISSUE #635: When adding dependencies, include type stubs
For codemods that add dependencies (e.g. `defusedxml`), we should also attempt to include type stubs (e.g. `types-defusedxml`). 

We should probably attempt to detect heuristically whether `mypy` or `pyright` is being used by the project in order to avoid adding this dependency to projects that don't require it. Probably we can make use of the project analysis module to detect the presence of `mypy` or `pyright` as a preexisting dependency.

clavedeluna:
## Overview
*Add type stubs to our codemodder dependencies (those that have it) and write those stubs in pyproject.toml when poetry is used*

## Description

* the only dependencies codemodder may inject which have type stub libraries are defusedxml and flask-wtf. The latter has two potential type libraries, types-Flask, types-WTForms, but I decided not to add types-Flask because https://pypi.org/project/types-Flask/#description because  states if you use flask 2.0+ (out since 2021), the stubs are in Flask itself.
* otherwise, I used the latest types library versions
* From there, we want to be able to add these type stub libraries if defusedxml or flask-wtf are added by one of our codemods. At this time we will only add these when a project has a pyproject.toml using poetry. The logic for that is based on some reasonable heuristics. We will add the typing library in one of a few locations where it may be.

testing with pygoat looks like this
```
  - pyproject.toml
    diff:
      ---
      +++
      @@ -15,6 +15,7 @@
       requests = "^2.25.1"
       pandas = "^1.2.3"
       libcst = ">1.0"
      +defusedxml = "==0.7.1"


       [tool.poetry.dev-dependencies]
      @@ -22,4 +23,5 @@
       black = "2^0.8b1"
       flake8 = "^3.8.4"
       mypy = "~0.800"
      +types-defusedxml = "==0.7.0.20240218"

```
Closes #635

### DIFF
--- a/src/codemodder/dependency_management/pyproject_writer.py
+++ b/src/codemodder/dependency_management/pyproject_writer.py
@@ -9,6 +9,8 @@ from codemodder.dependency_management.base_dependency_writer import DependencyWr
 from codemodder.diff import create_diff_and_linenums
 from codemodder.logging import logger
 
+TYPE_CHECKER_LIBRARIES = ["mypy", "pyright"]
+
 
 def added_line_nums_strategy(lines, i):
     return lines[i]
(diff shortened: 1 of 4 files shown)


src/codemodder/dependency_management/pyproject_writer.py
```diff
@@ -9,6 +9,8 @@
 from codemodder.diff import create_diff_and_linenums
 from codemodder.logging import logger
 
+TYPE_CHECKER_LIBRARIES = ["mypy", "pyright"]
```
clavedeluna, line 12: any others we care to add?

tests/dependency_management/test_pyproject_writer.py
drdavella, line 1: Just to be a bit pedantic, but could we also have a test case where both `[tool.poetry.dependencies]` and also one of the dev/test dependencies already exist?

drdavella (changes requested):
This looks really good to me overall. I just have a request for an additional unit test case.

tests/dependency_management/test_pyproject_writer.py
clavedeluna: I'm always in favor of more testing, done!

drdavella (approved):

PR #656, opened 2024-06-20
All tests will validate codemod registration

clavedeluna:
This will make sure all codemods are always registered.


src/codemodder/codemods/test/utils.py
```diff
@@ -126,25 +141,6 @@ def run_and_assert_filepath(
         )
 
 
-class BaseSemgrepCodemodTest(BaseCodemodTest):
```
clavedeluna, line 129: the methods here must've been used at some point but they are not anymore.

andrecsilva (approved):

drdavella (approved):

PR #670, opened 2024-06-27
Update diff line number change calculation

clavedeluna:
## Overview
*calculating changed line numbers from a diff was missing logic for changes in lines that are removed*

also added small helpful assertions for testing

@drdavella this will need to be released

### DIFF
--- a/tests/test_diff.py
+++ b/tests/test_diff.py
@@ -1,0 +1,19 @@
+from collections import deque
+
+from codemodder.diff import calc_line_num_changes
+
+
+def test_calc_line_nums_remove_line():
+    lines = [
+        "--- \n",
+        "+++ \n",
+        "@@ -5,7 +5,6 @@\n",
+        "     port: 443,\n",
+        "     path: '/',\n",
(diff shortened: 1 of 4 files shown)


tests/test_diff.py
```diff
@@ -0,0 +1,19 @@
+from collections import deque
+
+from codemodder.diff import calc_line_num_changes
+
+
+def test_calc_line_nums_remove_line():
```
drdavella, line 4: This is a very nice test but it might be nice to see a few other cases:
* one where a line is changed (with both `+` and `-`)
* one where a new line is added (just `+`)
* one where multiple line changes occur within the same hunk

drdavella (changes requested):
The code looks good to me. Just a request for additional tests 🙏

tests/test_diff.py
clavedeluna: I didn't add these mostly bc it's already tested via other non-obvious ways but happy to do so!

drdavella (approved):

PR #687, opened 2024-07-01
first semgrep sarif codemod for jinja autoescape

### LINKED ISSUE #677: Codemod:  flask jinja2 autoescape Semgrep
```
    tests/samples/jinja2_autoescape.py
    ❯❱ python.flask.security.xss.audit.direct-use-of-jinja2.direct-use-of-jinja2
          Detected direct use of jinja2. If not done properly, this may bypass HTML escaping which
          opens up the application to cross-site scripting (XSS) vulnerabilities. Prefer using the
          Flask method 'render_template()' and templates with a '.html' extension in order to
          prevent XSS.
          Details: https://sg.run/RoKe

            3┆ env = Environment()
            ⋮┆----------------------------------------
            4┆ env = Environment(autoescape=False)
```

this aligns totally with `enable-jinja2-autoescape` so we should use that transformer

clavedeluna:
## Overview
*Add support for OSS semgrep codemod for jinja2 autoescape, along with the necessary code to support semgrep codemods in general*

Closes #677

### DIFF
--- a/src/codemodder/codemods/semgrep.py
+++ b/src/codemodder/codemods/semgrep.py
@@ -1,6 +1,7 @@
 import io
 import os
 import tempfile
+from functools import cache
 from pathlib import Path
 
 import yaml
@@ -8,6 +9,7 @@ import yaml
 from codemodder.codemods.base_detector import BaseDetector
 from codemodder.context import CodemodExecutionContext
 from codemodder.result import ResultSet
+from codemodder.semgrep import SemgrepResultSet
 from codemodder.semgrep import run as semgrep_run
 
 
@@ -47,3 +49,25 @@ class SemgrepRuleDetector(BaseDetector):
         yaml_files = self.get_yaml_files(codemod_id)
         with context.timer.measure("semgrep"):
             return semgrep_run(context, yaml_files, files_to_analyze)
+
+
+class SemgrepSarifFileDetector(BaseDetector):
+    def apply(
+        self,
+        codemod_id: str,
+        context: CodemodExecutionContext,
+        files_to_analyze: list[Path],
+    ) -> ResultSet:
+        del codemod_id
+        del files_to_analyze
+        return process_semgrep_findings(
+            tuple(context.tool_result_files_map.get("semgrep", ()))
+        )  # Convert list to tuple for cache hashability
+
+
+@cache
+def process_semgrep_findings(semgrep_sarif_files: tuple[str]) -> ResultSet:
+    results = SemgrepResultSet()
+    for file in semgrep_sarif_files or ():
+        results |= SemgrepResultSet.from_sarif(file)
+    return results
--- a/src/codemodder/codemods/semgrep.py
+++ b/src/codemodder/codemods/semgrep.py
@@ -1,7 +1,6 @@
 import io
 import os
 import tempfile
-from functools import cache
 from pathlib import Path
 
 import yaml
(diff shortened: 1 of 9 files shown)


src/codemodder/codemods/semgrep.py
```diff
@@ -49,25 +47,3 @@ def apply(
         yaml_files = self.get_yaml_files(codemod_id)
         with context.timer.measure("semgrep"):
             return semgrep_run(context, yaml_files, files_to_analyze)
-
-
-class SemgrepSarifFileDetector(BaseDetector):
```
drdavella: It's not a big deal but I think this is a breaking API change. Is there any chance we can keep the import in this module for backwards compability?

src/core_codemods/semgrep/semgrep_enable_jinja2_autoescape.py
```diff
@@ -0,0 +1,9 @@
+from core_codemods.enable_jinja2_autoescape import EnableJinja2Autoescape
+from core_codemods.semgrep.api import SemgrepCodemod
+
+SemgrepEnableJinja2Autoescape = SemgrepCodemod.from_core_codemod(
+    name="enable-jinja2-autoescape-semgrep",
```
drdavella, line 5: I don't think we need `semgrep` in the codemod name since its encoded by the origin.

drdavella (changes requested):
Looks great, only substantive comment is about the codemod ID.

src/codemodder/codemods/semgrep.py
clavedeluna: ahh hmm good point, haven't though about backwards compat in a while. I can just put it back, no big deal.

src/core_codemods/semgrep/semgrep_enable_jinja2_autoescape.py
clavedeluna: you're right! Since now only core codemods can be requested by name. The Sonar ones have the id in the name which hinted at that when I wrote this.

drdavella (commented):
@clavedeluna I think the new test failure is real since it's asserting on the previous name 😅

clavedeluna:
~@drdavella latest failure actually reminds us that unique codemod names are at least important in docs generation...so need to see how to update that given two same keys~

nvm fixed

clavedeluna:
@drdavella sorry for the churn here, this is the first time we have a codemod with the same name as another so it's causing more test failure and questions than I expected. The latest failure suggests something new, which is that we can't register a codemod with the same name as another [here](https://github.com/pixee/codemodder-python/blob/main/src/codemodder/registry.py#L71). It just gets skipped. We have two options here
1. `self._codemods_by_name`  should only ever be used for pixee codemods.
2. fully deprecate `self._codemods_by_name` and only use `self._codemods_by_id` everywhere

Both are easy enough changes I can do it here. LMK what you think works best in terms of customer usability / backward compat. I personally would just go for #2 and nuke the idea of codemod name if we can, to avoid future confusion. 

**edit**: I guess another option just to get this PR thru is to just add -semgrep again to the name and then ticket this for later.

drdavella:
@clavedeluna I like solution #1 for the short term if it gets us over the hump for this PR. I think we were implicitly doing that anyway (or at least that was my personal assumption). #2 might not be a bad idea to ticket for the longer term.

clavedeluna:
offline we've agreed to just add -semgrep to the name and work on the options mentioned above in a separate ticket

clavedeluna:
See https://github.com/pixee/codemodder-python/issues/697

drdavella (approved):

PR #699, opened 2024-07-08
Deprecate requesting codemods by name

### LINKED ISSUE #697: Deprecate CodemodRegistry _codemods_by_name and codemod names in general
When we had few codemods, having both codemod name and id made sense, but now that we use the same transformer for various SAST tools, calling a codemod by a name and not by an id doesn't disambiguate. To prevent future confusion, we should go ahead and deprecate codemod names and storing them in the registry. Only codemod ids should be used to invoke codemods

This may mean we still retain the `name`field for the codemod metadata, but only as a simplified way to instantiate a codemod. It can only ever be accessed by it's full id, specifying tool provider.

clavedeluna:
## Overview
*Codemods can now only be requested by id*

We still have the ability to initialize a codemod by a name, but it's now know that this may not be a unique name since multiple tools could have the same codemod name. So codemods can now only be requested by id.

I changed all important locations, while some locations still use `codemod.name`. I also made a wrapper called `_internal_name` to indicate that the places we're using that it's for the internal semgrep call. 

Closes #697

Future work:  I can remove the `-semgrep` suffix in the first semgrep codemod I recently added since it's so new, but do we want to also go back and rename all the sonar codemods that have `-id` suffix?


andrecsilva (approved):

drdavella (approved):
Thank you @clavedeluna, looks really good.

PR #705, opened 2024-07-09
Semgrep defused xml codemod

### LINKED ISSUE #695: Codemod: use-defused-xml-parse Semgrep
https://semgrep.dev/playground/r/python.lang.security.use-defused-xml-parse.use-defused-xml-parse?editorMode=advanced

we already have a transformer for this so it's a matter of adding semgrep as the tool for a new codemod

clavedeluna:
Closes #695

See question on that issue for something related.

### DIFF
--- a/src/codemodder/codemods/import_modifier_codemod.py
+++ b/src/codemodder/codemods/import_modifier_codemod.py
@@ -52,9 +52,6 @@ class ImportModifierCodemod(SimpleCodemod, metaclass=ABCMeta):
         pass
 
     def transform_module_impl(self, tree: cst.Module) -> cst.Module:
-        if not self.node_is_selected(tree):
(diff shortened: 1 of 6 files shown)


src/codemodder/codemods/import_modifier_codemod.py
```diff
@@ -52,9 +52,6 @@ def mapping(self) -> Mapping[str, str]:
         pass
 
     def transform_module_impl(self, tree: cst.Module) -> cst.Module:
-        if not self.node_is_selected(tree):
```
clavedeluna, line 55: this would prevent any tool results from moving forward. doesn't seem like it's currently in use via testing

andrecsilva (approved):

PR #711, opened 2024-07-11
New RSA key size transformer and semgrep codemod

### LINKED ISSUE #522: Suggestion: Add warning messages in LibcstTransformerPipeline
The code where I think it might be worth adding a warning is in the `apply` method of `LibcstTransformerPipeline`.

```python
        try:
            with file_context.timer.measure("transform"):
                for transformer in self.transformers:
                    tree = transformer.transform(tree, results, file_context)
        except Exception:
            file_context.add_failure(file_path)
            logger.exception("error transforming file %s", file_path)
            return None

        if not file_context.codemod_changes:
            return None

        if not (diff := create_diff_from_tree(source_tree, tree)):
            return None
```

In the case where `tree` has changes but the codemod author has not added to the `file_context.codemod_changes` array the changes they are expecting to be made will be discarded. I would say that when both conditions are `False` it's likely the codemod author will expect nothing to happen but in cases where only one is `False` it would be useful for there to be a warning printed as I think any time there is a element in `file_context.codemod_changes` or there is a tree diff the codemod author intended for something to happen.

### LINKED ISSUE #692: Codemod: insufficient-rsa-key-size Semgrep
https://semgrep.dev/playground/r/python.cryptography.security.insufficient-rsa-key-size.insufficient-rsa-key-size?editorMode=advanced

We don't have a transformer for this right now but the fix is very straight forward.

clavedeluna:
We Didn't have a transformer for this so it was easy enough to implement one.

Closes #692
Closes #522 as I did find it helpful for debugging

### DIFF
--- a/src/codemodder/codemods/libcst_transformer.py
+++ b/src/codemodder/codemods/libcst_transformer.py
@@ -278,9 +278,11 @@ class LibcstTransformerPipeline(BaseTransformerPipeline):
             return None
 
         if not file_context.codemod_changes:
+            logger.debug("No changes produced for %s", file_path)
             return None
 
         if not (diff := create_diff_from_tree(source_tree, tree)):
+            logger.debug("No code diff produced for %s", file_path)
             return None
 
         change_set = ChangeSet(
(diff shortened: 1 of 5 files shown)


drdavella (approved):
There's nothing wrong with this codemod per se, but it's going to be hard to apply since in general this seems like a potentially breaking change. I'm happy to accept it as-is but we might hold off on introducing this into our downstream catalog.

PR #729, opened 2024-07-22
Regex Transformer

clavedeluna:
This PR adds a new transformer so that any codemod which would simply like to replace some piece of text via a pattern with some other text can easily do so. For example:
```
...
transformer=RegexTransformerPipeline(
        pattern=r'debug\s*=\s*["\']true["\']',
        replacement='debug="false"',
        change_description="...",
    )
...

### DIFF
--- a/src/codemodder/codemods/regex_transformer.py
+++ b/src/codemodder/codemods/regex_transformer.py
@@ -11,6 +11,9 @@ from codemodder.result import Result
 
 
 class RegexTransformerPipeline(BaseTransformerPipeline):
+    pattern: Pattern
+    replacement: str
+    change_description: str
 
     def __init__(self, pattern: Pattern, replacement: str, change_description: str):
         super().__init__()
--- a/src/codemodder/codemods/regex_transformer.py
+++ b/src/codemodder/codemods/regex_transformer.py
@@ -1,0 +1,60 @@
+import re
+from typing import Pattern
+
+from codemodder.codemods.base_transformer import BaseTransformerPipeline
+from codemodder.codetf import Change, ChangeSet
+from codemodder.context import CodemodExecutionContext
+from codemodder.diff import create_diff
+from codemodder.file_context import FileContext
+from codemodder.logging import logger
+from codemodder.result import Result
+
+
+class RegexTransformerPipeline(BaseTransformerPipeline):
+
+    def __init__(self, pattern: Pattern, replacement: str, change_description: str):
+        super().__init__()
+        self.pattern = pattern
+        self.replacement = replacement
+        self.change_description = change_description
+
+    def apply(
+        self,
+        context: CodemodExecutionContext,
+        file_context: FileContext,
+        results: list[Result] | None,
+    ) -> ChangeSet | None:
+
+        changes = []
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/codemodder/codemods/regex_transformer.py [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+from codemodder.context import CodemodExecutionContext
+from codemodder.diff import create_diff
+from codemodder.file_context import FileContext
+from codemodder.logging import logger
+from codemodder.result import Result
+
+
+class RegexTransformerPipeline(BaseTransformerPipeline):
+
+    def __init__(self, pattern: Pattern, replacement: str, change_description: str):
```
drdavella, line 18: We don't need to overengineer this right now but I can imagine scenarios where we'd like a more sophisticated callback-based replacement. `re.sub` already supports `repl` as a string or a function, so maybe we want to capture that in the type signature?

src/codemodder/codemods/regex_transformer.py
```diff
@@ -0,0 +1,60 @@
...
+        super().__init__()
+        self.pattern = pattern
```
drdavella, line 20: It would be a good idea to declare types for each of these attributes at the top of the class.

src/codemodder/codemods/regex_transformer.py
```diff
@@ -0,0 +1,60 @@
...
+        self,
+        context: CodemodExecutionContext,
+        file_context: FileContext,
+        results: list[Result] | None,
+    ) -> ChangeSet | None:
+
+        changes = []
+        updated_lines = []
+
+        with open(file_context.file_path, "r", encoding="utf-8") as f:
```
drdavella, line 34: `file_context.file_path` is a `Path` so you can use convenience methods like `read_text`.

src/codemodder/codemods/regex_transformer.py
```diff
@@ -0,0 +1,60 @@
...
+                )
+
+        if not changes:
+            logger.debug("No changes produced for %s", file_context.file_path)
+            return None
+
+        diff = create_diff(original_lines, updated_lines)
+
+        if not context.dry_run:
+            with open(file_context.file_path, "w+", encoding="utf-8") as original:
```
drdavella, line 56: `file_context.file_path.write_text`

drdavella (approved):
Looks very nice overall. Approving with some nitpicks/comments.

src/codemodder/codemods/regex_transformer.py
clavedeluna: in this case we do want readlines because we need a line by line check so we can create the lineno change. below I can change

src/codemodder/codemods/regex_transformer.py
clavedeluna: hm yeah ok, so technically I could `.wirte_text("\n".join(updated_lines)` but I kinda prefer the current way so we let `writelines` handles the newline stuff.

Plus I have less incentive to change it here since the libcst transformer also does `open ... write`.    see `update_code`. maybe we can change it globally later on?

Issue #738, opened 2024-07-24
remove assertion in pytest-raises codemod should report change for each line changed


clavedeluna:
See `# TODO: need to report change for each line changed`

PR #739, opened 2024-07-24
report semgrep findings and fix findings assertions

clavedeluna:
I initially thought asserting findings is not None was sufficient, but in cases when the result of matching findings is `[]` we also want to be aware.

I updated some codemods where it was easy to do the matching, but created https://github.com/pixee/codemodder-python/issues/738 for that codemod

This fix also brought to light that we weren't reporting semgrep findings so fixed that too

### DIFF
--- a/src/codemodder/result.py
+++ b/src/codemodder/result.py
@@ -9,7 +9,7 @@ import libcst as cst
 from libcst._position import CodeRange
 from typing_extensions import Self
 
-from codemodder.codetf import Finding
+from codemodder.codetf import Finding, Rule
 
 from .utils.abc_dataclass import ABCDataclass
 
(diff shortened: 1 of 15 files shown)


src/codemodder/result.py
```diff
@@ -68,18 +68,39 @@ def match_location(self, pos: CodeRange, node: cst.CSTNode) -> bool:
...
             locations=cls.extract_locations(sarif_result),
             codeflows=cls.extract_code_flows(sarif_result),
             related_locations=cls.extract_related_locations(sarif_result),
+            finding_id=rule_id,
+            finding=Finding(
+                id=rule_id,
+                rule=Rule(
+                    id=rule_id,
+                    name=rule_id,
+                    url=semgrep_url_from_id(rule_id),
```
drdavella: This is specific to Semgrep but it looks like this is being added to a more general `SarifResult` class?

drdavella (changes requested):

src/codemodder/result.py
clavedeluna: yeah do you think this is wrong? while it's a generic sarif result, wouldn't all sarif results have to have findings too?

andrecsilva (approved):

src/codemodder/result.py
drdavella: My concern is specifically about the call to `semgrep_url_from_id`, which only applies to Semgrep results, whereas this class appears to be used for multiple SARIF-based tools.

clavedeluna: wooops Im an idiot, of course! Good review Dan, will update

drdavella (approved):

PR #750, opened 2024-07-26
Use import call modifier pattern for url-sandbox

### LINKED ISSUE #746: `url-sandbox` does not properly handle multiple names using `from` import
See the test case added in #745. We need to determine whether this problem is more general or just limited to this particular codemod.

drdavella:
## Overview
*Refactor `url-sandbox` codemod to use import call modifier pattern*

## Description

* This refactoring allows us to reuse a lot of code that already existed and to get generalization 
* For this codemod we need to keep the Semgrep detector in order to avoid false positives from hard-coded strings
* This fixes #746


clavedeluna (approved):
👏 👏 👏

PR #758, opened 2024-07-30
Semgrep Nan Injection codemod

### LINKED ISSUE #691: Codemod: nan-injection Semgrep
https://semgrep.dev/playground/r/python.django.security.nan-injection.nan-injection?editorMode=advanced

We don't have a transformer fix for this and would need to do a bit of investigation to decide what the right fix is.

clavedeluna:
This codemod replaces typecast call at a line reported by semgrep (which is a sink of a request source)
`x = float(uuid)`
and wraps it in an if/else statement to check if the value could be nan.

Notice that this new transformer is specific to semgrep results. It wouldn't work stand-alone. I also decided to not handle other edge cases since this is likely a codemod that won't be used much. If you think of something absolutely necessary we do need, I can add it.

Closes #691

### DIFF
--- a/src/core_codemods/semgrep/semgrep_nan_injection.py
+++ b/src/core_codemods/semgrep/semgrep_nan_injection.py
@@ -44,16 +44,13 @@ class NanInjectionTransformer(LibcstResultTransformer):
         original_node: cst.SimpleStatementLine,
         updated_node: cst.SimpleStatementLine,
     ):
-        # We assume the call node's argument is a variable, otherwise semgrep
-        # wouldn't have created a finding.
-        try:
-            var = node.args[0].value.value
-        except Exception:
-            return updated_node
+        var_name = self._get_var_in_call(node)
+        if not var_name:
+            return original_node
 
         code = dedent(
             f"""\
-        if {var}.lower() == "nan":
+        if {var_name}.lower() == "nan":
             raise ValueError
         else:
             {self.code(original_node).strip()}
@@ -62,6 +59,15 @@ class NanInjectionTransformer(LibcstResultTransformer):
         self._report_new_lines(original_node)
         new_statement = cst.parse_statement(code)
         return new_statement.with_changes(leading_lines=updated_node.leading_lines)
+
+    def _get_var_in_call(self, node: cst):
+        match node.args[0].value:
+            case cst.Name():
+                # float(var)
+                return node.args[0].value.value
+            case cst.BinaryOperation():
+                # bool(float(var)), complex(float(var)), etc
+                return node.args[0].value.left.args[0].value.value
 
     def _report_new_lines(self, original_node: cst.SimpleStatementLine):
         self.report_change(original_node)
--- a/src/core_codemods/semgrep/semgrep_nan_injection.py
+++ b/src/core_codemods/semgrep/semgrep_nan_injection.py
@@ -44,12 +44,12 @@ class NanInjectionTransformer(LibcstResultTransformer):
         original_node: cst.SimpleStatementLine,
         updated_node: cst.SimpleStatementLine,
     ):
-        if not (var_name := self._get_var_in_call(node)):
+        if not (target_node := self._get_target_in_call(node)):
             return original_node
 
         code = dedent(
             f"""\
-        if {var_name}.lower() == "nan":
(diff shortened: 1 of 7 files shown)


src/core_codemods/semgrep/semgrep_nan_injection.py
```diff
@@ -0,0 +1,122 @@
...
+from codemodder.codemods.libcst_transformer import (
+    LibcstResultTransformer,
+    LibcstTransformerPipeline,
+)
+from codemodder.codemods.semgrep import SemgrepSarifFileDetector
+from core_codemods.semgrep.api import SemgrepCodemod, semgrep_url_from_id
+
+
+class NanInjectionTransformer(LibcstResultTransformer):
+    change_description = "Wrap typecast in if statement to check for `nan`"
```
drdavella: I'm not sure "wrap" is the right verb here since that has a more specific connotation (imo).
```suggestion
    change_description = "Add validation to untrusted numerical input to disallow `nan`"
```

tests/codemods/semgrep/test_semgrep_nan_injection.py
```diff
@@ -0,0 +1,71 @@
+import json
+
+from codemodder.codemods.test import BaseSASTCodemodTest
+from core_codemods.semgrep.semgrep_nan_injection import SemgrepNanInjection
+
+
+class TestSemgrepNanInjection(BaseSASTCodemodTest):
```
drdavella, line 11: I think we need some additional cases where the `float(potential_nan)` is used in a context other than simple assignment.

I also think we should add test cases for the other type casts that are identified by this Semgrep rule.

drdavella (dismissed):
I agree that this is a pretty niche codemod. I do think we need a bit of additional testing if we want to enable it, just to make sure that we don't introduce weird/bad code.

tests/codemods/semgrep/test_semgrep_nan_injection.py
clavedeluna: I've added more testing for cases that semgrep triggers the rule.

src/core_codemods/semgrep/semgrep_nan_injection.py
```diff
@@ -63,6 +60,15 @@ def replace_with_if_else(
...
 
+    def _get_var_in_call(self, node: cst):
+        match node.args[0].value:
+            case cst.Name():
+                # float(var)
+                return node.args[0].value.value
+            case cst.BinaryOperation():
+                # bool(float(var)), complex(float(var)), etc
+                return node.args[0].value.left.args[0].value.value
+
```
andrecsilva: This looks weird.
The type hint for `node` should be `cst.Call`.
The case for `cst.BinaryOperation()` does not match the the intention, at least judging from the comment above. Do you mean `cst.Call`? 
If that's the case, you should account for deeper nestings (e.g. `float(complex(int(...)))` ). Does the semgrep rule always guarantee a variable name there? It could be an expression.

clavedeluna: whoops, updated the type hint.
~BinaryOperation because when nested cst really does interpret it as a binaryoperation, for all calls of bool, int, float, and complex. I checked them all.~ this is incorrect

I'll check the last part you mention.

clavedeluna: so checking in the [playground](https://semgrep.dev/playground/r/python.django.security.nan-injection.nan-injection?editorMode=advanced)
1. looks like infinite nesting is a possibility ... `complex(bool(float(tid))` will match. But this only matches as long as all funcs are one of complex, bool, float. Note the  `pattern-sanitizers`, that means any other function will sanitize this away.
2. I do need to handle the non-var case for `float(request.POST.get("tid"))`, which is annoying it  wasn't a semgrep example.

~I think the reasonable thing to do here is to not implement #1, just test and not to it, but to do #2. Any other pattern won't be fixed for now.~ I'll try to do both :)

clavedeluna:
alright, so I've updated and improved this codemod to handle as many cases as I could get to fire in semgrep. There are probably more, but in those cases we won't do anything.

andrecsilva (approved):

PR #759, opened 2024-07-30
Use 1-based line numbers for RegexTransformer changes

drdavella:


nahsra (approved):

src/codemodder/codemods/regex_transformer.py
```diff
@@ -27,6 +29,7 @@ def apply(
         file_context: FileContext,
         results: list[Result] | None,
     ) -> ChangeSet | None:
+        del results
```
gilday, line 32: What's the significance of deleting this reference? Is it to communicate that this parameter is intentionally unused and only exists to satisfy the API requirements of the caller?

gilday (approved):

src/codemodder/codemods/regex_transformer.py
drdavella: Yes that's right. It makes explicit to the to the type checker that this parameter is intentionally unused.

PR #765, opened 2024-08-01
Transformers should read and write bytes

clavedeluna:
## Overview
*Handling files as bytes and then decoding / encoding will help support windows style newlines*

Also since `file_path` is a `Path` object we don't need to `with open....`. That also required some minor testing changes.

**this should be released as a minor version**


src/codemodder/codemodder.py
```diff
@@ -23,14 +23,6 @@
 from codemodder.semgrep import run as run_semgrep
 
 
-def update_code(file_path, new_code):
```
clavedeluna, line 26: this was unused, it's defined below in libcst transformer

andrecsilva (approved):

PR #780, opened 2024-08-07
Fix default path include/exclude behavior; update codemod hierarchy

### LINKED ISSUE #761: Default path includes/excludes should not be used for remediation codemods
See the spec update: https://github.com/pixee/codemodder-specs/pull/37

We should not make assumptions about default path inclusion/exclusion behavior for codemods that remediate tool results. These tools have their own defaults and/or configuration so codemods should simply respect whatever the tool reports. This behavior can still be overridden with `--path-include` or `--path-exclude`: the critical point is that we should not impose _defaults_ on these types of codemods.

drdavella:
## Overview
*Update default path include/exclude behavior; update codemod hierarchy*

## Description
* Remediation codemods should not use default exclude paths: this would cause us to potentially ignore paths that are explicitly called out in tool findings
  * Find and fix codemods, on the other hand, should retain reasonable defaults wrt files to be excluded from analysis (e.g. `.git`, `*.pyc`, etc.)
  * In both cases, the include and exclude paths should remain configurable by the user
* This PR refactors the base codemod hierarchy into two branches: one for find and fix, and one for remediation
* Each respective base class defines appropriate behavior for path inclusion/exclusion
* Both classes respect explicit settings
* Much of the path includes/excludes state has been moved to `CodemodExecutionContext` where it can be easily cached and reused
* This is a big, breaking change. We will likely release v2.0 as a result
* This closes #761


ci_tests/test_pygoat_findings.py
```diff
@@ -6,7 +6,9 @@
     "pixee:python/add-requests-timeouts",
     "pixee:python/secure-random",
     "pixee:python/sandbox-process-creation",
+    "pixee:python/subprocess-shell-false",
```
drdavella, line 9: This one is a result of adding `--dry-run` (which I was using locally while testing).

ci_tests/test_pygoat_findings.py
```diff
@@ -6,7 +6,9 @@
...
     "pixee:python/django-session-cookie-secure-off",
+    "pixee:python/django-model-without-dunder-str",
```
drdavella, line 11: I'm not really sure how to account for this change 😅

src/codemodder/codemodder.py
```diff
@@ -72,38 +71,22 @@ def log_report(context, argv, elapsed_ms, files_to_analyze):
...
         return
 
-    semgrep_finding_ids = semgrep_results.all_rule_ids()
-
     # run codemods one at a time making sure to respect the given sequence
     for codemod in codemods_to_run:
         # NOTE: this may be used as a progress indicator by upstream tools
         logger.info("running codemod %s", codemod.id)
-
-        if isinstance(codemod.detector, SemgrepRuleDetector):
```
drdavella, line 95: This filtering logic is now handled by the codemod itself.

ci_tests/test_pygoat_findings.py
clavedeluna: this is really surprising... I can see pygoat django Model classes that would merit this codemod, but why this wasn't getting triggered before 🤷

clavedeluna (approved):
the changes and tests all look as expected, awesome work!

PR #822, opened 2024-09-05
CodeTF: change descriptions are optional but must not be empty

drdavella:


clavedeluna (approved):
a little confusing since "" is often the right sentinel value for empty str but I Get it

PR #831, opened 2024-09-12
New Semgrep codemod: sandbox process creation

drdavella:
Remediates Semgrep's [dangerous-system-call](https://semgrep.dev/r?q=python.lang.security.dangerous-system-call.dangerous-system-call) using the core codemod `sandbox-process-creation`.


andrecsilva (approved):
Seems to lacking an integration test, otherwise LGTM

PR #872, opened 2024-10-09
Add Azure Llama client support

### LINKED ISSUE #871: Add client for Azure-hosted Llama models
We would like to support codemods that use Llama models hosted in Azure. 

We have proposed a spec update here that introduces two new environment variables: https://github.com/pixee/codemodder-specs/pull/39
```
CODEMODDER_AZURE_LLAMA_API_KEY=<KEY>
CODEMODDER_AZURE_LLAMA_ENDPOINT=<ENDPOINT>
```

If only one or the other variable is present, we should raise an exception.

If both of these variables are present, we need to create a new llama client using the [Azure AI Inference package](https://learn.microsoft.com/en-us/azure/ai-studio/how-to/deploy-models-llama?tabs=python-llama-3-2%2Cmeta-llama-3-1&pivots=programming-language-python#the-inference-package-installed).

Use of the client looks something like this:
```python
# pip install azure-ai-inference
import os
from azure.ai.inference import ChatCompletionsClient
from azure.core.credentials import AzureKeyCredential

api_key = os.getenv("<API key>", '')
if not api_key:
  raise Exception("A key should be provided to invoke the endpoint")

client = ChatCompletionsClient(
    endpoint='<endpoint URL>',
    credential=AzureKeyCredential(api_key)
)
```

The behavior of this client is completely orthogonal to the OpenAI/Azure OpenAI client. One or the other or both can be configured at the same time. The client should live on the `CodemodExecutionContext`. Implementing this may lead to some breaking API changes.

clavedeluna:
## Overview
*Codemodder can run codemods with Azure LLama models*

## Description

* This is a breaking change because I thought it best to change `context.llm_client > context.openai_llm_client` for clarity
* now there can be both clients enabled.
* it's not necessary to add the model to the ModelRegistry, since the model used is encoded in whatever endpoint is used in `CODEMODDER_AZURE_LLAMA_ENDPOINT`
I tested this out with
```
        from azure.ai.inference.models import SystemMessage, UserMessage

        response = self.azure_llama_llm_client.complete(
            messages=[
                SystemMessage(content="You are a helpful assistant."),
                UserMessage(content="How many feet are in a mile?"),
            ]
        )
```

and got the expected response
```
(Pdb) print(response.choices[0].message.content)
There are 5,280 feet in a mile.
```

Close #871

### DIFF
--- a/src/codemodder/llm.py
+++ b/src/codemodder/llm.py
@@ -26,6 +26,7 @@ from codemodder.logging import logger
 __all__ = [
     "MODELS",
     "setup_openai_llm_client",
+    "setup_azure_llama_llm_client",
     "MisconfiguredAIClient",
 ]
 
--- a/src/codemodder/llm.py
+++ b/src/codemodder/llm.py
@@ -17,7 +17,7 @@ from codemodder.logging import logger
 
 __all__ = [
     "MODELS",
-    "setup_llm_client",
+    "setup_openai_llm_client",
     "MisconfiguredAIClient",
 ]
 
@@ -46,7 +46,7 @@ class ModelRegistry(dict):
 MODELS = ModelRegistry(models)
 
 
-def setup_llm_client() -> OpenAI | None:
+def setup_openai_llm_client() -> OpenAI | None:
     if not AzureOpenAI:
         logger.info("Azure OpenAI API client not available")
         return None
--- a/src/codemodder/llm.py
+++ b/src/codemodder/llm.py
@@ -9,9 +9,17 @@ except ImportError:
     OpenAI = None
     AzureOpenAI = None
 
+try:
+    from azure.ai.inference import ChatCompletionsClient
+    from azure.core.credentials import AzureKeyCredential
+except ImportError:
+    ChatCompletionsClient = None
+    AzureKeyCredential = None
 
 if TYPE_CHECKING:
     from openai import OpenAI
+    from azure.ai.inference import ChatCompletionsClient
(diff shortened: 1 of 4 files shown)


src/codemodder/llm.py
```diff
@@ -81,5 +91,27 @@ def setup_llm_client() -> OpenAI | None:
...
 
+def setup_azure_llama_llm_client() -> ChatCompletionsClient | None:
+    """Configure the Azure Llama LLM client."""
+    if not ChatCompletionsClient:
+        logger.info("Azure API client not available")
+        return None
+
+    azure_llama_key = os.getenv("CODEMODDER_AZURE_LLAMA_API_KEY")
+    azure_llama_endpoint = os.getenv("CODEMODDER_AZURE_LLAMA_ENDPOINT")
+    if bool(azure_llama_key) ^ bool(azure_llama_endpoint):
```
drdavella, line 102: This is my fault because I did this originally but `!=` is definitely a lot clearer here 😅

drdavella (approved):
LGTM but I'm not sure what's up with the failing tests.

drdavella:
> This is a breaking change because I thought it best to change context.llm_client > context.openai_llm_client for clarity

I agree. I don't think there's any need for deprecation periods at this point, we should just remember to bump major versions.

src/codemodder/llm.py
clavedeluna: yeah I kinda agree but I also got very used to reading this and it's nice :)

PR #881, opened 2024-10-16
Gather results by rules

drdavella:
## Description

* For multi-file codemods we need a finding-oriented view rather than a file-oriented view of results. This is because fixing a single finding could potentially require changes to multiple files


nahsra (approved):
lgtm

PR #925, opened 2024-11-21
Remove restriction on duplicate sarif tools

drdavella:
This effectively reverts #742. It was based on an assumption that does not hold in a polyglot environment for tools that may scan each language separately (e.g. CodeQL).


nahsra (approved):
I'm approving without understanding why this fence was put up in the first place.

PR #927, opened 2024-11-21
Avoid duplicate findings from multiple sarif files

drdavella:
## Overview
*Avoid duplicating preexisting findings when processing multiple (potentially duplicate) SARIF files for a given tool*

## Description

* The basic solution is to use `IndexedSet` to enforce uniqueness when combining `ResultSet`s
  * `IndexedSet` preserves ordering 👍 
* In order to make this possible, the `Result` type needs to be hashable. This required the following changes
  * Using `frozen` datatypes
  * Using `frozen` pydantic classes
  * Replacing all instances of `list` (not hashable) in these types with `tuple` (hashable)
* The type changes introduce a few downstream breaking API changes, but they should be easy to absorb


dunningdan (approved):

pyproject.toml
```diff
@@ -11,6 +11,7 @@ readme = "README.md"
 license = {file = "LICENSE"}
 description = "A pluggable framework for building codemods in Python"
 dependencies = [
+    "boltons~=21.0.0",
```
clavedeluna, line 14: TIL! Will become my new favorite utility, I"m sure

clavedeluna (approved):

PR #935, opened 2024-12-02
implement finding msg for codeql

clavedeluna:
We need to gather as much finding metadata as possible. Implemented `finding_msg` for codeql, remaining tools as TODO

### DIFF
--- a/src/codemodder/codeql.py
+++ b/src/codemodder/codeql.py
@@ -45,7 +45,7 @@ class CodeQLResult(SarifResult):
     ) -> Self:
         rule_id = cls.extract_rule_id(sarif_result, sarif_run, truncate_rule_id)
         text_for_rule = get_text_for_rule(rule_id, sarif_run)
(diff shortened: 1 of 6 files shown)


src/codemodder/codeql.py
```diff
@@ -80,3 +80,12 @@ def from_sarif(cls, sarif_file: str | Path, truncate_rule_id: bool = False) -> S
                     )
                     result_set.add_result(codeql_result)
         return result_set
+
+
+# TODO: cache, make hashable
```
clavedeluna, line 85: I don't have a ton of bandwidth so left it as a todo, since we're passing dict of lists and some other unhashable data here, it's not super fast to do, but we should do it at some point

drdavella (commented):
@clavedeluna I believe the need for this metadata exists not at the _finding_ level but at the _rule_ level. It's possible that we should expose this at the level of `SarifResultSet` as a mapping between rule IDs and metadata.

PR #940, opened 2024-12-04
add rule metadata to results set

clavedeluna:
## Overview
*One sentence, high-level explanation of WHY changes are introduced and/or business need*

## Description

* What/WHY/how these changes are needed

## Additional Details

### DIFF
--- a/src/codemodder/result.py
+++ b/src/codemodder/result.py
@@ -11,7 +11,8 @@ from boltons.setutils import IndexedSet
 from libcst._position import CodeRange
 from typing_extensions import Self
 
-from codemodder.codetf import Finding, Rule
+from codemodder.codetf import Finding
+from codemodder.codetf import Rule as CodetfRule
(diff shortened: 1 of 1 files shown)


src/codemodder/result.py
```diff
@@ -176,18 +177,33 @@ def fuzzy_column_match(pos: CodeRange, location: Location) -> bool:
     )
 
 
-class ResultSet(dict[str, dict[Path, list[Result]]]):
-    results_for_rule: dict[str, list[Result]]
+class Rule(CodetfRule):
```
drdavella, line 180: @clavedeluna what if instead of being prescriptive about the specific metadata fields here, we instead add a field with an abstract type that conforms to some API like `get_metadata_summary() -> str` and allow each tool to define the way they want to present it?

I also don't want to overengineer this without having a lot of examples to work with.

PR #962, opened 2025-01-02
Fix sonar issues and hotspots parsing

drdavella:
## Overview
*Fixes a relatively small bug related to parsing empty sonar issues files*

## Description

* This problem was already handled safely but I think it's best to clarify/simplify the logic here
* As far as I can tell, this would only occur with an empty and/or malformed Sonar issues input
* Also cleaned up a few noisy deprecation warnings

### DIFF
--- a/src/codemodder/codetf.py
+++ b/src/codemodder/codetf.py
@@ -12,7 +12,7 @@ from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, Optional
 
-from pydantic import BaseModel, model_validator
+from pydantic import BaseModel, ConfigDict, model_validator
 
 from codemodder import __version__
 from codemodder.logging import logger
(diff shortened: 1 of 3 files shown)


src/codemodder/codemods/base_codemod.py
```diff
@@ -7,7 +7,7 @@
 from dataclasses import dataclass, field
 from enum import Enum
 from functools import cached_property
-from importlib.abc import Traversable
+from importlib.resources.abc import Traversable
```
drdavella: Fix deprecation warning

src/codemodder/codetf.py
```diff
@@ -138,16 +138,14 @@ class Rule(BaseModel):
     name: str
     url: Optional[str] = None
 
-    class Config:
-        frozen = True
+    model_config = ConfigDict(frozen=True)
```
drdavella, line 141: Fix deprecation warning

nahsra (approved):
lgtm

src/codemodder/codemods/base_codemod.py
drdavella: Unfortunately it looks like we need to wait until we drop 3.10 support for this one (but we can do that pretty much any time we want).

PR #991, opened 2025-02-06
Report LLM token usage

clavedeluna:
Any codemod which may use an LLM and spends prompt and completion tokens will now be reported, per each codemod and at the end of codemodder run.

Looks something like this (redacted)

```
Codemod REDACTED
	completion_tokens = 2309
	prompt_tokens = 13742
...
running codemod SOME OTHER CODEMOD
...
...
[report]
scanned: 8 files
failed: 0 files (0 unique)
changed: 8 files (4 unique)
report file: here.txt
All token usage
	completion_tokens = 5176
	prompt_tokens = 40960
```

### DIFF
--- a/src/codemodder/llm.py
+++ b/src/codemodder/llm.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
 import os
+from dataclasses import dataclass
 from typing import TYPE_CHECKING
+
+from typing_extensions import Self
 
 try:
     from openai import AzureOpenAI, OpenAI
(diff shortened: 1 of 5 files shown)


drdavella (approved):

src/codemodder/llm.py
```diff
@@ -58,8 +63,8 @@ def __getattr__(self, name):
 def setup_openai_llm_client() -> OpenAI | None:
     """Configure either the Azure OpenAI LLM client or the OpenAI client, in that order."""
     if not AzureOpenAI:
-        logger.info("Azure OpenAI API client not available")
         return None
+        logger.info("Azure OpenAI API client not available")
```
andrecsilva: I'm guessing this is a mistake, this log is unreacheable and will never execute.

andrecsilva (approved):

src/codemodder/llm.py
clavedeluna: wooops yes very much

PR #1012, opened 2025-03-03
Working draft for CodeTFv3 data model

drdavella:
## Overview
*Introduces new fix-oriented CodeTF data model as implementation for v3*


src/codemodder/codetf/v3/codetf.py
```diff
@@ -0,0 +1,181 @@
...
+
+class Strategy(Enum):
+    ai = "ai"
+    hybrid = "hybrid"
+    deterministic = "deterministic"
+
+
+class AIMetadata(BaseModel):
+    provider: Optional[str] = None
+    models: Optional[list[str]] = None
```
drdavella, line 124: Multiple models may be involved in fix generation

src/codemodder/codetf/v3/codetf.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class FixResult(BaseModel):
+    """Result corresponding to a single finding"""
+
+    finding: Finding
+    fixStatus: FixStatus
+    changeSets: list[ChangeSet]
+    fixMetadata: Optional[FixMetadata] = None
+    fixQuality: Optional[FixQuality] = None
```
drdavella, line 165: Fix quality evaluation belongs to the entire result, not just to a single file/diff.

src/codemodder/codetf/v3/codetf.py
```diff
@@ -0,0 +1,181 @@
...
+class FixQuality(BaseModel):
+    safetyRating: Rating
+    effectivenessRating: Rating
+    cleanlinessRating: Rating
+
+
+class FixResult(BaseModel):
+    """Result corresponding to a single finding"""
+
+    finding: Finding
```
drdavella, line 161: There is now a strict 1:1 mapping between finding and fix.

src/codemodder/codetf/v3/codetf.py
```diff
@@ -0,0 +1,181 @@
...
+    fixStatus: FixStatus
+    changeSets: list[ChangeSet]
+    fixMetadata: Optional[FixMetadata] = None
+    fixQuality: Optional[FixQuality] = None
+    # A description of the reasoning process that led to the fix
+    reasoningSteps: Optional[list[str]] = None
```
drdavella, line 167: Potentially each entry here should be something richer than a flat string: we might want to incorporate steps that represent user input/feedback.

src/codemodder/codetf/v3/codetf.py
```diff
@@ -0,0 +1,181 @@
...
+            if not self.changeSets:
+                raise ValueError("changeSets must be provided for fixed results")
+            if not self.fixMetadata:
+                raise ValueError("fixMetadata must be provided for fixed results")
+        return self
+
+
+class CodeTF(CodeTFWriter, BaseModel):
+    run: Run
+    results: list[FixResult]
```
drdavella, line 181: My expectation is that each finding in the input will have a corresponding result. I don't believe that's enforceable in the spec, and it probably should not be since we want to allow for the possibility of streaming subsets of results.

src/codemodder/codetf/v3/codetf.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class DiffSide(CaseInsensitiveEnum):
+    LEFT = "left"
+    RIGHT = "right"
+
+
+class PackageAction(BaseModel):
+    action: Action
+    result: PackageResult
```
drdavella, line 76: I'm not sure I ever fully understood the rationale for the `result` field: it might be worth discussing whether it's still necessary.

clavedeluna (approved):

src/codemodder/codetf/v3/codetf.py
nahsra: I could quibble that we have no idea what form that would take, and whether we should design for anything beyond a `str` now... but if you have short term thoughts that this is a promising data type, I'm okay with it.

src/codemodder/codetf/v3/codetf.py
nahsra: > I don't believe that's enforceable in the spec, and it probably should not be since we want to allow for the possibility of streaming subsets of results.

I'm not sure why these requirements are at odds. Aren't we only giving it 1 input as well? We had discussed as an input a single SARIF, and a single result ID (or something.) In this case, we'd expect 0 or 1 `FixResult`.

src/codemodder/codetf/v3/codetf.py
nahsra: I think associating a change to a dependency can allow downstream users to tell users add a dependency (like if it failed), or that this increases reliance on a dependency, etc.

src/codemodder/codetf/v3/codetf.py
drdavella: > Aren't we only giving it 1 input as well? We had discussed as an input a single SARIF, and a single result ID (or something.)

As a practical matter I'm still not convinced that's the best approach although we can work towards that setup. In any case I don't think we want to preclude the possibility of batch processing.

>  In this case, we'd expect 0 or 1 FixResult.

I would argue that the absence of any `FixResult` should be an exceptional condition. We should potentially consider making `FixResult` required. The fact that this would encode fix status means we should return a failure status if processing is failed or skipped

drdavella:
I am merging this as it is still currently in draft status and does not affect any production codepaths. However making it available will be useful for prototyping new features and functionality.

PR #1025, opened 2025-03-12
Disable noisy log message by default

drdavella:
Preserves existing behavior for the CLI


clavedeluna (approved):
I'm subconsciously been wanting this!

PR #1030, opened 2025-03-24
Added Independent Finding Execution Behavior

andrecsilva:
## Overview
Adds independent finding behavior for codemods

## Description
Independent finding behavior means that codemods will execute once for each finding (i.e. remediation), as opposed to once for all findings (i.e. hardening). Hardening aims to produce a file with all the changes, while remediation will produce diffs for each finding. Remediation is the default behavior accessible with the `codemodder` command, while hardening is now accessible as `codemodder_hardening`.

## Additional Details
* Remediation won't write any files as the default behavior, this includes the dependencies.
* While find and fix codemods will still be accessible over the remediation behavior, there is no difference over the hardening behavior. The only exception being codemods that use semgrep scans for findings.
* Most tests with 2+ changes are now being tested for remediation behavior. I've found this gives us a healthy balance of testing for both behaviors while avoiding duplicating each test.

### DIFF
--- a/.github/workflows/codemod_pygoat.yml
+++ b/.github/workflows/codemod_pygoat.yml
@@ -38,6 +38,6 @@ jobs:
           repository: pixee/pygoat
           path: pygoat
       - name: Run Codemodder
-        run: codemodder_hardening --dry-run --output output.codetf pygoat
+        run: codemodder --dry-run --output output.codetf pygoat
       - name: Check PyGoat Findings
         run: make pygoat-test
--- a/.github/workflows/codemod_pygoat.yml
+++ b/.github/workflows/codemod_pygoat.yml
@@ -38,6 +38,6 @@ jobs:
           repository: pixee/pygoat
           path: pygoat
       - name: Run Codemodder
-        run: codemodder --dry-run --output output.codetf pygoat
+        run: codemodder_hardening --dry-run --output output.codetf pygoat
       - name: Check PyGoat Findings
         run: make pygoat-test
--- a/integration_tests/test_dependency_manager.py
+++ b/integration_tests/test_dependency_manager.py
@@ -136,7 +136,7 @@ class TestDependencyManager:
         os.chmod(tmp_repo / self.requirements_file, 0o400)
 
         command = [
-            "codemodder_hardening",
+            "codemodder",
             tmp_repo,
             "--output",
             self.output_path,
--- a/integration_tests/test_dependency_manager.py
+++ b/integration_tests/test_dependency_manager.py
@@ -136,7 +136,7 @@ class TestDependencyManager:
         os.chmod(tmp_repo / self.requirements_file, 0o400)
 
         command = [
-            "codemodder",
+            "codemodder_hardening",
             tmp_repo,
             "--output",
             self.output_path,
--- a/integration_tests/test_multiple_codemods.py
+++ b/integration_tests/test_multiple_codemods.py
@@ -25,7 +25,7 @@ class TestMultipleCodemods:
         shutil.copy(pathlib.Path(SAMPLES_DIR) / source_file_name, directory)
 
         command = [
-            "codemodder_hardening",
+            "codemodder",
             directory,
             "--output",
             str(codetf_path),
--- a/integration_tests/test_multiple_codemods.py
+++ b/integration_tests/test_multiple_codemods.py
@@ -25,7 +25,7 @@ class TestMultipleCodemods:
         shutil.copy(pathlib.Path(SAMPLES_DIR) / source_file_name, directory)
 
         command = [
-            "codemodder",
+            "codemodder_hardening",
             directory,
             "--output",
             str(codetf_path),
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,7 +82,7 @@ test = [
     "flask_wtf~=1.2.0",
     "fickling~=0.1.0,>=0.1.3",
     "graphql-server~=3.0.0b7",
-    "unidiff>=0.75",
+    "unidiff>=0.7.5",
 ]
 complexity = [
     "radon==6.0.*",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ Repository = "https://github.com/pixee/codemodder-python"
 
 [project.scripts]
 codemodder = "codemodder.codemodder:main"
-codemodder_hardening = "codemodder.codemodder:harden"
+codemodder-remediation = "codemodder.codemodder:remediate"
 generate-docs = 'codemodder.scripts.generate_docs:main'
 get-hashes = 'codemodder.scripts.get_hashes:main'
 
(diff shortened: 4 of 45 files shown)


src/codemodder/codemodder.py
```diff
@@ -135,6 +136,7 @@ def run(
     sast_only: bool = False,
     ai_client: bool = True,
     log_matched_files: bool = False,
+    hardening: bool = False,
```
drdavella:
```suggestion
    hardening: bool = True,
```

pyproject.toml
```diff
@@ -45,6 +45,7 @@ Repository = "https://github.com/pixee/codemodder-python"
 
 [project.scripts]
 codemodder = "codemodder.codemodder:main"
+codemodder_hardening = "codemodder.codemodder:harden"
```
clavedeluna: for consistency with other commands, I suggest `codemodder-hardening`

.github/workflows/codemod_pygoat.yml
```diff
@@ -38,6 +38,6 @@ jobs:
           repository: pixee/pygoat
           path: pygoat
       - name: Run Codemodder
-        run: codemodder --dry-run --output output.codetf pygoat
+        run: codemodder_hardening --dry-run --output output.codetf pygoat
```
drdavella: I think we should preserve the existing behavior of the command line for now.

integration_tests/test_dependency_manager.py
```diff
@@ -136,7 +136,7 @@ def test_fail_to_add(self, tmp_repo):
         os.chmod(tmp_repo / self.requirements_file, 0o400)
 
         command = [
-            "codemodder",
+            "codemodder_hardening",
```
drdavella: Let's keep the existing behavior for now.

integration_tests/test_multiple_codemods.py
```diff
@@ -25,7 +25,7 @@ def test_two_codemods(self, codemods, tmpdir):
         shutil.copy(pathlib.Path(SAMPLES_DIR) / source_file_name, directory)
 
         command = [
-            "codemodder",
+            "codemodder_hardening",
```
drdavella: Preserve existing behavior for now.

pyproject.toml
drdavella: I don't think this is necessary. We expect per-finding mode to be accessed exclusively by client library calls for now and we should preserve the existing `codemodder` behavior as the default in the near term.

src/codemodder/codemodder.py
```diff
@@ -231,7 +233,7 @@ def run(
     return codetf, 0, token_usage
 
 
-def _run_cli(original_args) -> int:
+def _run_cli(original_args, hardening=False) -> int:
```
drdavella:
```suggestion
def _run_cli(original_args, hardening=True) -> int:
```

src/codemodder/codemodder.py
```diff
@@ -284,10 +287,22 @@ def _run_cli(original_args) -> int:
         codemod_registry=codemod_registry,
         sast_only=argv.sonar_issues_json or argv.sarif,
         log_matched_files=True,
+        hardening=hardening,
     )
     return status
 
 
+def harden():
```
drdavella: I don't think this is necessary; see comments above.

src/codemodder/codemods/base_codemod.py
```diff
@@ -226,24 +227,64 @@ def _apply(
...
-            self._process_file, context=context, results=results, rules=rules
-        )
-
-        contexts = []
-        if context.max_workers == 1:
-            logger.debug("processing files serially")
-            contexts.extend([process_file(file) for file in files_to_analyze])
-        else:
+        # Do each result independently and outputs the diffs
+        if not hardening:
```
drdavella: This if/else feels like each branch should maybe be a separate function. It would probably enable the call to be a one-liner as well.

src/codemodder/codemods/base_codemod.py
```diff
@@ -226,24 +227,64 @@ def _apply(
...
+            # gather positional arguments for the map
+            resultset_arguments: list[ResultSet | None] = []
+            path_arguments = []
+            if results:
+                for result in results.results_for_rules(rules):
+                    # this need to be the same type of ResultSet as results
+                    singleton = results.__class__()
+                    singleton.add_result(result)
+                    result_locations = self.get_files_to_analyze(context, singleton)
+                    # We do an execution for each location in the result
```
drdavella: I don't think this is correct. We only need to execute each fix per _finding_. A finding may report multiple locations, but will still result in only a single fix (which itself may touch multiple locations). I think the distinction is subtle but we still execute each codemod only once per _finding_.

Let me know whether there's something I'm missing or whether maybe the comment isn't quite right.

drdavella (changes requested):
Overall this looks good. My comments boil down to two basic issues:
* We should preserve the current hardening behavior as the default for now
* I'm not totally convinced that locations are being handled correctly for the hardening case

src/codemodder/codemods/test/utils.py
```diff
@@ -72,9 +73,10 @@ def run_and_assert(
 
         path_exclude = [f"{tmp_file_path}:{line}" for line in lines_to_exclude or []]
 
+        print(expected_diff_per_change)
```
clavedeluna:
```suggestion
```

clavedeluna (commented):
a few code comments that seem unnecessarily explaining code (LLM generated likely) that are worth removing

src/codemodder/codemods/base_codemod.py
andrecsilva: Two things to note: our transformers are file-based, that is, they execute per file. ChangeSet objects only reports changes to a single file.

This particular piece of code mimics the behavior we had before. If you only had one result with multiple locations, you would run the codemod for each location and produce a changeset object for each location.

What changed is that now each changeset is only associated with a single finding.

pyproject.toml
andrecsilva: We do use the new command line script for integration tests.

src/codemodder/codemods/base_codemod.py
```diff
@@ -185,38 +185,74 @@ def get_files_to_analyze(
...
+            return None
+
+        # Do each result independently and outputs the diffs
+        # gather positional arguments for the map
+        resultset_arguments: list[ResultSet | None] = []
+        path_arguments = []
+        if results:
+            for result in results.results_for_rules(rules):
+                # this need to be the same type of ResultSet as results
+                singleton = results.__class__()
```
drdavella: Sorry for such late feedback but it feels like this could be done more cleanly by implementing a class method on the `ResultSet` type:
```python
T = TypeVar("T", bound=Result) # we may already have this somewhere
@classmethod
def from_single_result(cls, result: T) -> Self:
    new = cls()
    cls.add_result(result)
    return new
```
Alternatively there may be a way to make it a method of `Result` instead, I'm not sure which is cleaner. In either case it avoids the introspection and feels like better encapsulation.

drdavella (approved):
Approved and looks good but please consider making the change to clean up creation of singleton result sets.

PR #1055, opened 2025-05-01
Fixed rule_id being inserted into finding_id field

andrecsilva:


drdavella (commented):
Can you explain the reason for this change? This behavior is intentional.

EDIT: fwiw my main objection is with making finding ID optional. If you are concerned about falling back on `rule_id` then we should just make it an error to not properly populate finding ID

andrecsilva:
> Can you explain the reason for this change? This behavior is intentional.
> 
> EDIT: fwiw my main objection is with making finding ID optional. If you are concerned about falling back on `rule_id` then we should just make it an error to not properly populate finding ID

Finding id is optional as per the codetfv2 spec: 
https://github.com/pixee/codemodder-python/blob/05bbefce72cf90b9ab1988ca324848daf7c9422f/src/codemodder/codetf/v2/codetf.py#L143-L145

The spec also expects the `finding_id` to be unique an thus falling back to `rule_id` may cause problems. Since we expect sarif (and similar formats) to have some unique id (i.e. `guid` in sarif) the fallback is not really necessary. Honestly, the `rule_id` fallback seems to be used by internal semgrep codemods, so we can safely ditch it.

I'll implement your suggestion and make it throw an error.

drdavella (approved):

PR #1079, opened 2025-07-04
Added v2 to v3 result transformation that includes package changes

andrecsilva:


src/codemodder/codetf/v3/codetf.py
```diff
@@ -148,6 +151,48 @@ def from_v2_aimetadata(ai_metadata: AIMetadatav2) -> AIMetadata:
...
 
+def from_v2_result_per_finding(result: Result) -> FixResult | None:
+    """
+    This transformation assumes that the v2 result will only contain a single fixedFinding for all changesets.
+    """
+    # Find the changeset with a fixedFinding
+    try:
+        changeset: v2ChangeSet = next(cs for cs in result.changeset if cs.fixedFindings)
+    except StopIteration:
+        logger.debug("No fixedFinding in the given Result")
```
clavedeluna: StopIteration could be raised from either result.changeset being empty or it's not empty but none of the changesets have a. fixefinding so the debug statement isn't quite accurate

src/codemodder/codetf/v3/codetf.py
```diff
@@ -148,6 +151,48 @@ def from_v2_aimetadata(ai_metadata: AIMetadatav2) -> AIMetadata:
...
+        ChangeSet(
+            path=cs.path, diff=cs.diff, changes=[c.to_common() for c in cs.changes]
+        )
+        for cs in result.changeset
+    ]
+
+    generation_metadata = GenerationMetadata(
+        strategy=Strategy.ai if changeset.ai else Strategy.deterministic,
+        ai=from_v2_aimetadata(changeset.ai) if changeset.ai else None,
+        provisional=False,
```
clavedeluna: will this function always be called from non-provisional contexts?

src/codemodder/codetf/v3/codetf.py
andrecsilva: True, I'll fix that

src/codemodder/codetf/v3/codetf.py
andrecsilva: No clue, not sure why this flag exists in the first place.

clavedeluna: provisional=True is "magic"... maybe we no longer make this distinction ?

drdavella: It should be derived from the v2 changeset the same way as these other metadata fields.
```suggestion
        provisional=changeset.provisional,
```

src/codemodder/codetf/v3/codetf.py
```diff
@@ -148,6 +151,48 @@ def from_v2_aimetadata(ai_metadata: AIMetadatav2) -> AIMetadata:
     )
 
 
+def from_v2_result_per_finding(result: Result) -> FixResult | None:
+    """
+    This transformation assumes that the v2 result will only contain a single fixedFinding for all changesets.
+    """
+    # Find the changeset with a fixedFinding
+    try:
+        changeset: v2ChangeSet = next(cs for cs in result.changeset if cs.fixedFindings)
```
drdavella: Fixed findings can belong to individual `Change` entries of a change set as well, so you need to check there too, both here and below.

src/codemodder/codetf/v3/codetf.py
drdavella: Is there any chance that for our purposes it makes more sense to pass in these metadata fields as parameters to this new function? For the use case I'm thinking of, I'd argue that the answer is yes.

drdavella (changes requested):

src/codemodder/codetf/v3/codetf.py
andrecsilva: Those are usually replicated. That is, the changeset `fixedFindings` should always include the ones contained in the changes.

I'll make it check each individual change in any case.

drdavella (approved):

